PR #36, opened 2021-03-30
Search rewrite

gareth-j:
I've rewritten the search function, at the moment it doesn't include searching by rank but I'll add this next. 

You can now search using any number of keyword arguments, these will then be matched with the metadata stored by each `Datasource`.


mrghg:
Can you give an example of the suggested usage?

gareth-j:
This has been finished under #86.

PR #37, opened 2021-03-30
Fixing the visualisation of observations data in the object store

gareth-j:


gareth-j:
This was fixed in #136

PR #45, opened 2021-03-31
Inlet labels

mrghg:
This PR adds back functionality to label an inlet with a value not from the GCWerks files. E.g. we don't want to rely on having "air" or "any" as the inlet label. Also, for CGO, it was giving us a load of inlets, which are all basically the same ("75m_1", "75m_2"... I've relabelled these all to "70m").

I'm submitting this now, but I haven't had chance to test it, because of the issues I'm having with the object store.

I think (but haven't tried) that the tests will fail...

Will try to work on it some more ASAP.


gareth-j:
This was implemented in a slightly different way in #54

PR #49, opened 2021-04-12
Fp data merge

rt17603:
Creating a notebook with examples for using footprint data merge


gareth-j:
I've made some changes to this but work needs to be done on the `Footprint` and `Emissions` classes to bring them up to spec with ObsSurface before any more changes should be made.

gareth-j:
I've just merged in the work needed to bring the footprint and emissions processing functionality up to a similar level as ObsSurface. Now I just need to update some the notebook to take into account some changes in the functions.

rt17603:
Have pulled the branch and the footprint_workflow notebook works fine for me now. I'm assuming additional code added in is from the devel branch? Did you want me to look at anything in particular for this?

gareth-j (approved):
Looks good.

PR #56, opened 2021-04-20
Adding in Datasource lookup to footprint and emissions objects

gareth-j:
Adding in the ability to lookup datasources using metadata as is done in `ObsSurface`.


gareth-j:
@rt17603 could you have a look at https://github.com/openghg/openghg/blob/f9d42312f0d969b15ac11788bf877e667a7062c9/openghg/modules/_footprints.py#L148 to make sure this way of identifying a footprint makes sense?

rt17603:
result = self._datasource_table[site][domain][model][height]

For a footprint, site domain and height are always needed to be able to identify with "model" and "species" optional additional identifiers i.e. "model" and "species" won't always be included.

Issue #59, opened 2021-04-23
Mismatch uuid error when reading files into object store


rt17603:
When populating the object store to contain the primary files for AGAGE, DECC and ICOS data, the following error was returned for some files:

```
Traceback (most recent call last):
  File "/home/rt17603/openghg/openghg/modules/_obs_surface.py", line 156, in read_file
    obs.add_datasources(datasource_uuids, metadata)
  File "/home/rt17603/openghg/openghg/modules/_base.py", line 160, in add_datasources
    raise ValueError("Mismatch between assigned uuid and stored Datasource uuid.")
ValueError: Mismatch between assigned uuid and stored Datasource uuid.
```

Details of the affected files and the inputs used for the ObsSurface.read_file() function are attached here:
[files_inputs_error.txt](https://github.com/openghg/openghg/files/6366565/files_inputs_error.txt)

Code for this is included in "create_obj_store" branch with new (rough) files in folder "openghg/notebooks/data"

PR #62, opened 2021-04-28
Fix for UUID error due to incorrect inlet in metadata

gareth-j:
This fixes the the issue in #59 which was due the incorrect inlet being saved in the metadata for that Dataset. This was due to the metadata dictionary being copied in the wrong place which resulted in the incorrect inlet label being present in the metadata.


openghg/data/process_gcwerks_parameters.json
```diff
@@ -1,47 +1,47 @@
 {
     "GCWERKS": {
         "instruments": {
-            "GCMD": "",
-            "GCMS": "",
+            "gcmd": "",
+            "gcms": "",
```
rt17603, line 5: Quick question: will the ObsSurface.read_file() function still accept instruments as upper case? Or will they need to lower case now?

rt17603 (approved):
I have looked over the code and re-run for the data files which were previously producing the this mismatch UUID error. All files can now be read without error. Looks good to me.

PR #63, opened 2021-04-28
Adding sampling period to metadata and attributes

rt17603:
For GCWERKS and CRDS classes, an overall "sampling_period" has been added as both metadata and a Dataset attribute. 

Current functionality:
 - Sampling period is interpreted and stored as an integer number of seconds.
 - When sampling_period is supplied to ObsSurface.read_file() this is currently expected as a string with a frequency unit. An error will be raised if this is not a string of the right format.
 - A supplied sampling period is checked against the expected value and an error raised if this does not match.
 - Removed: _sampling_period is no longer included as an explicit class attribute
 - CRDS - the sampling period is interpreted from the file name
 - GCWERKS - the sampling period is interpreted from the json file for each instrument type

As written now, for these two classes _any input sampling period is not used, only used as a check against the interpreted values_.

Future considerations:
 - For the more generic case where the sampling period can vary for each measurement, we may need to include this as a data variable (either instead of an attribute / metadata or as well as). In this case the overall "sampling_period" could be set as a flag such as "data" or "various".
 - Since the supplied sampling period is not used, we may wish to remove this option in future or to allow it to supercede the extracted value.


gareth-j:
Looks good.

PR #64, opened 2021-04-29
Fix for returning too wide a daterange

### LINKED ISSUE #58: Extracting data in a date range rounds up or down to the nearest year end
When using the `get_obs_surface()` function with a `start_date` and an `end_date` input (as string values) this rounds to the nearest year end.

For example:
```
from openghg.localclient import get_obs_surface
# Extracting for nearest year start and year end when 
# including start and end dates
data = get_obs_surface(site="mhd", species="ch4",
                       start_date="2005-03-01", end_date="2008-02-01")
```
Extract_ObsData_DateRange

In this case including:
 - Input date range: "2005-03-01" to "2008-02-01"
 - Date range in file: "2005-01-01T00:44:22.500000000" to "2008-12-31T23:14:22.500000000"

We may want to decide how fine-grained we want this (i.e. is the nearest month ok) but we will most likely want this to be less than yearly.

gareth-j:
This adds in a slice to ensure the data returned from get_obs_surface only falls within the dates passed. It was returning dates outside the daterange selected by the user due to the way the data is chunked and stored in the object store. This should close #58.


rt17603:
So the way it works in the code at the moment is we extract the date range from start_date and up to but not including end_date. I think pandas by default *does* include the end date which is reflected in the checks I just made for the Mace Head data (after loading this all in from the 3 sources).

data = get_obs_surface(site="mhd", species="ch4",
                       start_date="2005-03-01", end_date="2008-02-01")

Start and end dates:
2005-03-01T00:18:22.500000000
2008-02-01T23:26:22.500000000

The tests included at the moment won't be checking for this as the last date in the data is before the end_date inputted.

rt17603 (changes requested):
The date range can now be selected but I think we should update this to not include the end_date in our date range for consistency with current functionality elsewhere.

gareth-j:
Closing this as mostly implemented, have opened #161

PR #68, opened 2021-05-11
Reimplentation of the ranking search functionality

gareth-j:
This adds back in the ability to search for data and get the highest ranked data across the daterange the user specifies.


rt17603:
A few things I've picked up from looking into this branch:

If you don't set the ranking before calling the `search` function this produces an error. For example I loaded in a new Dataset for Tacolneston (TAC)

`tac_files = find_crds_files("TAC")`

`tac_results = ObsSurface.read_file(filepath=tac_files, 
                                    data_type="CRDS", 
                                    site="TAC", network="DECC")`

`search_res = search(site="tac", species="ch4")`

Returned:
`ValueError: No ranking data set for the given search parameters. Please refine your search to include a specific site, species and inlet.`

For my information, was the purpose of the search function to only return the highest ranked data or to return all the data (along with potential rankings)?

rt17603:
I don't think this line (229) within the code:

`File "/home/rt17603/openghg/openghg/modules/_base.py", line 229, in set_rank`
`    self._rank_data[uuid][rank].extend(date_range)`

`TypeError: 'Dict' object is not callable`

is valid with the `addict` package.

When this line is accessed (if the uuid exists but does not have an overlapping DataRange) this error was produced.

rt17603:
The `overwrite` input does not allow new ranks to be set at present:

Input code

`import os`
`os.environ["OPENGHG_PATH"] = "/work/chxmr/objectStore"`

`uid_248 = bsd_results["processed"]["bsd.picarro.1minute.248m.dat"]["ch4"]`

`obs = ObsSurface.load()`
`obs.set_rank(uuid=uid_248, rank=1, date_range="2012-01-01_2013-01-01", overwrite=True)`

Returns:

`Traceback (most recent call last):`

`  File "/home/rt17603/Code_BP1/OpenGHG/Load_BSD_try_ranking.py", line 35, in <module>`
`    obs.set_rank(uuid=uid_248, rank=1, date_range="2012-01-01_2013-01-01", overwrite=True)`

`  File "/home/rt17603/openghg/openghg/modules/_base.py", line 225, in set_rank`
`    Overlapping dateranges are {e} and {d}"`

`ValueError: This datasource has rank 1 for dates that overlap the ones given.  `                                                   
`Overlapping dateranges are 2012-01-01_2013-01-01 and 2012-01-01_2013-01-01`

The ranking data is retained regardless of if the data itself is overwritten (call ObsSurface.read_file() again with overwrite set to True). I don't think this would show up in your notebook example as you're setting up a temporary object store. This only happened when I switched to writing to the retained object store on BP1 (e.g. "/work/chxmr/objectStore" used above).

gareth-j:
> I don't think this line (229) within the code:
> 
> `File "/home/rt17603/openghg/openghg/modules/_base.py", line 229, in set_rank`
> ` self._rank_data[uuid][rank].extend(date_range)`
> 
> `TypeError: 'Dict' object is not callable`
> 
> is valid with the `addict` package.
> 
> When this line is accessed (if the uuid exists but does not have an overlapping date_range) this error was produced.

OK I think I've fixed this in 990fa1c3276d89f4a51dced835d7cb62f876579b and I've added a test to `test_obssurface.py`.

rt17603:
**Scenario 2:** Try to set the rank with the same dates twice. Not a useful scenario but could easily see this happening perhaps by accident or as part of a loop. 

Both runs:
from file_search import find_crds_files
from openghg.modules import ObsSurface

import os
os.environ["OPENGHG_PATH"] = "/work/chxmr/objectStore"

tac_files = find_crds_files("TAC")
tac_results = ObsSurface.read_file(filepath=tac_files, data_type="CRDS", site="TAC", network="DECC", overwrite=True)

uid_100 = tac_results["processed"]['tac.picarro.1minute.100m.dat']["ch4"]

Run 1 and Run 2
obs.set_rank(uuid=uid_100, rank=1, date_range="2012-01-01_2013-01-01", overwrite=True)

After Run 2:
`print(obs._rank_data)`
`{'4ba4b662-a871-4d0d-a834-2914a5758be0': {'2013-01-01-00:00:00+00:00_2013-01-01-00:00:00+00:00': 1}`

**Result**: this has update the rank label to '2013-01-01-00:00:00+00:00_2013-01-01-00:00:00+00:00' when it should return the original  value '2012-01-01_2013-01-01' (or tzaware equivalent).

`search_res = search(site="tac", species="ch4")`
Again, search function returns an empty dictionary for this.

rt17603:
I'm also having the issue that to restart any scenario I have to delete the object store to remove any retained ranking data - it would be useful to add a way to delete the stored ranking (especially if a mistake has been made).

rt17603:
Ignore previous comment - just saw about the `clear_rank` method being added - so will give that a go!

gareth-j:
I just realised this hasn't been implemented - #72 but I'll leave it for now until we know the rest of this works as we want it to.

rt17603:
Refactored Scenario 1 above, and I think the `overwrite` input is still having the same issue.

**First pass**

`r.set_rank(key="ch4_100m_picarro", rank=1, start_date="2012-01-01", end_date="2013-01-01")`
`r.set_rank(key="ch4_185m_picarro", rank=1, start_date="2013-01-01", end_date="2050-01-01")`

**Second pass**

`r.set_rank(key="ch4_100m_picarro", rank=1, start_date="2012-01-01", end_date="2014-01-01", overwrite=True)`
`r.set_rank(key="ch4_185m_picarro", rank=1, start_date="2014-01-01", end_date="2050-01-01", overwrite=True)`

*Print outs*

**First pass**

Ranking data for 100m
{'2012-01-01-00:00:00+00:00_2013-01-01-00:00:00+00:00': 1}
Ranking data for 185m
{'2013-01-01-00:00:00+00:00_2050-01-01-00:00:00+00:00': 1}

**Second pass**

Ranking data for 100m
{'2012-01-01-00:00:00+00:00_2011-12-31-23:59:00+00:00': 1}
Ranking data for 185m
{'2050-01-01-00:00:00+00:00_2050-01-01-00:00:00+00:00': 1}

However, calling if the `clear_rank` method is called before the second `set_rank` call then this updates successfully. e.g. writing this as:

i.e.
`r.set_rank(key="ch4_100m_picarro", rank=1, start_date="2012-01-01", end_date="2013-01-01")`
`r.set_rank(key="ch4_185m_picarro", rank=1, start_date="2013-01-01", end_date="2050-01-01")`

`r.clear_rank(key="ch4_100m_picarro")`
`r.clear_rank(key="ch4_185m_picarro")`
`r.set_rank(key="ch4_100m_picarro", rank=1, start_date="2012-01-01", end_date="2014-01-01", overwrite=True)`
`r.set_rank(key="ch4_185m_picarro", rank=1, start_date="2014-01-01", end_date="2050-01-01", overwrite=True)`

So one option could be to just use this workflow to update and remove overwrite as an option? The downside is that presumably this will delete all associated ranks for that site even if you just wanted to update one but would perhaps make this less complex.

rt17603:
Looking into **Scenario 1** again where we want to update the ranking with a new end_date this now updates correctly with `overwrite=True` input:

`r.set_rank(key="ch4_100m_picarro", rank=1, start_date="2012-01-01", end_date="2013-01-01")`
`r.set_rank(key="ch4_100m_picarro", rank=1, start_date="2012-01-01", end_date="2014-01-01", overwrite=True)`
`print("Ranking data for 100m")`
`print(r.get_specific_source(key="ch4_100m_picarro"))`

Ranking data for 100m
{'2012-01-01-00:00:00+00:00_2014-01-01-00:00:00+00:00': 1}

rt17603:
**Scenario 2** - the same dates are included twice with `overwrite=True` also now produces the expected output (the same dates).

rt17603:
**Scenario 3**: Testing the start_date in a date range can be updated when calling `overwrite=True`. 

`r.set_rank(key="ch4_185m_picarro", rank=1, start_date="2013-01-01", end_date="2050-01-01")`
`r.set_rank(key="ch4_185m_picarro", rank=1, start_date="2014-01-01", end_date="2050-01-01", overwrite=True)`
`print("Ranking data for 185m")`
`print(r.get_specific_source(key="ch4_185m_picarro"))`

Ranking data for 185m
{'2013-01-01-00:00:00+00:00_2050-01-01-00:00:00+00:00': 1}

**Result**: So although Scenario 1 demonstrated the `end_date` can be updated using `overwrite=True`, in this case, the `start_date` parameter hasn't been updated from "2013-01-01" to "2014-01-01"

rt17603:
A workflow using the `clear_rank` method works well as an alternative in the above scenarios and produces expected output.

rt17603 (approved):
For scenarios tested this seems to be performing sensibly. Will need to lay out expectation and workflow clearly for this in tutorial documents when finalised.

gareth-j:
This work has been continued in #86 .

PR #70, opened 2021-05-11
Updated resampling when combining footprints and obs data

rt17603:
Uses input "sampling_period" attribute when deciding on resampling strategy for footprints and data.

If this attribute is not present this will still attempt to derive this from the obs data frequency but will produce an error is this is too variable. This is an alternative to the suggested approach on applying a default sample_period of "1H".

Other changes:
 - The default for the `resample_to` input parameter has also been updated to "coarsest" (instead of "obs") as this is more appropriate for expected inputs.
 - This can now be used for "flask" samples as long as the "sample_period" is specified.
 - Using `get_obs_surface()` will now add "averaged_period" and "averaged_period_str" attributes to the output Dataset (this is not used anywhere yet but plan to use this within `footprint_data_merge()` in future).

This does not allow for the case of a variable sampling period per measurement.

### DIFF
--- a/openghg/processing/_process_footprint.py
+++ b/openghg/processing/_process_footprint.py
@@ -274,7 +274,8 @@ def align_datasets(
 
     if platform is not None:
         platform = platform.lower()
-        if platform in ("satellite"):#, "flask"):
+        # Do not apply resampling for "satellite" (but have re-included "flask" for now)
+        if platform in ("satellite"):
             return obs_data, footprint_data
 
     # Get the period of measurements in time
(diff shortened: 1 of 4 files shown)


rt17603:
Has been checked and are we happy to merge this in?

openghg/processing/_process_footprint.py
```diff
@@ -274,25 +274,31 @@ def align_datasets(
 
     if platform is not None:
         platform = platform.lower()
-        if platform in ("satellite", "flask"):
+        # Do not apply resampling for "satellite" (but have re-included "flask" for now)
+        if platform in ("satellite"):
```
gareth-j, line 278: This could just be a simple comparison

rt17603: It could, but was thinking we'd may need to expand this output to be set of values again in the future?

gareth-j: OK, I changed it with 0e0c5c687f9689ee8c42c0ad1255a3ddb96eebd2 before seeing this, we can change it back when we need to.

PR #74, opened 2021-05-21
Supporting legacy formats (compatability with acrg)

rt17603:
Adding legacy compatability to ObsData object (created when get_obs.. functions are called) to current ACRG formats. This is to mimic a dictionary to allow the dataset to be returned when the site key is used (e.g. `ObsData["MHD"]` would return the dataset - equivalent to `ObsData.data`). This is to ensure initial legacy compatibility with current ACRG formats which can be depreceated over time and then replaced.

This won't impact the current workflow and can be replaced when we don't need it anymore, but will make transition from current acrg formats to any new formats easier.


rt17603:
Seems to be an issue with generating the documentation from the workflow notebook but I didn't make any changes on this branch to the notebook format.

gareth-j:
> Seems to be an issue with generating the documentation from the workflow notebook but I didn't make any changes on this branch to the notebook.
> 
> This is causing one of the test cases below to the fail (ubuntu 3.7) but all other versions seem to pass fine.

Yeah this is an issue that's been introduced with the new jinja release going 2 -> 3. In the `searchRank` branch I've updated the requirements to pin the jinja version until I get a chance to look into what's going on / something changes on their end.

rt17603:
Example code to check this:

`from openghg.modules import ObsSurface`
`from openghg.localclient import get_obs_surface`

`import os`
`import tempfile`
`tmp_dir = tempfile.TemporaryDirectory()`
`os.environ["OPENGHG_PATH"] = tmp_dir.name # "/tmp/openghg_store"`

`home = os.getenv("HOME")`
`decc_file = os.path.join(home,"openghg/notebooks/tutorials/data/DECC/bsd.picarro.1minute.248m.dat")`

`decc_results = ObsSurface.read_file(filepath=decc_file, data_type="CRDS", site="bsd", network="DECC")`

`data = get_obs_surface(site="bsd", species="ch4", inlet="248m")`

`print("Type of data returned from get_obs_surface",type(data))`
`print(f"Extract Dataset {type(data.data)} using data.data")`
`print(data.data)`
`print(f"Can also extract Dataset {type(data['bsd'])} using legacy format data[sitecode]")`
`print(data["bsd"])`

The ability to extract the Dataset using `data["bsd"]` is what has been added in this pull request.

gareth-j (approved):

rt17603:
Waiting on fix related to  jinja version going into devel, to check the tests all pass before this can be merged.

Would it possible to move this fix into devel (rather than the `searchRank` branch) so this can be checked?

Issue #71, opened 2021-05-19
Default sampling period for modules


gareth-j:
Should we have a default sampling period for modules or insist on it being passed in by the user / read somehow?

Currently the BTT, Cranfield, EUROCOM modules need looking at.

rt17603:
BTT (BT Tower) - the sampling period seems to be 30 minutes for this (file naming and looking at the data itself).
Cranfield data - the sampling period seems consistently be 5 seconds for the data present.

For EUROCOM it may depend on which data falls into that category - where were you grabbing your EUROCOM data from?

For CRDS and GCWERKS these sampling periods were being derived from an underlying json file - do we want to adopt a similiar approach for all modules where possible?

rt17603:
On the question of a default sampling period, I think we decided we can have one on a case-by-case basis (like we're doing at the moment with having a "sampling_period" defined in json files) but we don't want to have a universal value for these files.

I think this may become information we need in advance when putting together the different data processing modules rather than asking the user to input this but we can leave allowing this as an input for now.

gareth-j:
Add the jinja2 pinning and retry test runners.

PR #75, opened 2021-05-21
Updating default sampling period for BTT module

rt17603:
Default sampling period has been added for BTT module.

Links to Issue #71.

### DIFF
--- a/openghg/data/attributes.json
+++ b/openghg/data/attributes.json
@@ -182,6 +182,7 @@
     },
     "instrument": "Picarro 2311-f",
     "inlet": "192m",
+    "sampling_period": 1800,
     "global_attributes": {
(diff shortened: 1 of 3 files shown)


rt17603:
Same problem as for PR #74 - seems to be an issue with generating the documentation from the workflow notebook but I didn't make any changes on this branch to the notebook. May be linked to this version of devel?

This is causing one of the test cases below to the fail (ubuntu 3.7) but all other versions seem to pass fine.

rt17603:
As for another PR, this is waiting on fix related to  jinja version going into devel, to check the tests all pass before this can be merged.

Again, would it possible to move this fix into devel (rather than the `searchRank` branch) so this can be checked?

gareth-j:
#93 should remove the block on this.

PR #88, opened 2021-06-16
Add basic Eulerian model class

rt17603:
This pull request adds a basic Eulerian model class for adding Eulerian model data (e.g. GEOSChem) onto the OpenGHG platform.

Several additional considerations here:
 - At the moment this is stored using the "model", "species" and "date" keywords but this will most likely not provide enough detail for this to be stored uniquely. This is something we need to iron out (and may involve not storing against keywords specifically).
 - This has been tested against GEOSChem data but will also want to test this against other Eulerian model inputs (in particular WRFChem and may want to try out MOZART as this is something we have access to at the moment).


rt17603:
Hopefully this is a fairly innocuous change to merge in, as it's a new piece of functionality but are we happy to merge this in? Would be useful to have this functionality in devel and other branches moving forward.

gareth-j (approved):
Looks good.

PR #89, opened 2021-06-17
Update to NOAA module to retain "units" attributes

rt17603:
Updated NOAA module to retain details of "units" attributes for each of the input variables and add 'units' metadata value.

**Requires update of xarray version to >= 0.18.2 (added to requirements-dev.txt).**

There is still a question around how to represent these units in the final output

- Should "units" within the metadata be a particular format e.g. "ppb", "ppm" etc.
- Should "units" for each data variable be 1e-9 etc., 

At the moment this units format is of the form "mol mol-1" or "micromol mol-1"


rt17603:
Made this as a draft because it would require an update to our underlying xarray version to work but would be useful to know what steps to take so we can merge this in.

gareth-j:
I've moved the `xarray` requirement to `requirements.txt` instead of the dev version but other than this looks great.

gareth-j:
I think this is ready to merge in now.

PR #91, opened 2021-07-08
Adding basic calculation of modelled mole fraction

rt17603:
Adding very basic calc_timeseries functionality to footprints_data_merge() and updated footprint_workflow notebook to include this.

*Doesn't work with high time resolution footprints yet - this workflow needs to be folded into footprints as well.*


gareth-j (commented):
Looks good.

PR #96, opened 2021-07-15
Add type checking with mypy

gareth-j:
Adding in type checking with `mypy`.


gareth-j:
Type checking relatively complete for the module code, still need to add some stubs, see #100 .

PR #130, opened 2021-09-15
Addict fix

mrghg:
In the visualise_store function, a class "aDict" was imported and used, but (current version of?) addict only has "Dict". I changed it, and it seems to work now...


gareth-j (commented):
This has now been fixed as is on devel.

PR #132, opened 2021-09-15
Demo bugfix

mrghg:
I think there's a bug in _file_search.find_files. I think it should be:

```
    for data_type in data_folders:
        data_details = all_instrument_details[data_type]
```

rather than:

```
    for data_type in all_instrument_details:
        data_details = all_instrument_details[data_type]
```

Because otherwise you get a key error if ```data_type``` isn't a key in ```data_folders``` (demo version happened to have identical keys for both dictionaries)


rt17603:
The original `find_all_files()` function didn't have any inputs, so I'm a little confused what inputs this new `find_files()` expects? A dictionary of data folders? I can't see an example of where this is called at the moment

rt17603:
From looking at the code, it seems that `data_folders` was added to allow this to be more generic. But it would be better if there was a way to wrap this to easily run on BP1 for our use - that's what it was doing by default. It doesn't seem like `data_folders` an obvious and simple simple input either (I can't see any details on the expected inputs for this)?

For the question of the loop, the proposed change does make sense, if we want to keep `data_folders` as a required input. At the very least `data_folders` would have to contain a subset of keys from `all_instrument_details` for that to work. However, we may want to allow `data_folders` to be optional and allow the data folder to be derived from the (or a specified) json file - so this may change how we want to update this function.

gareth-j:
This functionality has been removed for now so I'll close this.

PR #135, opened 2021-09-20
Fix for error raised when running tests if OPENGHG_PATH env var not set, now set in conftest.py

gareth-j:


gareth-j:
CI failed on this due to a cask not being retrieved by homebrew for one of the MacOS runners, all others passed fine.

PR #158, opened 2021-11-18
Adding in co2 fp_data_merge functionality (high time resolution data)

rt17603:
Adding in footprint_data_merge functionality (first pass) to allow data comparison for co2 data.

Simple test added for this.


gareth-j (approved):
Looks good, i've just made a few small changes to a docstring and formatted it with black.

Issue #119, opened 2021-08-26
Add ability to skip ranking to `get_obs_surface`


gareth-j:
Skip ranking needs to the implemented - related to #22

rt17603:
I think the issue here is that `get_obs_surface` needs an inlet even when this is not ambiguous - hence wanting to be able to skip the ranking.

PR #176, opened 2021-11-26
Update get_obs_surface to allow inlet to be skipped when not needed

rt17603:
Based on Issue #119, this PR is to allow the ranking to be skipped when not needed. This means `get_obs_surface` can be called as follows:

`get_obs_surface(site="tac", species="ch4")`

and as long as there are not multiple inlets, the single data source can be successfully retrieved. An explicit input for `skip_ranking` has also been added e.g.

`get_obs_surface(site="tac", species="ch4", skip_ranking=True)`


rt17603:
Need to fix test outcomes that have been changed based on this PR especially:
=========================== short test summary info ============================
FAILED tests/dataobjects/test_searchresults.py::test_retrieve_complex_ranked
======= 1 failed, 145 passed, 7 skipped, 49 warnings in 79.87s (0:01:19) =======

rt17603:
Now fixed tests - but this is fixed in relation to the new logic in this PR. So overall we need to check we're happy with this logic.

tests/retrieve/test_observations.py
```diff
@@ -0,0 +1,12 @@
...
+
+def test_get_obs_surface_one_inlet():
+    '''
+    Test we can access site and species data without needing to specify the inlet
+    when this is not needed.
+    '''
+    # Rely on data already loaded from conftest.py and in local object store
+    data = get_obs_surface(site="tac", species="ch4")
+
+    assert isinstance(data, ObsData)
```
gareth-j, line 12: I've added in a few extra asserts to make sure we get the correct data back here.

gareth-j (approved):
Looks good, I've updated the test.

PR #179, opened 2021-11-29
Changing end_date to be exclusive in get_obs_surface()

### LINKED ISSUE #161: Daterange handling in `get_obs_surface`
`The date range can now be selected but I think we should update this to not include the end_date in our date range for consistency with current functionality elsewhere.`

rt17603:
Changing `get_obs_surface()` to extract data from `start_date` and up to but not including `end_date`. Exclusive end.

There doesn't seem to be a nice neat way to do this using xarray functionality, so this has been implemented by subtracting a small time increment (1 nanosecond) from the end time when cutting down this date range .

Test has been added with continuous Mace Head data to check this selects the date range excluding the end date.

Closes #161


gareth-j (approved):

gareth-j:
Looks good, I think this is ready to merge.

PR #181, opened 2021-11-30
Standardising species name for GC

### LINKED ISSUE #169: GCWERKS returning two different names of chemicals
At the moment `parse_gcwerks` returns data in a dictionary with a key of `toluene_70m` but the variables in the Dataset have the chemical formula `c6h5ch3`, we should make sure this is standardised.

rt17603:
Closes #169

Updated the species name to be the same between the label in the retrieved data and the variable within Dataset.

Uses attributes.json file (as does the attributes function) to look up standard name and use this when creating the species label e.g. "toluene_70m" would now be "c6h5ch3_70m"

Associated tests also updated to align with this new naming scheme.

### DIFF
--- a/tests/localclient/test_get_surface_obs.py
+++ b/tests/localclient/test_get_surface_obs.py
@@ -159,7 +159,9 @@ def test_gcwerks_retrieval():
         "data_type": "timeseries",
     }
 
-    assert metadata == expected_metadata
+    for key, expected_value in expected_metadata.items():
+        value = expected_metadata[key]
+        assert value == expected_value
 
     assert data.time[0] == Timestamp("2018-01-01T02:24:00")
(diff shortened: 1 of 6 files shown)


tests/localclient/test_get_surface_obs.py
```diff
@@ -159,7 +159,9 @@ def test_gcwerks_retrieval():
         "data_type": "timeseries",
     }
 
-    assert metadata == expected_metadata
+    for key, expected_value in expected_metadata.items():
```
gareth-j, line 162: I've updated the expected metadata dict as this loop just checks the dictionary against itself.

gareth-j (approved):
Looks good, I've just updated the assert for the metadata.

PR #199, opened 2022-01-05
Adding metadata and attribute checking functions

gareth-j:
This adds some helper functions to the test suite to ensure the correct keys are found in the metadata and attributes.


gareth-j:
See #201 for what needs updating / checking.

PR #216, opened 2022-01-19
Ability to pass site long name to processing scripts

### LINKED ISSUE #204: Add list of site codes / ability to process using site name

gareth-j:
This allows the passing of names like `cape grim` instead of `CGO` and `angus tower` instead of `TTA`. Slightly mis-spelt words should also be matched as `rapidfuzz` has been used for some simple fuzzy text matching. This should close #204.


gareth-j:
Maybe we can be more strict and make users match the exact string, but if they don't use the fuzzy matching to say "did you mean?"

PR #226, opened 2022-01-26
Updates to first part of the workflow

rt17603:
Updating the first part of our workflow around setting up an object store, adding files and ranking the data. This has now been split into two notebooks and renamed. This should be added to the CI once this has been merged.

As part of this a function has been created to summarise the data types available within our ObsSurface inputs. This is reliant on the different input json files we have available at the moment but this can be expanded / standardised as needed.


gareth-j (approved):
Looks really good!

gareth-j:
I'll merge #233 into this once all the checks pass and then I think we should be ready to merge into `devel`.

Issue #229, opened 2022-01-27
Remove warning about missing keys


gareth-j:
The key missing warning such as

```
WARNING: long_name key not in attributes or metadata
```

Should be changed so it either logs something or doesn't print at all, it results in too much being printed to screen.

rt17603:
So, I added that in the one of the recent PRs - it's printing out the warning when cross referencing the attributes and metadata. How would you want any logging etc. to be done? And we can go through and make sure these keys are present so it doesn't print out the warning? (linked to Issue #201 )

PR #230, opened 2022-01-27
Fix for `long_name` warnings

gareth-j:
`long_name` gets converted to `station_long_name` in https://github.com/openghg/openghg/blob/d4d476707ded1df2822e8f9ee5244ca015c25eca/openghg/standardise/meta/_attributes.py#L307.

This just removes `long_name` from the list of expected keys.


gareth-j:
Closes #229.

Issue #201, opened 2022-01-06
Missing metadata / attributes for standardisation modules


gareth-j:
These modules have missing data that maybe should be present in the metadata / attributes

### Metadata

- [x] BTT - inlet, instrument

### Attributes

- [ ] Cranfield - no attributes
- [x] CRDS - 'sampling_period', 'instrument'
- [x] NOAA - ObsPack flask data -  'instrument'

rt17603:
I'm a little confused about Cranfield and the associated site code - in the "site_lookup.json" file there's a value for "THB" and "TMB" (Thames Barrier) but in the "acrg_site_info.json" file the only site associated with the CRANFIELD network is "TMB" (Thames Barrier).

So should the references to "THB" actually be "TMB"? I don't know anything about data from Cranfield but is this actually just another set of data from the Thames Barrier or is this something different?

rt17603:
Not completed by PR

rt17603:
Response from Daniel on 08/02

*Initial question: Would you mind if I clarified some details with you about how we're receiving the London GHG data? I'm just trying to work out what modules we need for openghg around Cranfield and Thames Barrier versus the other sites*

> in summary: the Cranfield (LICOR instruments) and thames barrier data gets uploaded to a dropbox folder, then I manually download and process them (the licor file format sometimes changes slightly then I need to change the process code in the repo...)

*Question: So if that going to be the ongoing situation - the thames barrier data being supplied by cranfield?*

> I believe so, I don't know how long the thames barrier data will be ongoing. The LICOR data is the LGHG specific instruments, then any other London sites we get access to is on an ad-hoc basis

rt17603:
@gareth-j - for Cranfield can you add a NotImplementedError (or otherwise) so we can come back to this if we get more details.

gareth-j:
Closing, deprecation warnings will be added for the BTT and Cranfield standardisation functions: #770.

Issue #193, opened 2021-12-13
Tidy up some of the NOAA Obspack standardisation


rt17603:
There a few decisions to make when processing the NOAA ObsPack data (noted as TODO comments in the code):

- [ ] What to do with duplicates. In the latest obspack e.g. "obspack_multi-species_1_CCGGSurfaceFlask_v2.0_2021-02-09", these may be genuine repeated measurements and so should be averaged (could be folded into e.g. `ch4_number_of_observations` variable)
- [ ] What to name `value_unc` from input obspack - this seems to be just classed as an "uncertainty" so may be fine to label as e.g. `ch4_variability`
- [x] #219
- [ ] Add any missing sites to "acrg_site_info.json" OR remove need for sites to be in this json file
- [ ] Add test for "obspack_multi-species_1_CCGGObservatoryInsitu_v2.0_2021-02-09" including derived sampling period estimate
- [ ] Add test for "obspack_multi-species_1_CCGGTowerInsitu_v1.0_2018-02-08" including derived sampling period estimate

rt17603:
There is currently a skipped test related to this -
```
@pytest.mark.xfail(reason="Select which ObsPack attributes we want - currently have non JSON serialisable data - see #207")
def test_read_noaa_obspack():
...
```

gareth-j:
As long as the attributes and metadata same keys match and we don't have silly duplicates, we can just keep the attributes.

rt17603:
All points not completed

rt17603:
Created new issue https://github.com/openghg/openghg/issues/277 for "Decide how to handle duplicates in NOAA data" so can remove from this list.

rt17603:
Can close as issues have been removed, opened as new issue or ticked off.

PR #240, opened 2022-02-03
Update NOAA obspack processing (nc) to handle data with multiple inlets

### LINKED ISSUE #219: Which ObsPack attributes to keep (if not all)
Keeping all the attributes is currently creating a bug when trying to run the NOAA tests and adding to the object store. The test for this is currently being skipped:
```
@pytest.mark.xfail(reason="Select which ObsPack attributes we want - currently have non JSON serialisable data - see #207")
def test_read_noaa_obspack():
...
```

### LINKED ISSUE #227: Add processing for multiple heights in NOAA netcdf data
For some input NOAA ObsPack netcdf data this includes a `intake_height` variable with potentially different inlets/heights for each data point. These should be split out before being written into the object store under different labels.

For the files where this is not the case this can still be written with one inlet (currently can try and infer this from "dataset_intake_ht" variable.

rt17603:
Added processing for NOAA netcdf data to handle multiple inlet heights in data

Additional updates
- Added additional sites to database from based on NOAA global data
- For NOAA data also needed to update "scale" keyword to "calibration_scale" - also applied this change throughout to make this the new consistent keyword used for both the metadata and attributes.

New test added for this and other tests updated where needed.

**Main changes are in `_noaa.py` file**. For `_populate.py` a function was moved from the bottom to the top of the file - so this is not a as large a change as this appears in the diff.

Closes #227
Closes #219 
Related to Issues #193 , #201


gareth-j (approved):
Looks good!

Issue #209, opened 2022-01-11
Adding `requests.packages.urllib3` to ignore in `mypy.ini` - check status


gareth-j:
They've removed the `requests.packages.urllib3` from typeshed, see https://github.com/python/typeshed/issues/6054.

This results in the Python 3.7 + 3.8 runners to fail. Will set to `ignore_missing_imports` in `mypy.ini` for now but will check on status again in a week or two.

gareth-j:
See https://pypi.org/project/types-urllib3/

gareth-j:
Done with #535

PR #241, opened 2022-02-03
Adding types-urllib3 for urllib3 type hints and removing mypy ignore

gareth-j:
…from ini


gareth-j:
Fixes #209.

gareth-j:
So the updates in `typeshed` still don't fix these errors so I'll shelve this for now.

Issue #275, opened 2022-03-04
Reintroduce ICOS processing functionality


gareth-j:
We've had a request to reintroduce ICOS processing functionality, I'll pull the functions back in from our history.

gareth-j:
Done with #278

PR #278, opened 2022-03-16
Re-adding ICOS processing ability

gareth-j:
This re-adds the processing abilit we previously had for ICOS data due to a user request.
Still need to:

- [ ] Check the processing function works with Joe's data
- [ ] Add further tests for that data


gareth-j:
This closes #275.

PR #293, opened 2022-03-31
Updating NetCDF output arguments

gareth-j:
I encountered a strange error on macOS which resulted in lots of errors. I've added passing the netcdf type and netcdf output engine to the `to_netcdf` command in Datasource.


gareth-j:
I'm going to close this as it's unecessary.

PR #302, opened 2022-04-04
Updating definition around high time and high spatial footprint resolution

### LINKED ISSUE #294: Fix footprint definitions around high *time* and high *spatial* resolution
The `openghg.store._footprints.Footprints.read_file()` static method has current expectations around "high_res" footprint data files which don't align with our current definitions.

This should be split into:
 - high time resolution (co2 footprints - expect `fp_HiTRes` data variable and `H_back` dimension)
 - high spatial resolution (any footprints which contain `lon_high` and `lat_high` with a higher resolution section e.g. for urban studies)

At the moment if the input `high_res=True` is set, this will expect `lat_high` and `lon_high` to be present and set the metadata to include `metadata["time_resolution"] = "high_time_resolution"` which is not accurate and conflates the two ideas together.

rt17603:
Updating footprint functions to make a distinction between high time resolution and high spatial resolution and the expectations around these inputs.

As well as applying updates to _footprint.py, also did the following:
 - Updated and added new tests for _footprint.py
 - Updated _process.py to incorporate updated keywords

Closes #294

### DIFF
--- a/openghg/client/_process.py
+++ b/openghg/client/_process.py
@@ -100,8 +100,9 @@ def process_footprint(
     species: Optional[str] = None,
     network: Optional[str] = None,
     retrieve_met: bool = False,
+    high_spatial_res: bool = False,
+    high_time_res: bool = False,
     overwrite: bool = False,
-    high_res: bool = False,
 ) -> Dict:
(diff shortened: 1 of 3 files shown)


gareth-j (approved):
Looks good!

PR #306, opened 2022-04-11
Fixes to `Datasource` daterange overlap and updates to `SearchResults` retrieval

gareth-j:
This fixes:

- A bug in `Datasource` where overlapping dateranges weren't handled correctly
- An error in the `SearchResults.retrieve` function that resulted in a `KeyError` when called with certain arguments

I've added further tests for both the Datasource and the SearchResults objects to try and make sure they're a bit more robust.


openghg/dataobjects/_searchresults.py
```diff
@@ -200,79 +217,49 @@ def retrieve(
...
+                    for _inlet in species_data:
+                        if inlet is not None and _inlet != inlet:
+                            continue
+
+                        obsdata = self._create_obsdata(site=_site, species=_species, inlet=_inlet)
+
+                        if obsdata is not None:
+                            results.append(obsdata)
+        else:
+            if inlet is not None:
```
rt17603, line 237: I may be mis-interpreting this but if someone passes an inlet and the data is ranked can we just bypass the ranking to return that specific inlet? I'm just thinking you may want to retrieve data from a particular inlet regardless of ranking e.g. for comparison between inlets.

openghg/store/base/_datasource.py
```diff
@@ -123,36 +123,57 @@ def add_timeseries_data(self, data: Dataset) -> None:
...
         """
         from openghg.util import daterange_overlap
+        from xarray import concat as xr_concat
+        from numpy import unique as np_unique
 
-        # Group by year
-        year_group = data.groupby("time.year")
-        year_data = [data for _, data in year_group if data]
+        # Chunk the data into year datasets by year
+        grouped_by_year = [data for _, data in data.groupby("time.year") if data]
```
rt17603, line 130: There's likely to be some conflicts with the `add_bc` PR changes around here after it's merged into devel

rt17603: Ok - yes there's a few conflicts to fix here

PR #316, opened 2022-04-20
Adding community related files / docs

gareth-j:
- Updating code of conduct,
- Updated python devel pre-commit section of docs,
- Added PR template

### DIFF
--- a/.github/ISSUE_TEMPLATE/README.md
+++ b/.github/ISSUE_TEMPLATE/README.md
@@ -1,0 +1,1 @@
+Thanks to the xarray devs for these issue templates.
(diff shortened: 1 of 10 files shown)


rt17603:
Out of interest, is the 'doc/source/whatsnew/vX.X.X.rst' different to a 'CHANGELOG.md' file? Or it essentially the same thing?

I just realised I haven't been keeping track of updated features and I guess we should if we're releasing versions now? Should that be added to the whatsnew version file?

rt17603 (approved):
All looks good to me (PR template seems like a great idea).

gareth-j:
> 

Yeah so I see it as a changelog is all the little changes we make, like bug fixes etc and release notes are for users that want to check what new features there are, things we've fixed but don't want to see it all in detail. Maybe we could try doing both and see how it works for us?

PR #320, opened 2022-04-27
Incorporating boundary conditions into ModelScenario functionality

rt17603:
Adding option to add boundary conditions to ModelScenario objects. This includes:
 - `add_bc` method which uses `openghg.retrieve.get_bc` in the background
 - `calc_modelled_baseline` to calculate the baseline values using the boundary conditions and footprints
    - This also incorporates the short-lived species options adapted from the acrg code
    - Updated "acrg_species_info.json" to incorporate this new information as well though updated some of the keywords here to be more clear ("lifetime_monthly" for 12 monthly lifetime values as a list rather than "lifetime")
 - `plot_comparison` incorporates "boundary_conditions" as an option when including a baseline which uses the `calc_modelled_baseline` method.

In the background this also includes:
 - adding a `get_bc()` function to `openghg.retrieve` submodule which uses new `BoundaryConditionsData` object

Added appropriate tests:
 - Incorporates functionality into some existing tests
 - Added new tests around `calc_modelled_baseline` method
 - Added new tests for short-lived species (dummy data only so far)

All tests pass.


gareth-j (approved):
Looks good!

rt17603:
Just FYI, running `black` over the `openghg.analyse._scenario.py` broke the `mypy` fixes. When the `loss_n` etc. definitions became split over multiple lines this seems to have nullified the necessary `# type: ignore` which were included. There is a comment in the code which explains why this was included but in short this is because xarray hasn't properly implemented an overload of the `np.exp` function so mypy insists this should be a `np.array` object when this will be a `xr.DataArray` object.

PR #328, opened 2022-05-03
Adding AGAGE / NOAA comparison to tutorials (and fixing some associated bugs)

rt17603:
Based on a request to provide a video comparing NOAA/AGAGE data I have updated the tutorials to incorporate this and also had to fix some additional bugs for this to function correctly.

Tutorials:
 - Switched over Tutorial 5 (Explore NOAA ObsPack) and 6 (Searching and Plotting) so that the `plot_timeseries` function could be used in the NOAA ObsPack tutorial.
 - Updated NOAA ObsPack tutorial (now 6) to include a comparison with AGAGE data
 - Small updates to Search and Plotting (now 5) tutorial but nothing significant

Bugs fixed:
 - `add_NOAA_obspack` function was not working with the retrieved data and was producing an error when trying to concatenate multiple data entries.
    - Cause: This was due to some variables (such as `ch4_number_of_observations`) not being present in all relevant data files.
    - Fix: Updated this to check for missing variables, add in the values before concatenating and populate with NaNs
 - Data retrieved for NOAA using `get_obs_surface` did not include sensible unit values. 
   - Cause: These were being converted in the obspack processing code but this then meant the correct quantities weren't extracted when looking at the `attributes.json` file.
   - Fix: Removed the conversion from the obspack code and added more options to the `attributes.json` file
- Similarly `plot_timeseries` didn't account for different units being plotted when comparing between networks
   - Cause: NOAA data is in mol mol-1 units (1) whereas AGAGE data has been converted to nmol mol-1 (ppb, 1e-9)
   - Fix: Added checker and converter to the `plot_timeseries` code which makes sure data matches either to input `units` if specified and the units of the first `ObsSurface` object passed to the function otherwise.
- The `get_obs_surface` function was producing an error when multiple data sources were found (related to line `data = retrieved_data.data` and this stated that a "list object does not have a data attribute").
   - Cause: `SearchResults.retrieve()` method has been updated to allow lists of ObsData to be returned (the type was being ignored from this function). This was being called within `get_obs_surface`.
   - Fix: Check whether returned data is a list and print relevant differences in the metadata. This currently returns None but we can update this to an error if needed.
 - (originally on `test_db_integrity` branch) Different daterange being defined when saving keys for new data and overlapping data.
   - Cause: Only one part of the code had been updated to use new time extent definition
   - Fix: Updated code which dealt with overlapping data as part of `add_timed_data` to use new representative date range when creating the key (`get_representative_daterange_str`)


openghg/retrieve/_access.py
```diff
@@ -67,14 +67,19 @@ def get_obs_surface(
...
         raise ValueError(f"Unable to find results for {species} at {site}")
 
-    retrieved_data: Union[ObsData, None] = obs_results.retrieve(site=site, species=species, inlet=inlet)  # type: ignore
-    
+    retrieved_data: Union[ObsData, List[ObsData], None] = obs_results.retrieve(site=site, species=species, inlet=inlet)
+
     if retrieved_data is None:
         return None
+    elif isinstance(retrieved_data, list):
+        print("No data returned.")
```
gareth-j, line 75: Is this supposed to be above?

openghg/retrieve/_access.py
```diff
@@ -481,3 +485,52 @@ def _scale_convert(data: Dataset, species: str, to_scale: str) -> Dataset:
...
+
+    Args:
+        data : Multiple data objects e.g. multiple ObsData as a list
+        params : Specific metadata parameters to check. If None all parameters will be checked
+        print_output : Summarise and print output to screen.
+
+    Returns:
+        list : Keys from the metadata with differences
+
+        If print_output is True:
```
gareth-j, line 507: I think this could be moved to the description section of the docstring instead of the Returns section.

openghg/store/base/_datasource.py
```diff
@@ -283,6 +267,74 @@ def get_dataset_daterange_str(self, dataset: Dataset) -> str:
 
         return daterange_str
 
+    def get_representative_daterange_str(self, dataset: Dataset, period: Optional[str] = None) -> str:
+        """
+        Get representative daterange which incorporates any period the data covers.
```
gareth-j, line 272: This could do with a full docstring.

openghg/store/base/_datasource.py
```diff
@@ -283,6 +267,74 @@ def get_dataset_daterange_str(self, dataset: Dataset) -> str:
...
+
+    def get_period(self) -> Optional[str]:
+        """
+        Extract period value from metadata. This expects keywords of either "sampling_period" (observation data) or
+        "time_period" (derived or ancillary data). If neither keyword is found, None is returned.
+
+        Returns:
+            str : time period in the form of number and time unit e.g. "12s"
+
+            This is a suitable format to use to create a pandas Timedelta or DataOffset object.
```
gareth-j, line 304: Another pedantic comment.
Maybe this could be in the description section.

gareth-j (changes requested):
Looks great, just a few small changes.

openghg/retrieve/_access.py
rt17603: No - this was implementing the logic where we only expect `get_obs_surface()` to return **one Dataset**. This is the first line of a set of lines that are printed out to explain why no data was returned and what extra details to include to get this to return something.

openghg/store/base/_datasource.py
rt17603: Ok I've added more details for this one.

rt17603:
@gareth-j - are you happy with the updates made for Changes Requested? And if so, happy for this to be merged?

openghg/retrieve/_access.py
gareth-j: Ah OK yes, makes sense now.

gareth-j (approved):

gareth-j:
> @gareth-j - are you happy with the updates made for Changes Requested? And if so, happy for this to be merged?

Yeah looks good.

PR #339, opened 2022-05-27
Adding preprocessed data type

### LINKED ISSUE #276: Add new data type for pre-processed netcdf observation data
If data has already been processed to match our stored observation data format, we should allow this to be added to the object store as is. For this the easiest thing to do would probably be to create new data_type e.g. "netcdf" or "standardised" which allows for this scenario and can apply a few basic checks.

rt17603:
Adding additional data type option for ObsSurface for data which is already in OpenGHG format.

Current limitations:
 - necessary metadata must be included in netcdf file attributes OR passed in as additional arguments in a manual way
 - site must already exist in our acrg_site_info.json file.

This is really a first pass and the interface could definitely improved - open to suggestions on this.

Closes #276


gareth-j (changes requested):
Looks good, just need a docstring and it should be ready for merging.

PR #373, opened 2022-07-27
Updating behaviour of `clean_string` to allow dashes

rt17603:
Updating `openghg.util.clean_string()` function to allow `"-"` within strings (as well as `"_"` etc.). This is because dashes are often used in the acrg descriptions as a separator to allow for more descriptive names.

Updated associated tests for this as well.


gareth-j:
So it looks like the xarray devs are expanding typing support (see https://docs.xarray.dev/en/stable/whats-new.html). For now we can pin the version of xarray to `xarray==2022.3.0`.

gareth-j:
I've opened #374 for this.

PR #376, opened 2022-07-28
`mypy` fixes for new xarray release

gareth-j:
This (thanks @rt17603) contains fixes / additions to our type hints to handle the new type hinting the xarray devs have added in their latest release.


gareth-j (commented):
I think all this looks sensible. Maybe we can remove some of the casts in future versions if xarray update their type hints.

rt17603:
I've added some fixes for the tests which were failing but there are 2 tests failing due to the `NotImplementedError` for the metarchive. @gareth-j - would you be able to have a look at this?

PR #378, opened 2022-07-28
Adding basic logging

### LINKED ISSUE #358: Local print or log for cloud
We need a logging function that just prints to screen if it's being run locally or logs the message if being run in the cloud. The the log of all the processing messags can be returned to the user.

rt17603:
Adding basic logging using python `logging` built-in module.

Creates logger at the module level which can then be propagated down to the sub-modules.
 - Implemented for openghg.store._obssurface.py so far as an examplar

**TODO:** `tests/retrieve/icos/test_icos_retrieve.py::test_icos_retrieve_and_store` currently fails since this is checking the captured output for a phrase which has now been written to the log file. Need to update this and all consider where the log file should be created.

Closes #358


rt17603:
@gareth-j - thoughts on where we could set up the "openghg.log" file to be created locally and / or on the cloud (and how to best implement this) would be appreciated.

gareth-j (commented):
Looks good! I've just added a bit of code to `__init__.py` to control where the logs are written to.

PR #386, opened 2022-08-15
Improving ICOS and CEDA pulling

gareth-j:
This fixes some issues regarding pulling data from ICOS and CEDA.


gareth-j:
This is superseded by #387.

PR #388, opened 2022-08-26
Removing previous fp workflow

rt17603:
Removing previous `analyse/footprints.py` workflow since this has now been replaced by `analyse/scenario.py`


rt17603:
Would be great to get this in before we start in earnest on Wednesday.

gareth-j (approved):

PR #406, opened 2022-09-01
Test database integrity

rt17603:
Adding tests to check the database integrity for repeat entries (and combining data). Tests included so far:

- Test adding the same data twice (flux)
- Test adding same inputs but small change to header (different hash but data will be the same)
- Test adding same inputs but small change to data values (data itself should be updated)
- Test same inputs but hourly and minutely data added (should be able to access both)

Closes Issue #408 but we should continue to add to these tests going forward as well.


rt17603:
See #408

gareth-j (approved):
Looks good.

Issue #411, opened 2022-09-01
Add a metadata store when SearchResults is returned


gareth-j:
At the moment when we want to retrieve data using the `retrieve` method we can only pass in the site, species and inlet, we can't be more specific. If we add a small metadata store with created from the returned results we can make the search far more flexible. This will also allow us to use the `SearchResults` object when returning flux, emissions etc.

gareth-j:
Instead of adding a metadata store I've added a function that searches through the metadata. As we won't have THAT much metadata I think that's fine. See #414.

rt17603:
Was this closed by #414 or #418?

Issue #405, opened 2022-08-31
Search fails if rank is set but not for the site searched for


gareth-j:
We should check if the site we're looking for has ranking set, at the moment if `ObsSurface` has ANY ranking data or if there's more than one Datasource available.

So there was ranking data for BSD, then loaded in some HFD data, search for HFD data, get error saying no ranking data set.

gareth-j:
This is fixed with #414.

Issue #366, opened 2022-07-19
Move footprint, flux, bc results to SearchResults


gareth-j:
At the moment surface measurements are treated differently to the other types by `search`. This means we end up with a `dict` when we should get a `SearchResults` object.

Issue #362, opened 2022-07-14
Update search function to use new metadata storage


gareth-j:
We need to update the search function so it uses the new metadata storage databases.

gareth-j:
Fix this after fixing the ranking issue at #405

PR #414, opened 2022-09-15
Tidying the search function

gareth-j:
This massively simplifies the `search` function. It now:

- Uses the metadata store
- Checks dates only if a date is passed in
- Performs no ranking search

### Limitations

- Can still only search for a single data type, do we want to support searching across multiple data types?

### Closes

- #282 - *does this close this one, or does the original filename need to be added?*
- #362 
- #366
- #405 
- #422


gareth-j:
mypy fails on the 3.10 runner due to a bug in mypy, see https://github.com/python/mypy/issues/13627.

gareth-j:
This will close #411.

openghg/retrieve/_search.py
```diff
@@ -102,7 +156,9 @@ def search_surface(
         start_date=start_date,
         end_date=end_date,
         data_source=data_source,
-        **kwargs,
+        network=network,
+        sampling_height=sampling_height,
+        # **kwargs,
     )
 
     # TODO - remove this cast once we've updated search to ensure return of SearchResults object
```
rt17603, line 164: Just noticed this comment - can the `cast()` be removed now?

openghg/dataobjects/_searchresults.py
```diff
@@ -95,290 +88,107 @@ def from_json(cls: Type[T], data: Union[bytes, str]) -> T:
...
-            return None
+        return cls(keys=loaded["keys"], metadata=loaded["metadata"])
 
-    def metadata(self, site: str, species: str, inlet: Optional[str] = None) -> Optional[Dict]:
-        """Return the metadata for the specified site and species
+    def retrieve(
+        self,
+        dataframe: Optional[DataFrame] = None,
+        sort: bool = True,
+        elevate_inlet: bool = False,
```
rt17603, line 97: Would be useful to have an explanation of additional inputs for the `retrieve()` function in the doc string - e.g. What does `elevate_inlet` mean?

rt17603 (approved):
I've added a few small comments of details to add, so would be nice if they could be updated, but other than that looks good. Lets get this in and then we can concentrate on the other PRs and issues.

openghg/dataobjects/_searchresults.py
gareth-j: Done

openghg/retrieve/_search.py
gareth-j: Done.

PR #417, opened 2022-09-26
A very basic implementation of a user configuration file setup

gareth-j:
This implenents a basic user config file setup. It writes a TOML config file to `~/.config/openghg/openghg.conf`. I still need to add some documentation covering editing this etc.


gareth-j:
This is another failing due to the mypy bug but can be merged if it looks OK.

Issue #401, opened 2022-08-31
Improve search - multiple types


gareth-j:
- Return a pandas DataFrame
- Search each data type
- Search only metadata store instead of loading all the Datasources

gareth-j:
Working on this with #418.

rt17603:
This could be called through a generic `search()` function which accepts `data_type` as an input. If no `data_type` input is supplied this will search using the keywords for all data types and return a pandas DataFrame with "data_type" as the first column.

gareth-j:
Done with #414 and #418

PR #418, opened 2022-09-27
Multi-type search capabilities

gareth-j:
This adds the ability to search across multiple types in a single search. Unless the user passes in a data type during search we loop over each of the storage objects and search their metadata stores. This is all then combined into a single SearchResults object.

I still need to add a few tests to make sure everything is setup correctly and retrieval works.


rt17603:
See comments on Issue #401 and #426 around this as well

Issue #416, opened 2022-09-22
`get_` functions, do we still need them? Search enough?


gareth-j:
Do we still need the get_* functions? Should we point people towards the search functions instead?

PR #444, opened 2022-10-07
Harmonising get functions to produce same errors

### LINKED ISSUE #297: Update `get_` functions to all raise a ValueError when no data can be found / too much data is found
Update `get_` functions to all raise a ValueError when no data can be found / too much data is found
    - If no data is found, make sure this (and keywords asked for) is clear (could include search command details here to assist?)
    - If > 1 set of data is found and ranking IS NOT present / used, print message to suggest adding ranking or being more specific
    - If > 1 set of data is found and ranking IS present / used, print message to suggest improving ranking details

rt17603:
Ensuring behaviour for openghg.retrieve.get_* functions is consistent and adding custom `SearchError` function for openghg
 - If no results found returns a `SearchError` with helpful message (restates keys which were passed in)
 - If multiple results are found returns `SearchError` with helpful message (compares metadata and prints differences)
 - All functions (including get_obs_surface_local) are now fed through _get_generic function

`openghg.analyse._scenario` has also been updated to account for this behaviour (interface and output from ModelScenario class and methods should be the same).

Closes #297, #416

### DIFF
--- a/openghg/exceptions/__init__.py
+++ b/openghg/exceptions/__init__.py
@@ -1,8 +1,0 @@
-
-
-class OpenGHGError(Exception):
-    pass
-
-
-class SearchError(OpenGHGError):
-    pass
--- a/openghg/exceptions/__init__.py
+++ b/openghg/exceptions/__init__.py
@@ -1,0 +1,6 @@
+
+class OpenGHGError(Exception):
+    pass
+
+class SearchError(OpenGHGError):
+    pass
--- a/openghg/exceptions/__init__.py
+++ b/openghg/exceptions/__init__.py
@@ -1,6 +1,8 @@
 
+
 class OpenGHGError(Exception):
     pass
+
 
 class SearchError(OpenGHGError):
     pass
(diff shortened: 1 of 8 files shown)


rt17603:
@gareth-j - not sure about this build documentation error? Seems to be unable to find a some of the sphinx documentation modules?

```
ModuleNotFoundError: No module named 'sphinx_togglebutton'

make: *** [Makefile:34: html] Error 2
Error: Process completed with exit code 2. 
```

openghg/exceptions/__init__.py
```diff
@@ -0,0 +1,8 @@
+
+
+class OpenGHGError(Exception):
```
gareth-j, line 3: These could be moved to `openghg.types` as we've already got some custom exceptions there.

gareth-j (changes requested):
Looks good, maybe we can just move the new exceptions to `openghg.types`.

gareth-j:
I think this looks good and can be merged.

rt17603:
Great, would you be able to give this an approve? Then I'll get it merged.

gareth-j (approved):
Looks good.

PR #471, opened 2022-10-28
Adding inlet as keyword and allowing height as an alias for both footprints and obs_surface

### LINKED ISSUE #466: Make height and inlet consistent for footprint data
For now, we can add both height and inlet to footprint definition.

rt17603:
At the moment, we have an input of `inlet` for ObsSurface data and `height` for Footprint data when they should represent the same property (and most of the time will be equivalent). 

This update makes the following changes:
 - Adds "inlet" as an additional input and metadata keyword within the `Footprint.read_file()` method
     -  This still retains "height" for backwards compatability but this should always be the same as the `inlet` value
     - Adds "inlet" to both `search_footprints` and `get_footprint` functions. Either inlet or height can be used to search.
 - Also allows "height" as an alias input for "inlet" within `ObsSurface.read_file`
     -  "height" is not added to the metadata here as this is not needed for backwards compatability
     - Adds "height" to both `search_surface` and `get_obs_surface` functions. Either "inlet" or "height" can be used to search (both will link to inlet metadata keyword)
 - Added new util function around inlets to try and apply additional formatting rules to this - should be 'NUM''UNIT' e.g. "20m" not "20magl" or "20". Incorporated into relevant functions mentioned above.

Additional tests added to check this.

Closes #466

### DIFF
--- a/openghg/analyse/_scenario.py
+++ b/openghg/analyse/_scenario.py
@@ -287,8 +287,8 @@ class ModelScenario:
 
             if height is not None and inlet is None:
                 inlet = height
-            inlet = format_inlet(inlet)
             inlet = clean_string(inlet)
+            inlet = format_inlet(inlet)
 
             # search for obs based on suitable keywords - site, species, inlet
             obs_keywords = {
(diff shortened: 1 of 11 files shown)


gareth-j (approved):
Looks good!

Issue #469, opened 2022-10-27
ModelScenario thinks it finds two footprint sources when they're the same


gareth-j:
Why are two footprints that should be the same datasource being treated as two different Datasources?
Is it due to the date being a piece of metadata and then we're differentiating between the two using this?

```
Adding obs_surface to model scenario
Updating any inputs based on observation data
site: wao, species: co2, inlet: 20m
Adding footprint to model scenario
date_created: '2022-05-05 14:59:05.907679', '2022-05-06 02:54:56.308504',
date: '2017', '2018',
end_date: '2017-12-31 23:59:59+00:00', '2018-12-31 23:59:59+00:00',
start_date: '2017-01-01 00:00:00+00:00', '2018-01-01 00:00:00+00:00',
processed: '2022-10-26 10:36:49.665339+00:00', '2022-10-26 10:47:32.046189+00:00',
ERROR:openghg.retrieve:
Multiple entries found for input parameters for species='co2', source='edgar-ukghg-ff', domain='EUROPE', data_type='emissions'.
Parameter differences:
     - date_created: 2022-05-05 14:59:05.907679, 2022-05-06 02:54:56.308504
date: 2017, 2018
end_date: 2017-12-31 23:59:59+00:00, 2018-12-31 23:59:59+00:00
start_date: 2017-01-01 00:00:00+00:00, 2018-01-01 00:00:00+00:00
processed: 2022-10-26 10:36:49.665339+00:00, 2022-10-26 10:47:32.046189+00:00
 Please supply additional parameters or set ranking.
```

rt17603:
I believe that error is actually stating it's found multiple *flux* sources not footprint sources. It's a different issue around the footprints (sort of related).

rt17603:
For flux (and boundary conditions):
 - The issue seems to be that the 'date' metadata keyword is used to classify the data sources and so multiple flux files with different states were not being associated. This means when retrieving the data these are seen as separate.
    - Fix Part 1: When we merge in PR #468, the fluxes which have been stored so far will need to be removed.
    - Fix Part 2: When we merge in PR #475, this keyword will no longer be used and will avoid this issue. The flux files can then be added again to the object store.

rt17603:
Will close this when PR #468 has been merged

PR #472, opened 2022-10-31
Fixing issue with 'not' finding multiple footprint sources.

rt17603:
Added tests to reproduce the issue of not finding the full range of footprint sources when these are separate files related to the same metadata. The tests found the data *was being found* but was misleading.

Added:
 - `test_search.py::test_search_footprints_multiple` along with appropriate files within the setup steps in `conftest.py`
    - Data is added with the same metadata for two footprint files containing data from 2016/07 and 2016/08.
    - Test extracts all data with those keywords for whole range
    - This is successful but the `start_date` and `end_date` metadata was created when adding the first footprint file and is not updated with subsequent files (reasonable behaviour). In previous code these values were `start_date:"2016-07-01T00:00:00"` and `end_date:"2016-07-31T23:59:59"` which was misleading.
 - `test_search.py::test_search_footprints_select`
   - This checks a subset can be selected from the same sets of files.

Updates:
 - Removed `start_date` and `end_date` keywords from metadata as this is misleading and not needed for underlying data.
    - Could update attributes within each stored netcdf file / dataset to contain these values as each file is added.
 
*Alternatively, we could update this to always check against stored data and update `start_date` and `end_date` accordingly BUT this is not in line with `ObsSurface` behaviour anyway.*

Links to #469 but does not close it.


rt17603:
The plan is to not merge this PR but to extract out any useful details (noteably the footprint tests) and incorporate into a new PR which addresses Issue #476 .

PR #480, opened 2022-11-08
Fixing issue when synchronising metadata and attributes for "openghg" source_format

### LINKED ISSUE #477: ValueError when using `standardise_surface` with inlet value
> For standardise.standardise_surface I'm now getting an error: ValueError: could not convert string to float: '185m'
I've put inlet = 185 so I think it must be changing it in the code to '185m' by default (we asked for it to do this for the other inputs so that it always uses 'm' instead of 'magl')

### LINKED ISSUE #478: search_surface can't find results if an inlet is given.
> Related to this, search_surface can't find results if an inlet is given.

> This code:
> search_results = search_surface(site='TAC',
                 species='co2',
                 inlet='185',
                 start_date='2018-01-01',
                 end_date='2018-02-01'
                 )

> Gives this warning:
> WARNING:openghg.retrieve:No data found for the dates given, please try a wider search.

> Using the code above without a start_date and end_date does not return a warning but also does not return any results.

_Originally posted by @hanchawn in https://github.com/openghg/openghg/issues/477#issuecomment-1307248897_

rt17603:
Small bug around synchronising the metadata and pre-exisiting attributes for netcdf input files (openghg source_format) for obs surface. Issue was related to "inlet" and "inlet_height_magl" definitions and when the attribute itself is stored as a float.

Closes #477 
Closes #478


gareth-j (approved):
Looks good!

Issue #330, opened 2022-05-09
Tutorial sections


gareth-j:
I think we could move the tutorials into sections something like:

- Standardisation
- Retrieval
- Plotting
- Comparison

rt17603:
I think we want to make sure our tutorials are descriptive (to draw people in a bit). We can split into these kinds of sections but think we should still focus on useful workflows (as we have been aiming for so far).

rt17603:
Proposed categories for tutorial structure in notebooks/local/
 - Quick start
   - NEW
 - Adding new data to the database / object store
   - Explain standardise, source formats for obs - 1_Adding_observation_data.ipynb
   - More data types - (Some of 3_Comparing_with_emissions.ipynb) - visualising 2D data types - NEW
   - Transformation functions - NEW
 - Exploring data
   - Searching and plotting - 5_Searching_and_plotting.ipynb
   - ICOS retrieval / CEDA retrieval - 7_Retrieving_remote_data.ipynb
   - NOAA ObsPack - 6_Explore_NOAA_ObsPack.ipynb
 - Analysing data
   - Footprints data merge - 3_Comparing_with_emissions.ipynb
   - CO2 data (/ short lived species) - 4_Working_with_co2.ipynb
   - Add in example with short lived species - NEW
 - Managing the object store
   - Setting up a group object store - NEW
   - Updating data - Modifying_and_deleting_data.ipynb
 - Contributing to openghg
   - Adding new standardisation functions - NEW
   - ... tranformation... functions - NEW

rt17603:
Plan:
1. Start by reorganising and renaming - ASAP
2. Adding new tutorials
 - Quick start - Gareth
 - Data types other than obs - Gareth
 - Transformation - Rachel
 - Short lived species - Rachel
 - DIFF BRANCH - Adding new stanardisation functions - Gareth
 - DIFF BRANCH - Adding new transformation functions - Rachel
3. Editing current tutorials

rt17603:
Would be best to merge this with Issue #363 in some way.

PR #486, opened 2022-11-15
Updating tutorial structure

rt17603:
- Restructuring the tutorials into categories rather than the current numbering system.

See Issue #330


gareth-j (approved):
Looks good, I think this is ready to merge.

PR #519, opened 2022-12-15
Separating info data into new repository

### LINKED ISSUE #515: supplementary_data dependency
### Feature that's missing

site_info, species_info and domain_info should now be called from the supplementary_data repo.  these should be removed from openghg/data

_No response_

_No response_

### Additional context

_No response_

### Additional code / helper functions

_No response_

rt17603:
A new repository has been created within the OpenGHG organisation to split out data details: [https://github.com/openghg/supplementary_data](https://github.com/openghg/supplementary_data)

This includes:
 - site_info.json
 - acrg_species_info.json (renamed to species_info.json in new repository)
 - domain_info.json
 - domain/*

The new repository will be a dependency of OpenGHG:
 - Direct link to install through pip from GitHub added for now but can updated if this becomes available on PyPi etc.

Closes #515


rt17603:
The `test_retrieve_icos::test_icos_retrieve_and_store` test is currently failing with the new data set up. This is not because the data itself has changed but because there is now an "ICOS" network (previously just "EUROCOM") in the `site_info.json` file which wasn't there previously. This is likely being accessed for comparison in this test and is now failing due to a data mismatch ("station_long_name", "height_station_masl"). 

Need to check over this data and the comparison to ensure this all aligns and makes sense as a check. There is also a potential discrepancy in the WAO data - see [https://github.com/openghg/supplementary_data/pull/1](https://github.com/openghg/supplementary_data/pull/1)

rt17603:
> The `test_retrieve_icos::test_icos_retrieve_and_store` test is currently failing with the new data set up. This is not because the values themselves have changed but because there is now an "ICOS" network (previously just "EUROCOM") in the `site_info.json` file which wasn't there previously. This is likely being accessed for comparison in this test and is now failing due to a data mismatch ("station_long_name", "height_station_masl").
> 
> Need to check over this data and the comparison to ensure this all aligns and makes sense as a check. There is also a potential discrepancy in the WAO data - see [openghg/supplementary_data#1](https://github.com/openghg/supplementary_data/pull/1). May want to change expectations around the source_format / network etc. to realign this.

An option has now been included within the code to allow the attribute value to supercede metadata when there is a class using an input flag. The now allows this test to pass despite inconsistent data. This option was mainly included for cases of retrieving data such as ICOS where our local data may be in small conflict with retrieved data. If this is the case is warning is added to the log.

openghg/util/_domain.py
```diff
@@ -3,21 +3,49 @@
...
+    Returns:
+        dict: Data from domain JSON file
+    """
+    from openghg_defs import domain_info_file
+    from openghg.util import load_json
+
+    if domain_filename is None:
+        domain_info_json = load_json(domain_info_file)
+    else:
+        domain_info_json = load_json(domain_info_file)
```
gareth-j, line 29: Should this be `load_json(domain_filename)` here?

openghg/util/_domain.py
```diff
@@ -3,21 +3,49 @@
...
+        domain: Pre-defined domain name
+        domain_filename: Alternative domain info file. Defaults to openghg_defs input.
 
     Returns:
         array, array : Latitude and longitude values for the domain in degrees.
     """
-    from openghg.util import load_json
 
-    domain_info = load_json(filename="domain_info.json")
+    domain_info = get_domain_info(domain_filename)
```
gareth-j, line 48: Add in keyword before argument

openghg/util/_species.py
```diff
@@ -25,7 +51,7 @@ def synonyms(species: str, lower: bool = True, allow_new_species: bool = True) -
     and changed if needed.
     """
     # Load in the species data
-    species_data = load_json(filename="acrg_species_info.json")
+    species_data = get_species_info(species_filename)
```
gareth-j, line 54: Add keyword here

openghg/util/_species.py
```diff
@@ -59,27 +85,26 @@ def synonyms(species: str, lower: bool = True, allow_new_species: bool = True) -
...
 
     Args:
         species : Species name e.g. "ch4" or "co2"
+        species_filename: Alternative species info file. Defaults to openghg_defs input.
 
     Returns:
         str / list / None : Extracted lifetime or None is no lifetime was present.
     """
-    species_info = load_json(filename="acrg_species_info.json")
+    species_data = get_species_info(species_filename)
```
gareth-j, line 103: Add keyword

openghg/util/_species.py
```diff
@@ -120,23 +145,20 @@ def check_lifetime_monthly(lifetime: LifetimeType) -> bool:
...
     Args:
-        species :
+        species : Species name
+        species_filename: Alternative species info file. Defaults to openghg_defs input.
 
     Returns:
         float : Molar mass of species
     """
-    species_info = load_json(filename="acrg_species_info.json")
+    species_data = get_species_info(species_filename)
```
gareth-j, line 159: Add keyword

gareth-j (changes requested):
Looks good, just a few small changes needed then I think it's ready to merge.

openghg/util/_domain.py
rt17603: As this only has one input (`domain_filename`) is this necessary do you think? And also wouldn't that lock us into that variable name when we don't necessarily need to?

openghg/retrieve/icos/_retrieve.py
```diff
@@ -328,21 +346,21 @@ def _retrieve_remote(
...
-        metadata["inlet"] = f"{int(float(_sampling_height))}m"
+        metadata["inlet"] = format_inlet(_sampling_height, key_name="inlet")
+        metadata["inlet_height_magl"] = format_inlet(_sampling_height, key_name="inlet_height_magl")
 
         loc_data = station_data["location"]
         metadata["station_long_name"] = loc_data["label"]
         metadata["station_latitude"] = str(loc_data["lat"])
         metadata["station_longitude"] = str(loc_data["lon"])
-        metadata["station_altitude"] = f"{int(float(loc_data['alt']))}m"
+        metadata["station_altitude"] = format_inlet(loc_data['alt'], key_name="station_altitude")
```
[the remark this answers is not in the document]
rt17603: And again, for the first input does this need to use a keyword? Again, I think we're locking in the `inlet` variable name when we don't necessarily want to do this?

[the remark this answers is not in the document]
rt17603: We need to decide on a style guide for this sort of thing really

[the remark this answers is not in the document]
rt17603: @gareth-j - any thoughts on this? Is there a style guide we're following which asks for `inlet=` etc.? And if not, is there a reason to include this?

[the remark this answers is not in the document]
gareth-j: Ah sorry, I didn't really think when adding these notes, if we only have a single argument I think it's fine not using the keyword. But if we do want to change the keyword name then it's easy to refactor even just with VSCode doing a search and replace. I'd the keyword and think we should for functions with more than one argument.

gareth-j:
I think this is ready for merge.

gareth-j (approved):

PR #524, opened 2023-01-10
Ensure inlet is formatted consistently for all standardisation functions

### LINKED ISSUE #321: How to handle sampling height / inlet for retrieved ICOS data
ICOS handles sampling_height as a float, we usually use an an string such as `85m`. At the moment I'm converting their float to our format, does this sound sensible?

### LINKED ISSUE #481: Make sure inlet_height_magl is stored consistently (as a number)
Make sure for all source_formats for ObsSurface that the "inlet_height_magl" is stored consistently. This should be stored as a string (or number if necessary) without a unit.

At the moment this is sometimes stored as e.g. "10m" rather than "10" (or 10) (e.g. for GCWERKS or CRDS formats)

rt17603:
Relies on PR #519 

For `standardise/surface` parse functions, make sure that these all use the same `format_inlet` function for at least `"inlet"` and `"inlet_height_magl"` metadata so this is consistent between source formats.

 - `"inlet"` e.g. "10m" or "10.1m" - string containing number then unit, 0dp or 1dp for inlet height
 - `"inlet_height_magl"` e.g. "10" or "10.1" - string containing just the number (unit is in the key name).
     - *Could have this as just a float / int but not sure if this is consistent with current format*

Closes #481
Closes #321


gareth-j (approved):
Looks good!

PR #525, opened 2023-01-17
Adding in check for 0-dimension time coordinate

### LINKED ISSUE #349: Boundary condition file time error
When trying to process `openghg/tests/data/boundary_conditions/co2_EUROPE_201407.nc` I get an error `TypeError: len() of unsized object`. This can be fixed with some try/excepts and using `time.values` instead of `time.values[0]` but there are also issues with chunking in the Datasource. Should we support files like this?

rt17603:
Some netcdf input files may contain the time coordinates but as zero-dimension. This fix checks for this and adds back in this dimension as required.

Implemented for appliable data_types:
 - boundary_conditions
 - emissions
 - footprints

Fixes #349


gareth-j (approved):
Looks good!

PR #531, opened 2023-01-24
`ModelScenarion.add_flux` now returns `None` if no flux available

### LINKED ISSUE #365: Running tutorial 3 with an empty object store results in a uncaught IndexError
If I run the cell where we initally create the`ModelScenario` with an empty object store (oops) I get an IndexError. We should fail more gracefully here.

```
Unable to add obs_surface data based on keywords supplied.
 Inputs - 

 site: tac

 species: ch4

 inlet: 100m

 network: None

 start_date: 2016-07-01

 end_date: 2016-08-01

---- Search results ---
No results
Unable to add footprint data based on keywords supplied.
 Inputs - 

 site: tac

 height: 100m

 domain: EUROPE

 model: None

 start_date: 2016-07-01

 end_date: 2016-08-01

Unable to add flux data based on keywords supplied.
 Inputs - 

 species: ch4

 source: waste

 domain: EUROPE

---------------------------------------------------------------------------
IndexError                                Traceback (most recent call last)
/var/folders/j8/7lsz4kr54wj4cxkxq46f_5q80000gn/T/ipykernel_40097/1951084941.py in <module>
      1 from openghg.analyse import ModelScenario
      2 
----> 3 scenario = ModelScenario(site="tac", 
      4                          inlet="100m",
      5                          domain="EUROPE",

~/Documents/Devel/openghg/openghg/analyse/_scenario.py in __init__(self, site, species, inlet, network, domain, model, metmodel, source, sources, bc_input, start_date, end_date, obs, footprint, flux, bc)
    168 
    169         # Add flux data (directly or through keywords)
--> 170         self.add_flux(
    171             species=species,
    172             domain=domain,

~/Documents/Devel/openghg/openghg/analyse/_scenario.py in add_flux(self, species, domain, source, sources, start_date, end_date, flux)
    404             if not hasattr(self, "species"):
    405                 flux_values = list(self.fluxes.values())
--> 406                 flux_1 = flux_values[0]
    407                 self.species = flux_1.metadata["species"]
    408 

IndexError: list index out of range

```

gareth-j:
Should close #365


openghg/retrieve/_access.py
```diff
@@ -19,7 +19,9 @@
 logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
 
 DataTypes = Union[BoundaryConditionsData, FluxData, FootprintData, ObsColumnData, ObsData]
```
rt17603, line 21: Quick question: just a thought but should we move these to the `openghg.types` sub-module and import them do you think? May be good to keep these in one place.

openghg/analyse/_scenario.py
```diff
@@ -88,6 +88,9 @@ class ModelScenario:
     operations to be performed combining these inputs.
     """
 
+    def __bool__(self) -> bool:
```
rt17603, line 91: Does this only return True if all of `obs`, `footprint`, `fluxes` and `bc` are True? For this objects it's actually ok if some of these are missing depending on what operations are being performed so this could be an `or` so at least it's true if *something* is attached?

openghg/analyse/_scenario.py
```diff
@@ -446,6 +450,10 @@ def add_flux(
         if self.fluxes is not None:
             if not hasattr(self, "species"):
                 flux_values = list(self.fluxes.values())
```
rt17603, line 452: Just wondering why `self.fluxes.values()` would be an empty list? We probably want to make this so `self.fluxes` would just be None in this circumstance rather than still having keys but no values?

openghg/analyse/_scenario.py
gareth-j: OK done

openghg/retrieve/_access.py
rt17603: I just know I've reinvented the wheel in the past before I discovered `openghg.types` but if you don't think they'll have broader applicability we can leave here.

rt17603:
@gareth-j - I don't seem to be able to add you as a reviewer for this PR but I put in my suggested change. Test still passes

gareth-j:
OK great, I'll have a look.

gareth-j:
I think this is ready for merge.

PR #533, opened 2023-01-24
Removing unused test data files

### LINKED ISSUE #248: Check for unused files in tests data folder

gareth-j:
This should close #248 and removes the unused test data files that were in the repo.


rt17603 (commented):
It looks like some of the EDGAR tests fail when these are run as part of the continuous integration? May be that some of the removed EDGAR files need to be added back.

rt17603 (approved):
Looks good. Tests now pass for me locally with latest environment set up.

PR #535, opened 2023-01-25
Fix for types-urllib3 and mypy

### LINKED ISSUE #209: Adding `requests.packages.urllib3` to ignore in `mypy.ini` - check status

gareth-j:
I've removed urllib3 requests section from the mypy.ini ignore.

Closes #209

### DIFF
--- a/mypy.ini
+++ b/mypy.ini
@@ -4,9 +4,6 @@ warn_unused_configs = True
 disallow_untyped_defs = True
 exclude = openghg/data
 
-[mypy-Acquire.*]
-ignore_missing_imports = True
-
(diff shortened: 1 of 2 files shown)


rt17603 (approved):
Looks good - `mypy` seems to be happy with this update so that should suffice for checks.

PR #537, opened 2023-01-25
Removes `site_lookup.json` - move to use supplementary data

### LINKED ISSUE #534: Consolidate (and remove) local site details in openghg
### What is your issue?

Even after `site_info.json` will be split out into a separate repository (PR #519 ), there is still some site information stored and used within openghg. This should be reviewed and references to local site data should be kept to a minimum (or outsourced / removed if possible).

This includes (at least):
 - `site_lookup.json`
    - this is used by the `openghg.util.verify_site()` function

gareth-j:
- Updates `verify_site` and `site_code_finder` to us supplementary data's `site_info.json`
- Removed `site_lookup.json`
- Added helper function to create inverted dictionary from site_info

Closes #534


rt17603:
One thing with this is that PR #519 is going to influence this update. So may be worth prioritising get that one in first, merging the changes into this branch and then merging this one.

gareth-j:
OK I think this is ready to merge. The failed runner was due to a network error with `apt`. All tests passed on the other runners and locally.

rt17603:
What was the resolution around the merging of the conflicting interface for loading the json files? I believe you suggested:

      What do you think about adding the load_json_internal or load_internal_json function?

Has that been implemented? I seem to recall that was related to this PR?

openghg/standardise/meta/_attributes.py
```diff
@@ -156,15 +161,15 @@ def get_attributes(
 
     # Load attributes files
     species_attrs = get_species_info()
-    attributes_data = load_json("attributes.json", internal_data=True)
+    attributes_data = load_json(filename="attributes.json")
```
rt17603, line 164: Moving over discussion from PR #554 as these questions are more relevant here since this where the original changes seem to have been made:

    rt17603 (3 weeks ago)
    Does this still run with the internal_data flag removed? This was to flag the data was within our openghg/data
    folder and so was needed for e.g. "icos_atmos_site_metadata.json" as this isn't a full path to the data file?

    gareth-j (3 weeks ago)
    Yeah I think we both made changes to that function at a similar time. I added the path kwarg so you could load
    any file, and you added the internal_data flag. Maybe we could more easily differentiate between them by 
    creating a load_internal_json function and another load_json that just takes a full filepath? Quite a few bugs 
    were introduced when I merged in both our changes.

    rt17603 
    Yes we could approach this by using separate functions - that way it's still clear where the data is coming from
    (and what is internal to openghg).

gareth-j: No, shall I make those changes before this goes in?

rt17603: Yes I think we should make sure those changes go in here.

gareth-j: OK cool, I'll do that now.

rt17603 (approved):
Looks good. Just looks like there's some conflicts to resolve then it's done.

PR #543, opened 2023-01-26
Adding new emissions keywords

### LINKED ISSUE #446: Update specific get_* functions to allow additional inputs such as kwargs
For the `get_obs_surface()` function at the moment we can't filter by additional keywords such as "sampling_period".

We may decide to limit inputs and add e.g. "sampling_period" as options but it may be better to future proof this and allow **kwargs rather than consistently needing to keep this up to date everytime a new input keyword is allowed/defined at the metadata creation stage. This doesn't stop us from adding all the input parameters we are aware of to the `get_*` functions.

### LINKED ISSUE #489: Differentiating between versions of emissions files for same year-species-domain when adding/retrieving from the object store
### What is your issue?

When using `standardise_flux()` there is not currently an option for being able to differentiate between versions of emissions files that are of the same species, year and domain. For example, if there are emissions data of CH4 for the year 2019 gridded to the EUROPE domain from both EDGARv6.0 and EDGARv7.0 there is not an option to distinguish between these datasets when retrieving them from the object store. Having this option would be really helpful. 

I found that you can distinguish between such data if using different strings for the `source` argument in `standardise_flux`. This could be problematic when having a large number of emissions files in the object store and having to recall the different source arguments assigned. 

I found if both data files of the same domain-year-species are added to the object store using the same `source`  argument, a merged emissions dataset is retrieved from the object store with dimension that does not match the domain of interest. This exactly happened when adding the 'ch4-ukghg-edgar5-2019.nc' and 'ch4-ukghg-edgar7-2019.nc' files (for total emissions) to the object store and using `source='total'` for both emissions files. 

Having an additional argument or two in `standardise_flux` seems like it would be the most pragmatic approach to resolving this.  Maybe an argument such as `data_version` to distinguish between the version of EDGAR but also  `embedded_emissions=True/False` as a way of indicating whether UKGHG emissions have been nested into EDGAR over the UK (this might have to be generalised).

rt17603:
Adding additional emissions keywords to be used for storing and allowing lookup based on these keywords. This includes:
 - database
 - database_version
 - model

These have also been added as keywords to the `openghg.retrieve.search_flux` and `openghg.retrieve.get_flux` functions

Tutorial added incorporating these details for adding flux, footprint and bc data
  - [doc/source/tutorials/local/Adding_new_data/Adding_ancillary_data.rst](https://github.com/openghg/openghg/blob/140a0b026d5b55eb87b869adce7b589521efaa90/doc/source/tutorials/local/Adding_new_data/Adding_ancillary_data.rst)

Tests added:
 - Check new metadata keywords can be added
 - Check object store entries can be selected and distinguished based on these keywords
 - If multiple datasources are found this fails nicely - (DatabaseLookupError with error message)

Closes #489 
Closes #446


EricSaboya (approved):
The additional keyword arguments are a good inclusion for differentiating between emissions files in the object store. I'm happy with the range and choices of the additional options (model, database, database_version) given, which should ensure most users can add a range of flux data files to the object store!

gareth-j (approved):
Looks great!

Issue #494, opened 2022-11-29
Retrieving from ICOSCP - multiple datasets match criteria


joe-pitt:
### What happened?

I've been retrieving data from the ICOS CP, for example:
wao_data_level2 = retrieve_atmospheric(site="WAO", species="CH4", data_level=2)

The problem is that there are three L2 CH4 datasets for WAO on the CP (all with the same inlet height) - the ICOS dataset, the InGOS dataset and the OBSPACK-Europe dataset. The code above returns only the OBSPACK-Europe dataset, and does not appear to have an option to specify an alternative. Most sites have at least two different files with the same species, level and inlet height - it is important that the user is able to choose between them.

After some digging, it seems to me like `_retrieve_remote` loops through the `dobj_urls` (starting L270), and overwrites data if "spec" and "inlet" are the same (which they will be for these files). OBSPACK-Europe happens to be the last one listed, but I don't think this behaviour can be relied upon.

I reckon you can get at the information we want to filter on using: `dobj.info["specification"]["project"]["self"]["label"]`. But I am nowhere near confident enough to implement a fix based on this, and most likely there is a neater way (possibly inbuilt in the icoscp package, although I couldn't find it).

### Minimal Complete Verifiable Example

```Python
from openghg.retrieve.icos import retrieve_atmospheric
wao_data = retrieve_atmospheric(site="WAO", species="ch4")
wao_data.data
```

### Environment

Python 3.8.8
Openghg - devel commit: ec166689ab080b0458428aaee4c2102e37135fc7

rt17603:
@ag12733 - Something to be aware of.

gareth-j:
I'm working on this now.

Issue #548, opened 2023-01-27
Avoid deploying docs on every PR before merge


rt17603:
### What is your issue?

Currently the docs are build and deployed every time there are changes within a PR (even a draft PR etc.). This means the public facing documentation on read-the-docs is always changing including half-completed changes before the PR is merged into devel.

To keep the nice interface of being able to look at new documentation before the merge the deploy step could still run but deploy to subfolder for that PR number and then delete this when the PR is merged.

PR #552, opened 2023-02-03
Fix for Issue 494 - ICOS retrieval with multiple dataset source error

gareth-j:
This updates the ICOS `retrieve_atmospheric` function to differentiate between the multiple datasets returned by the ICOS Python library for different dataset sources such as EU ObsPack, ICOS, InGOS. A new `dataset_source` key has been added to the metadata to track the source. You can also now pass a `dataset_source` argument to the `retrieve_atmospheric` function and to `search_surface`. 

Documentation has been updated and the `ipython` Sphinx directive used to run cells of the tutorial in place during documentation build, will see how this goes, it might need some tweaking.

I've also updated the workflow yaml to only run the docs build job when something is pushed to `devel`. This should reduce the amount of runtime on the actions and save a load of unecessary work.


gareth-j:
This should close #494 and #548

rt17603:
This test tests/retrieve/icos/test_retrieve_icos.py::test_icos_retrieve_and_store fails for me locally. Looks like the same is happening the CI as well.

...
assert first_expected_metadata.items() <= first_obs_metadata.items()
E       AssertionError: assert dict_items([('species', 'co2'), ('instrument', 'ftir'), ('site', 'wao'),
...

gareth-j:
> This test `tests/retrieve/icos/test_retrieve_icos.py::test_icos_retrieve_and_store` fails for me locally. Looks like the same is happening in the CI as well.
> 
> ```
> ...
> assert first_expected_metadata.items() <= first_obs_metadata.items()
> E       AssertionError: assert dict_items([('species', 'co2'), ('instrument', 'ftir'), ('site', 'wao'),
> ...
> ```

OK that's fixed, it was a wierd issue when comparing the metadata dictionaries.

rt17603 (commented):
Looks great - I just have a few clarifications I'd like around how this functions:
 - Is this pulling all available data which matches and storing all of it?
 - Is it possible to specify which of the dataset_sources you want to retrieve?

gareth-j:
> Looks great - I just have a few clarifications I'd like around how this functions:
> 
>     * Is this pulling all available data which matches and storing all of it?
> 
>     * Is it possible to specify which of the dataset_sources you want to retrieve?

- Yeah this will pull all available data and store it.
- Yeah you can pass the `dataset_source` argument and it'll only pull the data you want.

Issue #527, opened 2023-01-23
Check high time res footprint memory usage with new `update_zero_dim` function


gareth-j:
We should check if the new `update_zero_dim` function introduced with #525 affects memory consumption of things like `ModelScenario`.

rt17603:
Covered by #513

PR #554, opened 2023-02-06
Looking to recreate and fix `overwrite=True` creates multiple objects in store.

gareth-j:
I tried recreating the bug Hannah mentioned but can't recreate it with a simple test. I had added a bit to the `Datasource.add_timed_data` function though that checks for identical data and should avoid some uneceessary processing.


openghg/retrieve/icos/_retrieve.py
```diff
@@ -282,7 +282,7 @@ def _retrieve_remote(
     # Now extract the PIDs along with some data about them
     dobj_urls = filtered_sources["dobj"].tolist()
 
-    site_metadata = load_json("icos_atmos_site_metadata.json", internal_data=True)
+    site_metadata = load_json(filename="icos_atmos_site_metadata.json")
```
rt17603, line 285: Does this still run with the `internal_data` flag removed? This was to flag the data was within our openghg/data folder and so was needed for e.g. "icos_atmos_site_metadata.json" as this isn't a full path to the data file?

rt17603 (commented):
So it looks to me like changes around the json files are being reverted to a previous version before the link to the external supplementary_info (openghg_defs) repository was added.

Whatever updates are made (e.g. if there were bugs creeping in) need to be done on top of these changes otherwise this won't integrate with the external repository.

openghg/retrieve/icos/_retrieve.py
gareth-j: Yeah I think we both made changes to that function at a similar time. I added the `path` kwarg so you could load any file, and you added the `internal_data` flag. Maybe we could more easily differentiate between them by creating a `load_internal_json` function and another `load_json` that just takes a full filepath? Quite a few bugs were introduced when I merged in both our changes.

rt17603: Yes we could approach this by using separate functions - that way it's still clear where the data is coming from (and what is internal to openghg).

rt17603:
We should get PR #527 merged first, then can finish reviewing this PR.

gareth-j:
> We should get PR #527 merged first, then can finish reviewing this PR.

Do you mean #537 ?

rt17603:
Yes, I did - PR #537

openghg/retrieve/icos/_retrieve.py
rt17603: Copied over discussion to PR #537 where these original changes were made.

gareth-j:
OK so I'm going to link #591 to this, we should decide what we want to happen before more happens with this.

gareth-j:
Also see #570

rt17603:
@gareth-j Were there any updates from this PR we wanted to preserve? Or are we happy this has now been superceded by updates around #591 ?

rt17603:
@gareth-j Were there any changes here we need to save? Or can this PR be deleted now because of more recent changes superceding this?

gareth-j:
So looking at this again it's got some code that skips the adding of data if the Datasets are equal. Would this be useful in a case that the metadata / header info in a file changed but not the data itself? The file hashes would be different but the data would be the same.

```
            # Check the equality of what we've got at the moment and what
            # we're going to add, doing a simple dict eqality comparison
            # isn't enought to check the contents of the Datasets here
            if new_data == self._data:
                for _old, _new in zip(self._data.values(), new_data.values()):
                    if not _old.equals(_new):
                        break
                else:
                    logger.warning("No new data, nothing to do.")
                    return None
```

rt17603:
> So looking at this again it's got some code that skips the adding of data if the Datasets are equal. Would this be useful in a case that the metadata / header info in a file changed but not the data itself? The file hashes would be different but the data would be the same.
> 
> ```
>             # Check the equality of what we've got at the moment and what
>             # we're going to add, doing a simple dict eqality comparison
>             # isn't enought to check the contents of the Datasets here
>             if new_data == self._data:
>                 for _old, _new in zip(self._data.values(), new_data.values()):
>                     if not _old.equals(_new):
>                         break
>                 else:
>                     logger.warning("No new data, nothing to do.")
>                     return None
> ```

Sure - would it be worth creating a separate PR for this? There's a [test this may or may not initially break in the db_integrity checks](https://github.com/openghg/openghg/blob/devel/tests/db_integrity/test_data_updates.py#L255) but that can be fixed to reflect the new expected output if it does.

gareth-j:
Closing after opening issue with code that may be useful.

PR #572, opened 2023-02-17
Draft tutorial for creating data from EDGAR database

rt17603:
Adding new tutorial for creating EDGAR emissions file from EDGAR database. This required the following additional updates (some based on external package updates):
 - Tutorial file was added to documentation making use of example files
     - Data for these tutorials were added to [openghg/example_data](https://github.com/openghg/example_data). These were EDGAR v6.0 files on native EDGAR domain (limited years).
 - ESMPy >=8.4 only works with xesmf>=0.7 so this was added as a pinned package to the environment.yml
 - Regridding across the native EDGAR domain (0.1 x 0.1 global) was too slow for tutorial using the latest version of the xemsf package so this functionality was updated to cut down the domain to the minimal needed based on the lat_out, lon_out values
 - Update to creation of grids to align with [xesmf documentation on rectilinear grids](https://xesmf.readthedocs.io/en/latest/notebooks/Rectilinear_grid.html) and to bypass new bug where expected "lat" and "lon" dimensions are no longer in the regridded output.
 - Method call to delete weight files has been removed from xesmf, so call to this function has been commented out


openghg/util/_domain.py
```diff
@@ -118,24 +118,135 @@ def _get_coord_data(coord: str, data: Dict[str, Any], domain: str) -> ndarray:
...
+                               lon_name: Optional[str] = None,
+                               reorder: bool = True) -> XrDataLikeMatch:
+    """
+    Convert longitude coordinate within an xarray data structure (DataArray or Dataset).
+
+    Args:
+        data: Data with longitude values to convert.
+        lon_name: By default will look a coord called "lon" or "longitude".
+            Otherwise must be specified.
+        reorder:
```
gareth-j, line 169: This docstring needs finishing.

gareth-j (changes requested):
Looks good! Just a fix to the docstring needed and I thnk it'll be ready to go in.

openghg/util/_domain.py
rt17603: Good spot - I'll take a look at that now.

rt17603: This has been updated

PR #576, opened 2023-02-17
Adding ability to use alternative site_filename and adding mock to use static version in tests

### LINKED ISSUE #571: Update tests to use static info files
### What is your issue?

Now that we have an external dependency on data from https://github.com/openghg/supplementary_data, the tests should be updated to bypass this with a static (paired-down if needed) version in the right format so these can run without being reliant on external data.

This may require an input for site_info_file, species_info_file etc. to be added to the top level functions for adding data to the object store and in other places.

rt17603:
Two main changes within this PR:
 - Adding `site_filename` as an input for front-end functions. This allows a "site_info.json" file to be specified directly rather than using the "site_info.json" file available from the `openghg_defs` dependency. This includes:
    - `ObsSurface.read_file` method
    - `standardise_obs` function
    - `parse_*` functions (where appropriate) this includes:
         - `parse_crds`, `parse_gcwerks`, `parse_icos`, `parse_noaa`, `parse_openghg`, `parse_tmb`
         - NOT added for others as these often did not include link with direct/indirect calls to `get_site_info()` which is the function that calls `openghg_defs`
 - Mocking the call to `openghg_defs` within the tests to use a static version of "site_info.json" within tests data folder.

The second fix is to ensure the tests aren't influenced by external changes to the openghg/supplementary_data repository. This also fixes tests which started failing in relation to recent changes made.

Closes #571

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -46,7 +46,7 @@ jobs:
     - name: Test with pytest
       run: |
         pip install --upgrade pytest
-        pytest -vs --run-cfchecks --cov=./ --cov-report=xml
+        pytest --run-cfchecks --cov=./ --cov-report=xml
     - name: Upload coverage to Codecov
       run: |
         curl -s https://codecov.io/bash > codecov;
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -46,7 +46,7 @@ jobs:
     - name: Test with pytest
       run: |
         pip install --upgrade pytest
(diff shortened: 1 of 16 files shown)


gareth-j (approved):
Looks good!

gareth-j:
The tests pass for me locally, I'll have a look to see what's happening.

rt17603:
Yeah I've tried re-running the tests in a new brand new place as well and they pass for me so I'm struggling to reproduce this error as well. Great - thanks for looking into this.

PR #579, opened 2023-02-23
Improving messaging around attribute - metadata mismatches

### LINKED ISSUE #517: Required inputs depending on data attributes
### Feature that's missing

When using openghg.standardise.standardise_surface to load observations, some attributes are required to be passed as inputs if they are not present in the file attributes. These attributes are checked (and errors raised if needed) one by one. 

Check all info missing from the file attributes that is required to be passed as inputs all at once, raising a single error stating inputs needed (rather than an error for each missing attribute)

_No response_

### Additional context

_No response_

### Additional code / helper functions

_No response_

rt17603:
When synchronising the metadata and attributes (`sync_surface_metadata` function) this currently returns an `AttrMismatchError` the first time a mismatch is found, meaning this error can keep occurring as each mismatch is fixed.

This change is to collect together all the mismatches and return an error after all have been checked with one `AttrMismatchError` instead.

Closes #517


gareth-j (approved):
Looks good to me.

PR #584, opened 2023-02-28
Fix bug in sampling period checking

### LINKED ISSUE #583: Sampling period of "3600" evaluated as "3e-6"
### What happened?

If the `sampling_period` value for `ObsSurface.read_file()` is used and set to a string value of `"3600"` or a float/int value of `3600`, this evaluates to `"3e-6"`.

The input for `sampling_period` is specified as "Sampling period in pandas style (e.g. 2H for 2 hour period, 2m for 2 minute period)." within the documentation. So this implies that passing a value of `"3600"` or `3600` rather than `"3600s"` should raise a ValueError at the appropriate check (~L186 in openghg/store/_obssurface.py). Rather than raising a ValueError this evaluates this as `"3e-6"` instead due to the attempt to convert into seconds. (This may then cause a ValueError to be raised later on.)

In fact, the current check would evaluate any value of >=1000 to be >0 seconds and so would not be caught by this error check.

```Python
from openghg.store import ObsSurface

ObsSurface.read_file(..., sampling_period="3600")

# Does not raise a ValueError with message "Invalid sampling period result, please pass a valid pandas time such as 1m for 1 minute."
```

### Relevant log output

_No response_

_No response_

### Environment

<details>

</details>

rt17603:
Updating a few bugs in sampling period checking:
 - updating checks within `ObsSurface.read_file()` on whether this has been specified incorrectly (based on current documentation i.e. doesn't contain a unit). Updated to explicitly check whether this can be evaluated to just a number and raising a ValueError.
 - updating logic within `parse_gcwerks()` - converting sampling period to a Timedelta was raising a ValueError as this was a string when a float was expected by the function.

Closes #583

### DIFF
--- a/openghg/standardise/surface/_gcwerks.py
+++ b/openghg/standardise/surface/_gcwerks.py
@@ -255,8 +255,9 @@ def _read_data(
 
     if sampling_period is not None:
         # Compare input to definition within json file
-        file_sampling_period = pd_Timedelta(seconds=extracted_sampling_period)
-        comparison_seconds = abs(sampling_period - file_sampling_period).total_seconds()
+        file_sampling_period_td = pd_Timedelta(seconds=float(extracted_sampling_period))
+        sampling_period_td = pd_Timedelta(seconds=float(sampling_period))
+        comparison_seconds = abs(sampling_period_td - file_sampling_period_td).total_seconds()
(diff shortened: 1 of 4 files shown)


gareth-j (approved):
Looks good!

Issue #566, opened 2023-02-10
Add `check_config` function


gareth-j:
We should add a `check_config` function so users can be sure their configuration file is valid after they make changes to it.

gareth-j:
Done with #585

Issue #565, opened 2023-02-10
Remove print of config dictionary from `--quickstart`


gareth-j:
Printing the config dictionary could confuse users that then see it as `toml`. Either print the `toml` correctly or just remove the print.

gareth-j:
Done with #585

PR #585, opened 2023-03-02
Fix for config file not being written

gareth-j:
This fixes an issue with the configuration file not being written out correctly. It also adds a quick `check_config` function that makes sure the file is in the correct format and contains the correct items.


gareth-j:
Closes #566 and #565

openghg/util/_user.py
```diff
@@ -153,3 +146,30 @@ def read_local_config() -> Dict:
...
+
+    config = read_local_config()
+    uid = config["user_id"]
+    object_stores = config["object_store"]
+
+    try:
+        uuid.UUID(uid, version=4)
+    except ValueError:
+        logger.exception("Invalid user ID. Please re-run quickstart to setup a valid config file.")
+        raise
```
rt17603, line 171: Was this meant to raise a specific error here?

openghg/util/_user.py
```diff
@@ -153,3 +146,30 @@ def read_local_config() -> Dict:
...
+    """Check that the user config file is valid and the paths
+    given in it exist.
+
+    Returns:
+        bool
+    """
+    config_path = get_user_config_path()
+
+    if not config_path.exists():
+        logger.warning("Configuration file does not exist. Please create it by running openghg --quickstart.")
```
rt17603, line 161: Does this mean we would now expect the user to always run `$ openghg --quickstart` on the command line?

.github/workflows/workflow.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
     - name: Test with pytest
       run: |
         pip install --upgrade pytest
-        pytest --run-cfchecks --cov=./ --cov-report=xml
+        pytest -s --run-cfchecks --cov=./ --cov-report=xml
```
rt17603, line 49: Do we still need `-s` in the pytest checks or was this just needed for checking the issue this PR is fixing?

rt17603 (commented):
Would be good to add a description to this PR to make it clear what is being added/changed. A few comments/questions added here.

openghg/util/_user.py
gareth-j: Ah good point, I'll update that so it also asks them to call `create_config`. I'll check the setup documentation as well.

.github/workflows/workflow.yaml
gareth-j: Yeah there was an issue with the capture of output so I added that, I've removed it now and everything works.

rt17603: Great, makes sense.

rt17603:
By the way, I *think* if you updated your PR description to include the "Closes #566 and #565" this will automatically close these issues when this is merged. Otherwise, I don't think it will.

openghg/util/_user.py
gareth-j: Because it's in the except of a ValueError, it's so you can do something before re-raising the exception.

rt17603: Ah that's useful to know and I've learnt something. Thanks for explaining.

rt17603 (approved):
Looks good. The only question is how we want to handle running the tutorials and whether we want to allow this to bypass the create config step but this can be addressed on a separate PR.

PR #605, opened 2023-03-21
Read in `long_name` from openghg_defs site_info - Issue598

### LINKED ISSUE #598: ICOS retrieve error - found via doc build
So I just encountered this error whilst building the docs, in `Retrieving_remote_data.rst` we use the `ipython::` Sphinx directive which runs the lines as it goes. I get this error which looks like somethings changed since we outsourced the metadata maybe?

```python
File ~/Documents/Devel/openghg/openghg/standardise/meta/_metadata.py:84, in sync_surface_metadata(metadata, attributes, keys_to_add, update_mismatch)
     82 if str(value).lower() != str(attr_value).lower():
     83     if not update_mismatch:
---> 84         raise AttrMismatchError(
     85             f"Metadata mismatch for '{key}', metadata: {value} - attributes: {attr_value}"
     86         )
     87     else:
     88         logger.warning(
     89             f"Metadata mismatch for '{key}', metadata: {value} - attributes: {attr_value}\n"
     90             f"Updating metadata to use attribute value of {key} = {attr_value}"
     91         )

AttrMismatchError: Metadata mismatch for 'station_long_name', metadata: WAO - attributes: Weybourne Observatory, UK
```

gareth-j:
This closes #598 by adding a read in of `station_long_name` from `openghg_defs`.

### DIFF
--- a/openghg/retrieve/icos/_retrieve.py
+++ b/openghg/retrieve/icos/_retrieve.py
@@ -1,6 +1,7 @@
 from typing import Any, Dict, List, Optional, Union
 from openghg.dataobjects import ObsData
-from openghg.util import running_on_hub
+from openghg.util import running_on_hub, load_json
+import openghg_defs
 import logging
 
 logger = logging.getLogger("openghg.retrieve")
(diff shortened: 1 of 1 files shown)


rt17603 (commented):
It looks like this change has already been merged as I can see this in devel?
 - https://github.com/openghg/openghg/blob/devel/openghg/retrieve/icos/_retrieve.py#L388

Did this go in on another PR?

gareth-j:
Yeah I think this can be closed.

PR #607, opened 2023-03-28
Fixed logic for specifying default objectstore path in config file

mrghg:
Previously, if a user didn't enter path when promoted during create_config, the default path was being overwritten.

For me, the tests for these functions don't work. It could be to do with the mocking. I get errors when the tests try to clean up the testing store, like:

No such file or directory: '/var/folders/ck/vrz4wwp115qb8jbc7y_wk1r00000gq/T/openghg_testing_store


rt17603 (commented):
Looks good to me - just would be good to get the CI passing for this.

rt17603 (approved):

PR #611, opened 2023-03-30
Make ICOS retrieval function filter by sampling height correctly

### LINKED ISSUE #599: ICOS retrieval function retrieving more items than expected
When running through the ICOS retrieval function with an empty object store we get 6 items returned. If you run through again with the same function calls we only get 2 objects returned.

So to reproduce:

1. Empty object store
2. `use_tutorial_store()`
3. `sac_data = retrieve_atmospheric(site="SAC", species="ch4", sampling_height="100m")`
4. `len(sac_data) == 6` first time
5. `sac_data = retrieve_atmospheric(site="SAC", species="ch4", sampling_height="100m")`
6. `len(sac_data) == 2` second time

joe-pitt:
This closes #599 . There was a missing sampling_height argument in the call to _retrieve_remote from local_retrieve within openghg.retrieve.icos. This meant that it wasn't possible to filter by sampling_height.

### DIFF
--- a/openghg/retrieve/icos/_retrieve.py
+++ b/openghg/retrieve/icos/_retrieve.py
@@ -183,6 +183,7 @@ def local_retrieve(
         standardised_data = _retrieve_remote(
             site=site,
             species=species,
+            sampling_height=sampling_height,
             data_level=data_level,
             dataset_source=dataset_source,
             update_metadata_mismatch=update_metadata_mismatch,


joe-pitt:
The test is failing at the stage where it filters by sampling_height (L282 of _retrieve.py). I try running the identical test on BP1 and the retrieval works fine. Possibly it is because I am using v1.5.3 of pandas? Or it could be something to do with #612 ? This issue was causing the retrieval to fail for me on BP1 before I rolled back the version of icoscp, although I was getting a different error to this.

gareth-j:
OK the issue here is due to the mocks of the calls to `icoscp` functions. I'll update the mocks to make sure they reflect what `icoscp` now returns as this may have changed.

gareth-j (approved):

gareth-j:
This is good to go in, I've also added some more fixes and checks in #630.

Issue #608, opened 2023-03-29
Sorting after recombination of data sources


mrghg:
### What is your issue?

It looks like the default sorting during datasource recombination has been disabled to reduce memory usage. However, I think users will find this confusing, and it's going to mean that we'll need to add a sorting step to most outputs that we generate. E.g., at the moment, a simple line plot of unsorted data looks a bit of a mess. Also, if users output the data to an array, I reckon they will be confused as to why the trends look weird.

I propose setting the default to sort=True, and then disabling if we know an operation is going to use up a lot of memory.

PR #614, opened 2023-03-31
Improved timeseries plot

mrghg:
Improvements to time series plot:
- Uses species name in y-axis title
- Tries to position legend sensibly
- Formats units properly
- Adds OpenGHG logo
- Some bug fixes

This PR also addresses Issue #608, which was leading to some plotting artefacts, unless sorted before plotting.


openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
+
+def _latex2html(
+    latex_string: str
+) -> str:
+    """Replace latex sub/superscript formatting with html.
+    Written because the latex formatting in Plotly seems inconsistent
+    (works in Notebooks, but not VC Code at the moment).
+
+    Args:
+        latex_string:
```
rt17603, line 18: Documentation string is not quite complete here.

openghg/plotting/_timeseries.py
```diff
@@ -95,28 +227,64 @@ def plot_timeseries(
...
 
         y_data *= unit_conversion
 
-        fig.add_trace(go.Scatter(name=legend, x=x_data, y=y_data, mode="lines"))
+        unit_string = attributes_data["unit_print"][unit_value]
 
-    return fig
+        # Determine whether data is ascending or descending (positioning of legend)
+        ny = len(y_data)
+        data_left = y_data.isel(time=slice(int(ny * 0.9), -1)).mean()
```
rt17603, line 239: Just to check, are we able to assume the data will always be sorted? (I know `recombine_datasets` does this by default now but determining ascending/descending in this way does make that assumption).

As an alternative, ascending/descending could also be determined using [the xarray `DataArray.diff` method](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.diff.html) which looks at the difference between each value in one go - you could also grab the mean of this or check every value is > or < 0 for instance.

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
+        xref="x domain",
+        yref="y domain",
+        sizex=0.1,
+        sizey=0.1)
+    logo_dict.update(logo_pos)
+
+    return logo_dict
 
 
 def plot_timeseries(
```
rt17603, line 130: @gareth-j - should/could these changes be incorporated into the `ObsData.plot_timeseries()` method? How would it be best to do that do you think?

rt17603 (commented):
Looks good to me - a few small suggested changes included as comments included below (incomplete docstring, adding back in a default for xlabel).

A larger thing is making sure useful changes are also incorporated into the `ObsData.plot_timeseries` method which is another way to plot the observation data so I've asked @gareth-j for thoughts on how this is best approached.

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
     yvar: Optional[str] = None,
     xlabel: Optional[str] = None,
     ylabel: Optional[str] = None,
     units: Optional[str] = None,
+    logo: Optional[bool] = True,
 ) -> go.Figure:
     """Plot a timeseries
 
     Args:
         data: ObsData object or list of objects
```
rt17603, line 142: Was there a reason to remove the title from the figure? This does mean the figure can be made more explicit about what is being displayed?

openghg/plotting/_timeseries.py
```diff
@@ -95,28 +227,64 @@ def plot_timeseries(
...
+        raise NotImplementedError("Can't plot two different units yet")
+
+    # Write species and units on y-axis
+    if ylabel:
+        fig.update_yaxes(title=ylabel)
+    else:
+        ytitle = ", ".join(set(species_strings)) + " (" + unit_strings[0] + ")"
+        fig.update_yaxes(title=ytitle)
+
+    if xlabel:
```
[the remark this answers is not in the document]
mrghg: Yeah, I guess we should put that back. I actually removed it on purpose, to remove clutter, but it's the kind of thing I tell the undergrads off for!

openghg/plotting/_timeseries.py
mrghg: No. I think I forgot to add it back in. Happy to put an optional title back.

mrghg:
Thanks for looking at this @rt17603. I can try to fix these issues when I next have a minute, or I'm happy for you to tidy this branch up, if you'd rather just get it done.

openghg/plotting/_timeseries.py
rt17603: PR #624 has now been added (set to merge into this branch) for this.

rt17603:
@mrghg - it's up to you really. At the moment I've put this on the 0.5.1 milestone as a tidy up from the Code Retreat but can reprioritise this is you wanted the time to make these changes yourself. Just let us know either way.

openghg/plotting/_timeseries.py
gareth-j: Yeah this needs completing.

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
+    """Create Plotly dictionary for logo
+
+    Args:
+        logo_pos: Dictionary containing the position of the logo
+
+    Returns:
+        Dictionary containing logo + position parameters
+    """
+
+    logo = base64.b64encode(open(get_datapath("OpenGHG_Logo_NoText_transparent_200x200.png"), 'rb').read())
```
gareth-j, line 118: We've got a dangling file open here, the `open(get_datapath("OpenGHG_Logo_NoText_transparent_200x200.png")` coud be replaced by something like

```python
logo_path = get_datapath("OpenGHG_Logo_NoText_transparent_200x200.png")
logo_b64encoded = base64.b64encode(logo_path.read_bytes())
```

openghg/plotting/_timeseries.py
gareth-j: Yeah I think having an option to add a title is good.

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
         units: Units for y axis
+        logo: Show the OpenGHG logo
     Returns:
         go.Figure: Plotly Graph Object Figure
     """
-    from openghg.util import load_json
+    from openghg.util import load_internal_json
+
+    if not data:
+        print("No data to plot, returning")
```
gareth-j, line 155: Maybe this should use logger.info or logger.warn instead?

gareth-j: They both print to screen

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
+
+
+def _plot_legend_position(
+        ascending: bool
+) -> Tuple[Dict, Dict]:
+    """Position of legend and logo,
+    depending on whether data is ascending or descending
+
+    Args:
+        ascending (bool): Is the data ascending
```
gareth-j, line 70: Here we don't need the type after the variable name as we're using type hints.

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,191 @@
...
+def _plot_logo(
+        logo_pos: Dict,    
+) -> Dict:
+    """Create Plotly dictionary for logo
+
+    Args:
+        logo_pos: Dictionary containing the position of the logo
+
+    Returns:
+        Dictionary containing logo + position parameters
```
gareth-j, line 115: This should be `dict: Dictionary containing logo + position parameters`

openghg/plotting/_timeseries.py
```diff
@@ -95,28 +227,64 @@ def plot_timeseries(
...
+        # Save units and species names for axis labels
+        unit_strings.append(unit_string_html)
+        species_strings.append(species_string)
 
-#     for dat in data:
+    if len(set(unit_strings)) > 1:
+        raise NotImplementedError("Can't plot two different units yet")
+
+    # Write species and units on y-axis
+    if ylabel:
```
gareth-j, line 271: Here we should check `if ylabel is not None`

openghg/plotting/_timeseries.py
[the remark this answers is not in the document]
gareth-j: This should be `if `xlabel is not None`

gareth-j (changes requested):
This needs some changes before it's ready to merge in.

gareth-j:
This also needs a `black --line-length 110` run over it.

openghg/plotting/_timeseries.py
rt17603:
```suggestion
        ascending: Is the data ascending
```

openghg/plotting/_timeseries.py
rt17603:
```suggestion
        dict: Dictionary containing logo + position parameters
```

openghg/plotting/_timeseries.py
rt17603: Added details to doc-string

openghg/plotting/_timeseries.py
rt17603: Added in this suggested change and a check for whether the overalll diff is positive or negative.

openghg/plotting/_timeseries.py
rt17603: Changed to your suggestion above (small tweak to line below as well to use `logo_b64encoded`)

openghg/plotting/_timeseries.py
rt17603: Updated doc-string

openghg/plotting/_timeseries.py
rt17603: Updated this

rt17603:
> This also needs a `black --line-length 110` run over it.

Have now run $ black --line-length 110 over this.

openghg/plotting/_timeseries.py
```diff
@@ -1,61 +1,172 @@
...
     ylabel: Optional[str] = None,
     units: Optional[str] = None,
+    logo: Optional[bool] = True,
 ) -> go.Figure:
     """Plot a timeseries
 
     Args:
         data: ObsData object or list of objects
-        title: Title for figure
         xvar: x axis variable, defaults to time
```
rt17603, line 126:
```suggestion
        xvar: x axis variable, defaults to thyme
```

gareth-j (approved):

openghg/plotting/_timeseries.py
```diff
@@ -1,12 +1,16 @@
+from __future__ import annotations
 import logging
 from typing import List, Optional, Union, Dict, Tuple
 import plotly.graph_objects as go
 import numpy as np
 import base64
+from typing import TYPE_CHECKING
```
rt17603, line 7: Out of interest - what is this doing?

rt17603 (approved):
Looks good to me - I did have a question on what the TYPE_CHECKING detail added is doing so I understand this change made but otherwise happy for this to be merged.

openghg/plotting/_timeseries.py
gareth-j: `TYPE_CHECKING` is always False at runtime so you can do imports during type checking to allow mypy to understand the types correctly.

PR #615, opened 2023-03-31
Reviving ECMWF met download

mrghg:
Elena has made this work again with the ECMWF API. 

@gareth-j , can you take a look at how this is being put into the object store? 

Need to deal with re-downloading data from an overlapping time period (currently, doesn't update), and how to deal with "temporary" downloaded files.

Tests also need to be revived.

### DIFF
--- a/openghg/retrieve/met/_ecmwf.py
+++ b/openghg/retrieve/met/_ecmwf.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 
-import cdsapi  # type: ignore
+
 import numpy as np
 from typing import TYPE_CHECKING, List, Optional, Union
 import os
@@ -48,11 +48,14 @@ def retrieve_met(
         raise AttributeError("You must pass either the argument years or both start_date and end_date")
     
     # check that user has Copernicus key setup
-    is_file = os.path.isfile((os.getenv("HOME") if key_path is None else key_path) +"/.cdsapirc")
-    if is_file==False:
+    key_path =(os.getenv("HOME") if key_path is None else key_path) + "/.cdsapirc"
+    if not os.path.isfile(key_path):
         raise FileNotFoundError(
             f'No API key file was found at {(os.getenv("HOME") if key_path is None else key_path) +"/.cdsapirc"}. \n' 
-            'Follow the instructions at xxx to create it or pass the right file path to key_path')
+            'Follow the instructions at https://cds.climate.copernicus.eu/api-how-to to create it or pass the right file path (where your file .cdapirc is) to key_path')
+    
+    import cdsapi 
+
     
     if variables is None:
         variables = ["u_component_of_wind", "v_component_of_wind"]
@@ -82,11 +85,14 @@ def retrieve_met(
 
     cds_client = cdsapi.Client()
     dataset_name = "reanalysis-era5-pressure-levels"
-    default_save_path="/user/work/ef17148/openghg/metdata/" ## fix this to have a dynamic one!
+    default_save_path=os.path.join(os.getcwd().split("openghg")[0], "openghg/metdata") ## fix this to have a dynamic one!
     save_path=default_save_path if save_path is None else save_path
+    assert os.path.isdir(save_path), f"The save path {save_path} is not a directory. Please create it or pass a different save_path"
+
+    print(f"Retrieving to {save_path}")
+
 
     years, months = list(set([str(download_date).split("-")[0] for download_date in download_list])), list(set([str(download_date).split("-")[1] for download_date in download_list]))
-    print
     request = {
         "product_type": "reanalysis",
         "format": "netcdf",
--- a/openghg/retrieve/met/_ecmwf.py
+++ b/openghg/retrieve/met/_ecmwf.py
@@ -1,8 +1,15 @@
 from __future__ import annotations
 
-# import cdsapi  # type: ignore
-# import numpy as np
+import cdsapi  # type: ignore
+import numpy as np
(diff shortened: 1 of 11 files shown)


mrghg:
Note that there is a tutorial in the openghg-tutorials repo

openghg/store/_metstore.py
```diff
@@ -94,7 +115,7 @@ def search(
         from openghg.store.base import Datasource
 
         datasources = (Datasource.load(uuid=uuid, shallow=True) for uuid in self._datasource_uuids)
-
+        #print(self._datasource_uuids)
         # We should only get one datasource here currently
```
gareth-j, line 119: The search function should be removed and any searching passed to a `search_met` function.

openghg/store/_metstore.py
```diff
@@ -119,14 +140,30 @@ def _store(self, met_data: METData) -> None:
...
 
         name = "_".join((metadata["site"], metadata["network"], date_str))
         self._datasource_uuids[datasource.uuid()] = name
+        
+        print(self._datasource_uuids[datasource.uuid()], datasource.uuid())
+
+        met.add_single_datasource(uuid={name:datasource.uuid()}, data=met_data, metastore=metastore)
+
+        met.save()
+        #metastore.close()
```
gareth-j, line 165: When using the metadata store it should either be used with a context manager or explicitly closed, records won't get written correctly otherwise.

openghg/store/base/_base.py
```diff
@@ -191,6 +191,27 @@ def add_datasources(self, uuids: Dict, data: Dict, metastore: TinyDB) -> None:
                 metastore.insert(meta_copy)
                 self._datasource_uuids[uid] = key
 
+    def add_single_datasource(self, uuid: Dict, data: Dict, metastore: TinyDB) -> None:
```
gareth-j, line 194: This is handy but we could use `add_datasources` here. That's another function that's due a rework to make this kind of this easier.

gareth-j (changes requested):
This looks great Elena, thanks! I've added some comments requesting some changes, I'll also add some more general notes to the PR.

gareth-j:
Also @elenafillo if you haven't already, it could be a good idea to setup pre-commit so the code is formatted, linted etc and helps to avoid pushing code that will fail CI. You can get this installed by doing a `pre-commit install` in the root of the repo.

openghg/retrieve/met/_ecmwf.py
```diff
@@ -34,25 +42,100 @@ def retrieve_met(
...
+    # check that user has Copernicus key setup
+    key_path =(os.getenv("HOME") if key_path is None else key_path) + "/.cdsapirc"
+    if not os.path.isfile(key_path):
+        raise FileNotFoundError(
+            f'No API key file was found at {(os.getenv("HOME") if key_path is None else key_path) +"/.cdsapirc"}. \n' 
+            'Follow the instructions at https://cds.climate.copernicus.eu/api-how-to to create it or pass the right file path (where your file .cdapirc is) to key_path')
+    
+    import cdsapi 
+
+
```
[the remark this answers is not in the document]
elenafillo: currently, retrieve_met always retrieves the file from the API; METStore.retrieve does what you have described in the comments - searches the current store and if no file is found calls retrieve_met and stores the new data. Should we merge these two? how?

openghg/retrieve/met/_ecmwf.py
```diff
@@ -142,122 +225,123 @@ def retrieve_met(
...
-# def _get_site_data(site: str, network: str) -> Tuple[float, float, float, List]:
-#     """Extract site location data from site attributes file.
 
-#     Args:
-#         site: Site code
-#     Returns:
-#         dict: Dictionary of site data
-#     """
-#     from openghg.util import load_json
+def _get_site_data(site: str, network: str) -> Tuple[float, float, float, List]:
```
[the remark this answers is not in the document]
elenafillo: renamed for now

openghg/store/base/_base.py
elenafillo: I couldn't get `add_datasources` to work, I think it would only when using the correct datasource_lookup -> assign_data -> add_datasources, as currently met data does not have the right format. Will fix to use the new store method rather than the existing METStore._store and then it should work

openghg/store/_metstore.py
```diff
@@ -37,7 +40,7 @@ def save(self) -> None:
         set_object_from_json(bucket=bucket, key=obs_key, data=self.to_data())
 
     @staticmethod
-    def retrieve(site: str, network: str, years: Union[str, List[str]]) -> METData:
+    def retrieve(site: str, network: str, years: Optional[Union[str, List[str]]] = None, start_date: Optional[str] = None, end_date: Optional[str] = None, variables: Optional[List[str]] = None, key_path: Optional[str] = None, save_path: Optional[str] = None,) -> METData:
```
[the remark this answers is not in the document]
elenafillo: retrieve checks the local store and if the object is not there, calls retrieve_met to retrieve from copernicus

gareth-j:
Can I close this? I think @elenafillo 's PR #833 supercedes it.

rt17603:
> Can I close this? I think @elenafillo 's PR #833 supercedes it.

@elenafillo and @mrghg - following on from @gareth-j 's inquiry - does PR #833 supercede this now so this draft PR can be closed?

rt17603:
This has been superceded by PR https://github.com/openghg/openghg/pull/833

Issue #618, opened 2023-04-06
Pandas >=2.0 currently failing tests


rt17603:
### What happened?

At the moment, if we upgrade to pandas 2.0+ this seems to break some underlying assumptions in the tests causing some additional failures.

### Relevant log output

```Python
=========================== short test summary info ============================
FAILED tests/analyse/test_scenario.py::test_combine_obs_footprint - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_calc_modelled_obs - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_calc_modelled_obs_period - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_calc_modelled_baseline - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_footprints_data_merge - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_model_resample_ch4 - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_model_modelled_obs_ch4 - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_modelled_baseline_ch4 - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_model_modelled_obs_co2 - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_modelled_baseline_radon - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/analyse/test_scenario.py::test_modelled_baseline_short_life - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/retrieve/test_access.py::test_get_obs_surface_average_works_without_longname - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/retrieve/test_access.py::test_get_obs_surface[inlet-248m] - TypeError: __init__() got an unexpected keyword argument 'base'
FAILED tests/retrieve/test_access.py::test_get_obs_surface[height-248m] - TypeError: __init__() got an unexpected keyword argument 'base'
[...]

PR #619, opened 2023-04-06
Temporarily pinning pandas to < 2.0

rt17603:
The upgrade to pandas 2.0 seems to have caused a set of tests to fail. Suggest pinning this to 2.0 to allow development to continue until we can address these issues.

Relates to #618


gareth-j (approved):
Looks good!

PR #624, opened 2023-04-24
Updating ObsData.plot_timeseries() method to use openghg.plotting.plot_timeseries() function

rt17603:
Updating the relevant method on ObsData to use the plot_timeseries() function directly. This is to ensure optimisation and changes don't have to be made twice.

See comments on PR #614

### DIFF
--- a/openghg/dataobjects/_obsdata.py
+++ b/openghg/dataobjects/_obsdata.py
@@ -84,46 +84,18 @@ class ObsData(_BaseData, abc.Mapping):
         xlabel: Optional[str] = None,
         ylabel: Optional[str] = None,
         units: Optional[str] = None,
+        logo: Optional[bool] = True,
     ) -> go.Figure:
         """Plot a timeseries"""
+        import openghg.plotting
 
-        species = self.metadata["species"]
-        site = self.metadata["site"]
-        inlet = self.metadata["inlet"]
-
-        if title is None:
-            title = f"{species.upper()} at {site.upper()} - {inlet}"
-
-        if xlabel is None:
-            xlabel = "Date"
-
-        if ylabel is None:
-            ylabel = "Concentration"
-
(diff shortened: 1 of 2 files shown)


openghg/dataobjects/_obsdata.py
```diff
@@ -84,46 +84,18 @@ def plot_timeseries(
...
-        font = {"size": 14}
-
-        title_layout = {"text": title, "y": 0.9, "x": 0.5, "xanchor": "center", "yanchor": "top"}
-
-        layout = go.Layout(
-            title=title_layout,
-            xaxis=dict(title=xlabel),
-            yaxis=dict(title=ylabel),
-            font=font,
+        import openghg.plotting
```
gareth-j, line 90: This feels pedantic but we usually do `from openghg.plotting import plot_timeseries`, be good to keep our import style consistent.

gareth-j (changes requested):
Update to keep import style consistent then it'll be ready to merge.

openghg/dataobjects/_obsdata.py
rt17603: Yeah that's fine - my only concern was namespacing as then we'd have two `plot_timeseries` functions and wasn't sure if that would potentially cause a problem? But happy to change it to be consistent if that isn't the case.

rt17603: Ok so I've updated it to:
```from openghg.plotting import plot_timeseries as general_plot_timeseries```
to make sure it's clear where this is coming from.

Is that consistent enough with the previous style?

gareth-j (approved):

openghg/dataobjects/_obsdata.py
gareth-j: Yeah looks good

PR #625, opened 2023-04-24
Metadata mismatch tidy

### LINKED ISSUE #613: mismatch between metadata and obs file attributes preventing data from being added to object store
### What is your issue?

When there is a mismatch between metadata in the openghg supplimentary data files and between attributes in a data file, the data file cannot be added to an object store.

It had been decided that when there is a mismatch the code should fall back on using the information in the data file's attributes, and ignore and conflicting information in supplimentary_data.

The code should then print out a warning about which source of information has been used.

rt17603:
This is a first pass at dealing with mismatches between sources of data: 'attributes' (information largely derived from the data being added) and 'metadata' (information largely derived from separate details e.g. site_info) which currently block data from being added to the object store.

This fix exposes the `update_mismatch` variable to the front end user interface when adding data which allows mismatches to be updated using the attributes and a warning logged. Tests added for this.

More knock-on changes made as a part of this:
 - `global_attributes` and `site_attributes` have been rearranged in `get_attributes()` function. Previously any pre-existing global attributes were being over written by the site attributes (from site_info.json). This may have been wanted behaviour but it meant some values were being silently ignored which may cause more confusion in the long term.
 - The way the `parser_fn` is called within `ObsSurface.read_file()` has been updated to check the underlying function for which inputs it accepts. This also defines explicit required inputs and optional inputs (`update_mismatch` is included as an optional input so not all parser functions accept and use it as appropriate).
 - Defined a list of metadata keywords expected to be floats to ensure we keep these as floats even if the attributes have stored them as strings instead. *This was included a consequence of the changes to global attributes above - they were previously being overwritten by site_info data but now they're not.*
 - `updata_metadata_mismatch` already included in ICOS retrieve functions has been renamed to `update_mismatch` to conform to other inputs (and in preparation for further updates).

Note: the longer term fix should allow attributes OR metadata to be used but this solves the short term reported problem of not being able to add the data at all - #638 added to track this.

Closes #613


openghg/standardise/meta/_metadata.py
```diff
@@ -33,6 +33,22 @@ def metadata_default_keys() -> List:
     return default_keys
 
 
+def metadata_keys_as_floats() -> List:
+    """
```
gareth-j, line 37:
```
   """Define which keys should be consistently stored as numbers in the metadata
    (even if they are not numbers within the attributes).

    Returns:
        list: List of keys to convert to floats

    """
```

gareth-j: Hmm, not sure how this commit change suggestion thing works.

gareth-j (changes requested):
Looks good, that docstring just needs updating.

openghg/standardise/meta/_metadata.py
rt17603: I believe the change suggestion only allows one line to be changed.

I'm a little confused about the suggested change here? This should have passed all the linting already?

openghg/standardise/meta/_metadata.py
```diff
@@ -33,6 +33,21 @@ def metadata_default_keys() -> List:
     return default_keys
 
 
+def metadata_keys_as_floats() -> List:
+    """Define which keys should be consistently stored as numbers in the metadata
+    (even if they are not numbers within the attributes).
```
gareth-j, line 38: I think this still needs a Returns: section in the docstring.

gareth-j (changes requested):

gareth-j (approved):
Looks good!

Issue #617, opened 2023-04-05
Deletion of the object store fails if it isn't created.


gareth-j:
So this is very rare but I've just encountered this where the cleanup fails if the store wasn't created. We just need a try except around the rmtree call.
```
    File "~/Documents/Devel/openghg/tests/conftest.py", line 61, in pytest_sessionfinish
    shutil.rmtree(tmp_store_path)
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/shutil.py", line 724, in rmtree
    onerror(os.lstat, path, sys.exc_info())
  File "/opt/homebrew/Cellar/python@3.9/3.9.16/Frameworks/Python.framework/Versions/3.9/lib/python3.9/shutil.py", line 722, in rmtree
    orig_st = os.lstat(path)
FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/j8/7lsz4kr54wj4cxkxq46f_5q80000gn/T/openghg_testing_store'
```

gareth-j:
Some of the tests don't write to the object store, so if these are run then the `pytest_sessionfinish` fixture still tries deleting an object store that doesn't exist. I just encountered this looking at an ICOS bug.

PR #626, opened 2023-04-27
Catch edge case where nothing is written to the store

gareth-j:
This adds a catch for the FileNotFoundError which is raised when individual tests are run but nothing is written to the object store.


gareth-j:
Closes #617

rt17603:
I think if you update your Pull Request description to include "Closes #617" this will automatically link this to the issue and close it when the PR is merged.

rt17603 (approved):
Great - looks like a good fix. This change should go in right away.

PR #632, opened 2023-05-02
Fix for incorrect and missing functions from utils

gareth-j:
This is just a quick one to correct a function name and add a couple of missing functions in the configuration section of the documentation.


rt17603 (commented):
Looks good to me but would be good to get the CI passing again before this is merged. After this is done, happy for this to be merged.

gareth-j:
PR https://github.com/openghg/openghg/pull/630 needs to go in before this.

PR #643, opened 2023-05-10
Adding standardise_column function

### LINKED ISSUE #569: Add a `standardise_column` (or similarly named) function
At the moment the only way to add column data is by using `ObsColumn.read_file`. We should add a standardisation function to make this easier and match the interface for the other data types.

rt17603:
Adding in a `standardise_column` function as a front-end to using the `ObsColumn.read_file()` function. This is to align with other `standardise_*` functions which offer the same functionality for other data types.

Closes #569


openghg/standardise/column/_openghg.py
```diff
@@ -260,6 +260,7 @@ def parse_openghg(
     # Update attributes to match metadata after cleaning
     attributes.update(metadata)
 
+    # TODO: Decide if the key here should be more descriptive that just `species`
```
gareth-j, line 263: Yeah I think we can do away with this nested dictionary returning when we refactor the assignment of data to the Datasources. I think we'll be able to simplify that section of code quite a bit.

gareth-j (approved):
Looks good!

Issue #496, opened 2022-12-08
Allow users to access data from a group object store and a local one


rt17603:
We need to allow users to be able to access data from multiple object stores at once - their own local one and a group object store. This also means that it needs to be easy to change and work out which object store is being written to when adding data.

This may supercede Issue #262.

rt17603:
Need to set up a meeting to discuss this and make sure we understand the underlying requirements.

rt17603:
Links to PR #555 
Key summary from meeting:
 - Multiple object stores should be flat (no "precedence" between them)
 - Add ability to specify which object store to add to, search and to retrieve from
 - Allow searching all object stores (can return multiple SearchResults objects)
 - Allow retrieving from all object stores, *as long as there is no ambiguity* otherwise fail
 - Allow read/write equally from/to all object stores for now (consider read-only object stores in future)

gareth-j:
I'm working on this now.

PR #647, opened 2023-05-10
Adding multiple object store functionality

gareth-j:
This will eventually add the ability to read and write from multiple object stores. See discussion at #496.


gareth-j:
Closing due to working being done on #664 .

PR #648, opened 2023-05-12
Adding ability to read and interpret height_name for footprints

### LINKED ISSUE #560: Update matching between site inlet and footprint inlet/height to add flexibility
### What is your issue?

For the WAO site there is going to be a mismatch from between the inlet height (new value of "10m") and the height as which the footprints were run (~~"21magl"~~ "20magl"). We need to add the ability to map those two heights together to correctly associate the site and the footprint.

Option:
 - Use the "height_name" value from within the site_info.json file file (openghg/supplementary_data) to map this between the inle height for the site and the potential names for the inlet
 - Update the "height_name" value for WAO (at least) within the site_info.json file to be a dictionary of the form
    - "height_name": {"10m": ["10magl", ~~"21magl"~~ "20magl"]}

Other considerations:
 - Need to consider the naming of ~~"21magl"~~ "20magl" etc. since this doesn't map to how OpenGHG interprets this now. May want to run this through the `format_inlet()` converter function.
 
Correction: from looking at files available on BP1 these seem to be labelled as "20magl" rather than "21magl"

rt17603:
When the analyse functions (`ModelScenario(...)`) align observation data with available footprints this will now attempt to retrieve the 'height_name' definition from [site_info](https://github.com/openghg/supplementary_data/blob/main/openghg_defs/data/site_info.json) and use this to align different inlet values where appropriate.

This can handle 'height_name' definition as:
     -  list e.g. ["10magl"] or ["10magl", "100magl"] - list of footprint height values (should map to site inlet values)
     - dict e.g. {"10m": ["10magl", "20magl"]} - translation between site inlet and potential footprint inlet values

 - [x] Appropriate tests for reading and interpreting this data have been added.
 - [x] CHANGELOG.md updated

Closes #560.


openghg/analyse/_scenario.py
```diff
@@ -132,6 +133,7 @@ def __init__(
             domain : Domain name e.g. "EUROPE"
             model : Model name used in creation of footprint e.g. "NAME"
             metmodel : Name of met model used in creation of footprint e.g. "UKV"
+            fp_inlet : Specify footprint release height optionsif this doesn't not match to site value.
```
rt17603, line 136:
```suggestion
            fp_inlet : Specify footprint release height options if this doesn't not match to site value.
```

tests/analyse/test_scenario.py
```diff
@@ -307,6 +310,73 @@ def test_scenario_too_few_inputs():
...
+    assert model_scenario.footprint is None
+
+
+def test_scenario_matches_fp_inlet():
+    """
+    Test ModelScenario is able to use "height_name" data from site_info file to
+    map between different inlet values for observation data and footprints.
+
+    In this case for "WAO" data we want to be able to use a dictionary to allow
+    older footprints at which were run at 20m to be used for 10m inlet e.g.
```
rt17603, line 346:
```suggestion
    older footprints which were run at 20m to be used for 10m inlet e.g.
```

openghg/retrieve/_access.py
gareth-j, line 1: Can some more comments on what the code is doing be added to the `_metadata_difference` function? Maube it's just my morning brain but I'm struggling to follow it.

openghg/retrieve/_access.py
```diff
@@ -769,9 +769,13 @@ def _metadata_difference(
         if print_output:
             logger.info(f" {param}: ")
         for m in metadata:
-            summary_difference[param].append(m[param])
+            if param in m:
+                value: Optional[str] = m[param]
```
gareth-j, line 773: Can this be a `None`?

tests/data/info/site_info.json
```diff
@@ -416,24 +416,24 @@
   },
   "WAO": {
     "ICOS": {
-      "height": ["21m"],
-      "height_name": ["20magl"],
+      "height": ["10m"],
+      "height_name": {"10m": ["10magl", "20magl"]},
```
gareth-j, line 420: Should 20magl be here?

gareth-j (changes requested):
Just a couple of questions and maybe some extra comments needed on some bits of code.

tests/data/info/site_info.json
rt17603: No, this is a mapping between the inlet height (defined within "height" above) and the potential footprint height values. We want to map the "10m" inlet against both the "10magl" and "20magl" footprints i.e. first look for "10magl" then "20magl" but either can be used.

openghg/retrieve/_access.py
rt17603: It guess it depends on if it's possible for the parameter in the metadata can be None? Potentially not.

openghg/retrieve/_access.py
rt17603: I've added some more comments to lay out the steps for this function. It is mainly to check the differences between different metadata to allow that to me more easily fed back via output.

openghg/retrieve/_access.py
rt17603: This has been updated to remove Optional[]

openghg/retrieve/_access.py
gareth-j: OK, looks good, thanks

gareth-j (approved):
Looks good! Thanks for the extra comments.

Issue #661, opened 2023-05-26
Spelling mistake in docs - Devloper API Reference


gareth-j:
We've got a spelling mistake in on the home page of the docs - Devloper API Reference

achhayapathak:
I would like to work on this one

achhayapathak:
I have created a pull request, kindly review it.

PR #662, opened 2023-05-26
Updating index.rst [skip ci]

achhayapathak:
Correcting developer spelling which was earlier devloper in index.rst
This closes issue #661


gareth-j:
Hi @achhayapathak, we're actually opening some issues like that to help train some new people joining our team so unfortunately I can't merge this.

PR #663, opened 2023-05-30
Allowing ICOS internal attributes and data derived from site_info to be compared

### LINKED ISSUE #656: Make sure retrieved ICOS data can be linked against either attributes or site_info.json metadata
### What is your issue?

At the moment the ICOS internal attributes AND external metadata are built up using the details retrieved from the ICOS Carbon Portal rather than using the site_info.json details.

This needs to be updated to ensure the external metadata is linked against the site_info.json data. This update will mean the data from the ICOS Carbon Portal and from the site_info.json can be compared. If there is a mismatch this will raise an error by default but mismatches can be resolved by also passing the `update_mismatch=True` flag.

At the moment this will always pick the ICOS Carbon Portal data to resolve a mismatch (see closed Issue #613 and PR #625, `update_mismatch=True`) but there are plans to update this as laid out in Issue #638 to allow the user to choose which source to use (`update_mismatch="from_source"` - ICOS Carbon Portal or `"from_definition"` - site_info.json).

rt17603:
At the moment the `retrieve_atmospheric` function uses internal attributes to directly create the metadata. This update is to allow these internal values to checked against the site_info data and spot mismatches (these can be allowed to be updated if needed).

This also reduces the expected metadata keys for the ICOS data and adds `inlet` in preference to `sampling_height` to align with other source_formats for the `obs_surface` data type.

Closes #656

### DIFF
--- a/openghg/retrieve/icos/_retrieve.py
+++ b/openghg/retrieve/icos/_retrieve.py
@@ -323,7 +323,7 @@ def _retrieve_remote(
         # This is the metadata, dobj.info and dobj.meta are equal
         dobj_info = dobj.meta
 
-        metadata = {}
+        attributes = {}
 
         specific_info = dobj_info["specificInfo"]
         col_data = specific_info["columns"]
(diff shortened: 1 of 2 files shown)


gareth-j (approved):
Looks good!

PR #681, opened 2023-06-13
Update default CO2 scale to X2019

joe-pitt:
I noticed the default CO2 scale in the json files was X2007 - virtually everyone has moved to X2019 now so I have updated the defaults accordingly.


gareth-j (approved):
Thanks Joe!

rt17603:
This seems good to me - but should just quickly check with @mrghg that there are no problems with this update? This was to  change the default calibration scale in our internal files from X2007 to X2019?

rt17603 (approved):
Looks good

PR #687, opened 2023-06-16
Updated HiTRes Flux time array to be independent from footprints

### LINKED ISSUE #685: Using HiTRes Fluxes and Footprints from Different Years
### What are you trying to do?
Trying to do forward simulations of CO2 using the HiTRes datafiles. Specifically, using fluxes from 2014 combined with 2021 NAME footprints for Mace Head. The error is raised when using the `footprints_data_merge()` function from ModelScenario. 

### What do you expect to happen?
I expect the footprints to successfully merge with the flux files to produce the simulated mole fractions. This has previously worked for methane when using standard resolution footprints with flux files for different years

### What have you tried so far to debug this issue?
Briefly chatted with RT about this problem. The problem seems to arise in `_scenario.py` where the HiTRes fluxes use timestamps from the footprints (when performing data slicing) and since fluxes and footprints are from different times the problem is raised. 

### Do you have a proposed solution to the issue?
Modify the HiTRes blocks in `_scenario.py` so that the flux data are not relying on the footprints timestamps and the code mirrors what is done for the standard resolution footprints. 

### Anything else?
I'm happy to try modify this code using the proposed solution

EricSaboya:
HiTRes fluxes previously used the footprints time array for resampling in `_scenario.py`. Now, updated so HiTRes fluxes use their own time array for resampling. This should allow flux fields to be combined with footprints from different years in ModelScenario.  

- [ ] Closes #685 

### DIFF
--- a/openghg/analyse/_scenario.py
+++ b/openghg/analyse/_scenario.py
@@ -1160,6 +1160,8 @@ class ModelScenario:
         # create time array to loop through, with the required resolution
         # fp_HiTRes.time is the release time of particles into the model
         time_array = fp_HiTRes["time"]
+        time_array_flux = flux_ds["time"]
+
         lat = fp_HiTRes["lat"]
         lon = fp_HiTRes["lon"]
         hback = fp_HiTRes["H_back"]
@@ -1173,9 +1175,15 @@ class ModelScenario:
         max_h_back = int(hback.values[-1])
 
         # Define full range of dates to select from the flux input
+        # ____ Footprints time data ____
         date_start = time_array[0]
         date_start_back = date_start - np.timedelta64(max_h_back, "h")
         date_end = time_array[-1] + np.timedelta64(1, "s")
+
+        # ____ Flux time data ____
+        date_start_flux = time_array_flux[0]
+        date_start_back_flux = date_start_flux - np.timedelta64(max_h_back, "h")
+        date_end_flux = time_array_flux[-1] + np.timedelta64(1,"s")
 
         start = {
             dd: getattr(np.datetime64(time_array[0].values, "h").astype(object), dd)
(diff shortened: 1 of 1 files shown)


EricSaboya:
Most of the commits on this was me trying to get the linting to cooperate, darn commas and spacing

rt17603 (commented):
I think the main thing with this change it to see if you can run with this with your inputs and check if it's doing what you want it to do.

I would start by writing a notebook to try this out and check your outputs and then we could look at how to turn this into a test (e.g. see https://github.com/openghg/openghg/blob/devel/tests/analyse/test_scenario.py for the tests we have at the moment for ModelScenario)

openghg/analyse/_scenario.py
```diff
@@ -1173,10 +1175,16 @@ def _calc_modelled_obs_HiTRes(
         max_h_back = int(hback.values[-1])
 
         # Define full range of dates to select from the flux input
+        # ____ Footprints time data ____
         date_start = time_array[0]
         date_start_back = date_start - np.timedelta64(max_h_back, "h")
         date_end = time_array[-1] + np.timedelta64(1, "s")
 
+        # ____ Flux time data ____
+        date_start_flux = time_array_flux[0]
```
rt17603, line 1184: I have a question about how best to define `date_start_flux` - with the current setup the flux data read in will be whatever flux data is available in the object store matching the keywords (not based on input `start_date` and `end_date`) so if e.g. we have flux from e.g. 2010 - 2018 and you were comparing against 2023 - what would this set `date_start_flux` to? Would this grab the 2010 flux data for comparison?

rt17603:
@EricSaboya - did you have any thoughts on this and did you want us to take this on? I believe you said you were already making use of this but I do think we would need to just think through a few generalisation steps to make sure we're happy with covers the cases we want covered.

rt17603:
Closing for now as we need to address this using updated code - this is on the current milestone to address.

PR #696, opened 2023-06-30
432 update searchresults for pretty console output

### LINKED ISSUE #432: Update SearchResults use pretty print when print() function is called
Make printing out SearchResults use pretty print (i.e. how Jupyter notebooks prints pandas DataFrames)

SutarPrasad:
This PR contains a function that converts metadata from a data frame to a table using the rich package. 
e.g. calling "cfc_results.results" will show the table in the console formatted to adjust as per terminal.

* **Please check if the PR fulfils these requirements**

- [x] Closes #432  (Replace xxxx with the Github issue number)


rt17603 (changes requested):
This looks great (and the data provider team seem happy too!). I just have a few small updates to request before we merge this from looking over the checklist:

 - "Added an entry in the latest CHANGELOG.md file" - could you add a short line detailing the update in output style here to the CHANGELOG.md file. This may have to be within 0.6.1 as we'll likely merge this before that release is done (but I think that's ok).
 - "Added any new requirements to requirements.txt and recipes/meta.yaml" - this has been done for the "requirements.txt" file but we should also the `rich` package to the "environment.yml" as well (I'll make a note that we should update our PR checklist to include this).
 - *update* - I added an additional comment that we should catch the ModuleNotFoundError for the rich package and use the old style (see details within code).

After those small updates are made, please request an explicit re-review from me and I'll add an Approval for this to be merged.

openghg/dataobjects/_searchresults.py
```diff
@@ -249,3 +251,41 @@ def _retrieve_dataset(
...
+        """
+        Process the DataFrame and display it as a formatted table in the console.
+
+        Args:
+            df (DataFrame): The DataFrame to be processed and displayed.
+
+        Returns:
+            None
+        """
+        from rich import print
```
rt17603, line 266: Another small update I just considered is that we should probably catch the ImportError for the rich package if this is not present and include a bypass to use the old style. e.g. something like:

```suggestion
        try:
            from rich import print
            from rich.table import Table, box
        except ImportError:
            # Should likely add a logger line here but would also need to set up the logging
            print("Unable to use rich package to display search results. Please install rich to do so.")
            return None
```

SutarPrasad: > Another small update I just considered is that we should probably catch the ModuleNotFoundError for the rich package if this is not present and include a bypass to use the old style. e.g. something like:
> 
> ```
>         try:
>             from rich import print
>             from rich.table import Table, box
>         except ModuleNotFoundError:
>             # Should likely add a logger line here but would also need to set up the logging
>             print("Unable to use rich package to display search results. Please install rich to do so.")
>             return None
> ```

Won't the new release means re-running the requirements.txt?

rt17603 (approved):

brendan-m-murphy (approved):

Issue #333, opened 2022-05-12
Stop `Datasource.add_metadata` calling `to_lowercase`


gareth-j:
Lowercasing all the metadata blindly can break things like case sensitive URLs for ICOS. We should remove this and each standardisation module should be responsible for handling its own metadata. This could all be simplified by adding in case insensitive comparisons / search over the metadata.

PR #702, opened 2023-07-14
Iss701 - Fix lowercasing of metadata causing object store lookup issues

### LINKED ISSUE #701: upper case letters can't be used in paths to object stores
### What is your issue?

After pull request #664 was added, files cannot be retrieved from an object store if the path to the object store contains any capital letters. 

For example, `openghg.retrieve.search_flux()` claimed that there was no ._data file linked to the uuid it had retrieved (even if I could find the exact file in the repository it was searching for), because the path to the object store given in the uuid metadata had been converted to entirely lower case, so it was searching in the wrong location.

When I changed the name of my object store to be only lower-case and updated the path to this in my `.openghg/openghg.conf` file, this bug didn't happen and I could retrieve the data as expected.

I suspect this conversion from upper to lower case happens when an object is added to the store and the uuid is created, but at a first look I couldn't find where this occured in the code. But I could look into this in more detail later, if that would be helpful.

If it's standard practise to only include lower cases in paths to object stores, then this bug could be ignored and instead an error/warning could be output when the user tries to add an object store path with any captial letters.

gareth-j:
* **Summary of changes**
This fixes the issue with the object store path being lowercased in the metadata resulting in
paths with uppercase letters being invalid on case-sensitive systems. It does this by allowing the passing in
of a `skip_keys` value in the `BaseStore.assign_data` function which is passed down to `Datasource.add_metadata`.

- [x] Closes #701 

### DIFF
--- a/openghg/store/base/_base.py
+++ b/openghg/store/base/_base.py
@@ -130,8 +130,13 @@ class BaseStore:
             else:
                 datasource = Datasource.load(bucket=self._bucket, uuid=uuid)
 
+            # TODO - remove this when the lowercasing of metadata gets removed
+            # We currently lowercase all the metadata and some keys we don't want to change, such as paths to the object store
+            skip_keys = ["object_store"]
             # Add the dataframe to the datasource
-            datasource.add_data(metadata=meta_copy, data=_data, overwrite=overwrite, data_type=data_type)
+            datasource.add_data(
+                metadata=meta_copy, data=_data, overwrite=overwrite, data_type=data_type, skip_keys=skip_keys
+            )
             # Save Datasource to object store
             datasource.save(bucket=self._bucket)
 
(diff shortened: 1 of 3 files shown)


rt17603:
Was this linked to a specific problem you spotted?

gareth-j:
> Was this linked to a specific problem you spotted?

Yeah, @aliceramsden found an issue (#701) that meant the object store path was being lowercased when it was added to the Datasource. This meant that on systems with case sensitive paths it would fail. I've updated the tests to use some uppercase in the test paths and stopped the path being lowercased now using the `skip_keys` argument.

gareth-j:
> Looks good for now.
> 
> One consideration is that longer term we will likely need to address the uppercase/lowercase issues we've been having for metadata/attributes and make sure this is being solved in a consistent way.

Yeah I agree, we've actually had an issue open for this for almost a year (#333) and it's caused quite a few problems. We just need to make sure we can do a search of the metastore in a case insensitive way.

PR #711, opened 2023-07-20
Update search to add in object store data for backwards compatibility

### LINKED ISSUE #709: `object_store` key not present in metadata of older object stores
The new version `0.6.0` relies on `object_store` being in the metadata of the Datasource, this isn't the case.
We need a function to migtate the metadata stored.

gareth-j:

Adds in a check for the `object_store` key in the `Datasource` metadata in the search function, if it does exist it's checked, otherwise it's added.

- [x] Closes #709 

### DIFF
--- a/openghg/retrieve/_search.py
+++ b/openghg/retrieve/_search.py
@@ -613,9 +613,18 @@ def _base_search(**kwargs: Any) -> SearchResults:
         updated_metadata = {}
         store_data = {"object_store": bucket}
 
-        for uid, data in metadata.items():
-            data.update(store_data)
-            updated_metadata[uid] = data
+        for uid, _metadata in metadata.items():
+            if "object_store" in _metadata:
+                ds_bucket = _metadata["object_store"]
+                if ds_bucket != bucket:
+                    raise ObjectStoreError(
+                        f"Mismatch between Datasource object_store: {ds_bucket} and "
+                        + f" expected value: {bucket}."
+                    )
+            else:
+                _metadata.update(store_data)
+
+            updated_metadata[uid] = _metadata
 
         metadata = updated_metadata
 
--- a/openghg/retrieve/_search.py
+++ b/openghg/retrieve/_search.py
@@ -605,10 +605,19 @@ def _base_search(**kwargs: Any) -> SearchResults:
             logger.exception(msg=error_msg)
             raise ValueError(error_msg)
 
-        # Here we create a dictionary of the metadata keyed by the Datasource UUID
-        # we'll create a pandas DataFrame out of this in the SearchResult object
-        # for better printing / searching within a notebook
+        # Due to versions of OpenGHG < 0.6.0 not storing the object store
+        # in the metadata we nede to pass that information in here for backwards
+        # compatibility.
         metadata = {r["uuid"]: r for r in metastore_records}
+
+        updated_metadata = {}
+        store_data = {"object_store": bucket}
+
+        for uid, data in metadata.items():
+            data.update(store_data)
+            updated_metadata[uid] = data
+
+        metadata = updated_metadata
 
         # Narrow the search to a daterange if dates passed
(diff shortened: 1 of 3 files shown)


openghg/retrieve/_search.py
```diff
@@ -605,11 +605,20 @@ def _base_search(**kwargs: Any) -> SearchResults:
             logger.exception(msg=error_msg)
             raise ValueError(error_msg)
 
-        # Here we create a dictionary of the metadata keyed by the Datasource UUID
-        # we'll create a pandas DataFrame out of this in the SearchResult object
-        # for better printing / searching within a notebook
+        # Due to versions of OpenGHG < 0.6.0 not storing the object store
+        # in the metadata we nede to pass that information in here for backwards
```
rt17603, line 609:
```suggestion
        # in the metadata we need to pass that information in here for backwards
```

openghg/retrieve/_search.py
```diff
@@ -605,11 +605,20 @@ def _base_search(**kwargs: Any) -> SearchResults:
...
+        # compatibility.
         metadata = {r["uuid"]: r for r in metastore_records}
 
+        updated_metadata = {}
+        store_data = {"object_store": bucket}
```
rt17603, line 614: Is this adding in the "object_store" key for every object store this interacts with then? I'm just thinking if the "object_store" key is already present we wouldn't want this to be silently replaced here without us realising so may want to add a check if this is present before replacing it.

rt17603 (changes requested):
Question around behaviour when "object_store" key is already present in the metadata - is this always being replaced at the moment? If so, we should update this to be a little more cautious and not overwrite existing keys.

openghg/retrieve/_search.py
gareth-j: Yeah I'm going to add a check in there and make sure the values don't differ.

openghg/retrieve/_search.py
```diff
@@ -605,11 +605,29 @@ def _base_search(**kwargs: Any) -> SearchResults:
...
-        # Here we create a dictionary of the metadata keyed by the Datasource UUID
-        # we'll create a pandas DataFrame out of this in the SearchResult object
-        # for better printing / searching within a notebook
+        # Due to versions of OpenGHG < 0.6.0 not storing the object store
+        # in the metadata we need to pass that information in here for backwards
+        # compatibility.
         metadata = {r["uuid"]: r for r in metastore_records}
 
+        updated_metadata = {}
+        store_data = {"object_store": bucket}
```
rt17603, line 614: How is bucket being defined (and how do we want it to be defined) - at the moment it looks like it's just the last item in the loop on L589?

openghg/retrieve/_search.py
```diff
@@ -605,11 +605,29 @@ def _base_search(**kwargs: Any) -> SearchResults:
...
+
+        for uid, _metadata in metadata.items():
+            if "object_store" in _metadata:
+                ds_bucket = _metadata["object_store"]
+                if ds_bucket != bucket:
+                    raise ObjectStoreError(
+                        f"Mismatch between Datasource object_store: {ds_bucket} and "
```
rt17603, line 621: I'm a little cautious about adding in an error here that users won't be able to do anything about - any ideas why this error might come in and what advice we could give to fix this?

openghg/retrieve/_search.py
gareth-j: Yeah we iterate over the dictionary `readable_buckets` that's a store_name: store_path pair.

rt17603: But won't `bucket` just be set to the last value taken from `readable_buckets` from the loop? Is that what is expected?

rt17603: Ok - right this part is in the loop - I missed the nesting here.

PR #720, opened 2023-07-25
Iss683 hi t res retrieval

brendan-m-murphy:


brendan-m-murphy:
One of the changes I made was to use "high_time_resolution" instead of "high_time_res" for footprint standarisation and search. "resolution" is used instead of "res" in some other search functions, so I thought it would be best to have the same name everywhere.

This makes some of the tutorials a bit inaccurate, since they use "res" instead of "resolution" for footprints. I don't really want to change the tutorials yet, since there is that tutorial typos PR still open.

openghg/standardise/_standardise.py
```diff
@@ -395,8 +395,8 @@ def standardise_footprint(
             function_name="standardise", data=compressed_data, metadata=metadata, file_metadata=file_metadata
         )
 
-        fn_response = call_function(data=to_post)
-        response_content: Dict = fn_response["content"]
+        fn_resolutionponse = call_function(data=to_post)
```
rt17603, line 398: I think "res" here got accidentally replaced with "resolution" - same for the line below.

openghg/store/base/_base.py
```diff
@@ -176,16 +184,24 @@ def datasource_lookup(
         results = {}
         for key, _data in data.items():
             metadata = _data["metadata"]
-            required_metadata = {k.lower(): str(v).lower() for k, v in metadata.items() if k in required_keys}
+
+            def lower_if_string(val: Any) -> Any:
```
rt17603, line 188: Have a look at the [`to_lowercase` function](https://github.com/openghg/openghg/blob/devel/openghg/util/_strings.py#L71) within the openghg.util submodule (_strings.py file) to see if this covers this case.

If not, this functionality would be better off added there and imported to be used here.

tests/store/test_footprints.py
```diff
@@ -111,22 +112,22 @@ def test_read_footprint_standard(keyword, value):
...
 
-def test_read_footprint_high_spatial_res():
+def test_read_footprint_high_spatial_resolution():
     """
     Test high spatial resolution footprint
      - expects additional parameters for `fp_low` and `fp_high`
      - expects additional coordinates for `lat_high`, `lon_high`
      - expects keyword attributes to be set
-       - "spatial_resolution": "high_spatial_resolution"
+       - "spatial_resolution": "high_spatial_resolutionolution"
```
rt17603, line 130:
```suggestion
       - "spatial_resolution": "high_spatial_resolution"
```

tests/store/test_footprints.py
```diff
@@ -281,7 +283,7 @@ def test_read_footprint_co2(site, inlet, metmodel, start, end, filename):
      - expects additional parameter for `fp_HiTRes`
      - expects additional coordinate for `H_back`
      - expects keyword attributes to be set
-       - "spatial_resolution": "high_time_resolution"
+       - "spatial_resolution": "high_time_resolutionolution"
```
rt17603, line 286:
```suggestion
       - "spatial_resolution": "high_time_resolution"
```

CHANGELOG.md
```diff
@@ -5,6 +5,15 @@ All notable changes to OpenGHG will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased](https://github.com/openghg/openghg/compare/0.6.2...HEAD)
+
+## [0.6.2] - 2023-08-xx
+
+### Fixed
+
+- Bug where high time resolution footprints were being clobbered into low
```
rt17603, line 14: Maybe something like "Bug when adding footprints where this was unable to distinguish between high time and low time resolution footprints" may be a bit more clear?

openghg/retrieve/_search.py
```diff
@@ -243,46 +243,55 @@ def search_footprints(
...
                         Note this will be set to True automatically if species="co2" (Carbon Dioxide).
         short_lifetime: Indicate footprint is for a short-lived species. Needs species input.
                         Note this will be set to True if species has an associated lifetime.
         kwargs: Additional search terms
     Returns:
         SearchResults: SearchResults object
     """
+    # NOTE: the following two lines are a shorter way to get args...
+    # args = locals().copy()  # get dict of arguments to process and pass to `search`
+    # args.pop("kwargs")  # the value of kwargs is a dictionary, so we remove it
```
rt17603, line 257: You can try including this if you prefer and check the tests pass?

openghg/retrieve/_search.py
```diff
@@ -243,46 +243,55 @@ def search_footprints(
...
+    }
+
+    # Keys in metastore are stored as strings; convert non-string arguments to strings.
+    for k in ["start_date", "end_date"]:
+        if args[k] is not None:
+            args[k] = str(args[k])
+
+    # Either (or both) of 'inlet' and 'height' may be in the metastore, so
+    # both are allowed for search.
     from openghg.util import format_inlet
```
rt17603, line 284: Should keep the import statement at the top of the function call.

openghg/standardise/_standardise.py
```diff
@@ -395,8 +395,8 @@ def standardise_footprint(
...
+        response_content: Dict = fn_resolutionponse["content"]
```
rt17603, line 399:
```suggestion
        response_content: Dict = fn_response["content"]
```

rt17603:
> One of the changes I made was to use "high_time_resolution" instead of "high_time_res" for footprint standarisation and search. "resolution" is used instead of "res" in some other search functions, so I thought it would be best to have the same name everywhere.
> 
> This makes some of the tutorials a bit inaccurate, since they use "res" instead of "resolution" for footprints. I don't really want to change the tutorials yet, since there is that tutorial typos PR still open.

Yeah I understand what you're thinking, and we should prioritise getting that PR in if we can, but I think it may be better to just update the tutorials so we can finalise this PR and deal with the merge conflicts on the typos PR instead.

openghg/retrieve/_search.py
brendan-m-murphy: This made some tests fail, I'm not really sure why. I think a better model (to implement later) would be to have metadata objects that we pass between internal functions, and have the wrapper functions like `standardise_footprint` populate the metadata objects. Then `search_footprints` might just be creating a metadata object (which does formatting), and passing the data type and metadata object to `search'. (Or we could have a 'FootprintsMetadata' child class...) I guess the way I rewrote the function is a bit more like that logic, but I can revert it to the original version if you like.

brendan-m-murphy:
I addressed all of the comments. My responses above cover the major changes, except the tutorials, where I updated any uses of "high_time_res" etc, and updated the hardcoded output to match what is produced on this branch.

openghg/retrieve/_search.py
rt17603: I'm fine with how you rewrote the function below - your approach below seems great to me.

openghg/store/base/_base.py
```diff
@@ -126,7 +134,7 @@ def assign_data(
 
                 # Make sure all the metadata is lowercase for easier searching later
                 # TODO - do we want to do this or should be just perform lowercase comparisons?
-                meta_copy = to_lowercase(d=meta_copy, skip_keys=skip_keys)
+                # meta_copy = to_lowercase(d=meta_copy, skip_keys=skip_keys)
```
rt17603, line 137: Just to check - should this still be commented out after your changes around booleans / strings etc.?

rt17603 (commented):
Looking good - I just have one small question to be addressed included above.

openghg/retrieve/_search.py
```diff
@@ -254,35 +254,40 @@ def search_footprints(
     """
```
brendan-m-murphy, line 254: The import was moved to the top, it just doesn't show as a change because that's where it was originally

rt17603 (dismissed):
I'm adding an Approval so you're able to merge this but since it looks like this is going in before 0.6.1 is released please update the CHANGELOG.md to move your changes to 0.6.1 and delete the 0.6.2 block before merging.

SutarPrasad (approved):
looks good

PR #722, opened 2023-07-26
Storage suggestions

gareth-j:
* **Summary of changes**
This moves the Datasource versioning file handling towards the objectstore way of doing things. I think I've got the logic behind the versioning system right but do please check.

### DIFF
--- a/openghg/store/base/_datasource.py
+++ b/openghg/store/base/_datasource.py
@@ -2,6 +2,7 @@ from collections import defaultdict
 from typing import DefaultDict, Dict, List, Optional, Tuple, Type, TypeVar, Union
 import logging
 import numpy as np
+import shutil
 from pandas import DataFrame, Timestamp, Timedelta
 from xarray import Dataset
 from openghg.objectstore import exists, move_objects
@@ -664,7 +665,6 @@ class Datasource:
             new_keys = {}
 
             version_key = self.define_version_key(version_str)
-            # version_folder = Path(bucket) / version_key
 
             # If version is already present, and we are not creating a new version
             # we may need to delete the previous data.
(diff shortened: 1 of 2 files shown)


openghg/store/base/_datasource.py
```diff
@@ -701,44 +700,41 @@ def save(self, bucket: str, new_version: bool = True) -> None:
...
-                    if delete_version:
-                        version_folder_backup.rename(version_folder)
-                        self._data_keys.pop(version_str_backup)
-                    raise Exception("Unable to write new data. Restored previous data.")
+                    try:
+                        filepath.parent.mkdir(parents=True)
+                        data.to_netcdf(filepath, engine="netcdf4")
+                    except IOError:
+                        # If unable to write, return original data from back up.
+                        # NOTE - I wasn't sure how the os.removedir() was restoring the backup here?
```
rt17603, line 725: This was removing the new `version_folder` and partial contents which had been created first and then restoring the original details from the back up (`version_folder_backup`). The version folder had to be deleted first because you can't do the equivalent of:
```$ mv version_folder_backup version_folder```
if version_folder contains anything.

Looking at the new workflow, it's likely we'd still need that step in the code.

rt17603: Correct me if I've missed it but I'm not seeing the equivalent of `os.removedirs(version_folder)` to delete the folder and any contained data before the backup data is restored? Just checking as I know you said this had been added?

rt17603:
@gareth-j - were you able to take a look at the final changes for this? I can take over this PR if that's easier as I'd like to get into the branch and carry on along the chain.

openghg/store/base/_datasource.py
gareth-j: When we move data back to `version_key` the data at that key will get overwritten so we shouldn't need to delete it beforehand.

rt17603: As we discussed, this is related to the folder transfer problem. This is to cover the case where there's a been a problem creating the new version folder and we need to restore the backup. If we try to move a folder (e.g. version backup) to a non-empty folder (original version name) this will cause an error so to be safe we should delete what has been partially created.

Note that this is a fail case we're covering so it's to account for unexpected behaviour and to not leave the database in a half completed state.

rt17603 (approved):

PR #725, opened 2023-07-27
Fix for object_store not being present in Datasource metadata

### LINKED ISSUE #709: `object_store` key not present in metadata of older object stores

### LINKED ISSUE #723: Remove `object_store` from Datasource metadata
Shall we just remove the `object_store` key from the Datasource metadata?
I can see it causing issues if a user moves their store or changes the foldername.
We should just be able to read Datasources from an object store without this information using the search function.
We can then add in the object store path to the metadata on retrieval.

gareth-j:

Removes the `object_store` key from the Datasource metadata to avoid issues when folders are moved etc.
The object store key is now added to the metadata presented to the user by the `SearchResults` object.

- [ ] Closes #709 
- [ ] Closes #723 

### DIFF
--- a/openghg/dataobjects/_datamanager.py
+++ b/openghg/dataobjects/_datamanager.py
@@ -14,6 +14,14 @@ logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handle
 
 class DataManager:
     def __init__(self, metadata: Dict[str, Dict], store: str):
+        # We don't want the object store in this metadata as we want it to be the
+        # unadulterated metadata to properly reflect what's stored.
+        for m in metadata.values():
+            try:
+                del m["object_store"]
+            except KeyError:
+                pass
+
         self.metadata = metadata
         self._store = store
         self._bucket = get_writable_bucket(name=store)
@@ -66,6 +74,11 @@ class DataManager:
 
         uuids = list(self.metadata.keys())
         res = search(uuid=uuids)
+        # We don't want the object store in this metadata as we want it to be the
+        # unadulterated metadata to properly reflect what's stored.
+        for m in res.metadata.values():
+            del m["object_store"]
+
         self.metadata = res.metadata
 
     def restore(self, uuid: str, version: Union[str, int] = "latest") -> None:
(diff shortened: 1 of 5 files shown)


openghg/dataobjects/_datamanager.py
```diff
@@ -66,6 +74,11 @@ def refresh(self) -> None:
 
         uuids = list(self.metadata.keys())
         res = search(uuid=uuids)
+        # We don't want the object store in this metadata as we want it to be the
+        # unadulterated metadata to properly reflect what's stored.
+        for m in res.metadata.values():
+            del m["object_store"]
```
rt17603, line 80: Would this raise an error if the "object_store" key isn't in this metadata?

gareth-j: OK I've fixed this by creating a cleaning function within the `DataManager` class.

rt17603 (approved):
Looks good. A few small suggestions (but not blockers to being merged)

 - I added a small suggested modification to the xfail to make sure we don't forget about putting this back in
 - @brendan-m-murphy made a suggestion on the related issue #725 "If there is logging info when a datasource is created, we could say what object store the datasource was added to. (E.g. "created in {object_store} object store".)" - may be something useful to add?

rt17603 (approved):

rt17603 (approved):
Looks good.

gareth-j:
> Looks good. A few small suggestions (but not blockers to being merged)
> 
>     * I added a small suggested modification to the xfail to make sure we don't forget about putting this back in
> 
>     * @brendan-m-murphy made a suggestion on the related issue [Fix for object_store not being present in Datasource metadata #725](https://github.com/openghg/openghg/pull/725) "If there is logging info when a datasource is created, we could say what object store the datasource was added to. (E.g. "created in {object_store} object store".)" - may be something useful to add?

OK cool, I've opened https://github.com/openghg/openghg/issues/736 for the Datasource logging idea.

Issue #451, opened 2022-10-14
Check output of ModelScenario methods match to footprints_data_merge function (within acrg code)


rt17603:
For multiple input scenarios, check values created by the `openghg.analyse.ModelScenario.footprints_data_merge` function match to values from `acrg.name.footprints_data_merge` function in the acrg code. Run through both with the exact same data.

- [ ] methane data (for 1 site)
- [ ] carbon dioxide data (for 1 site)
- [ ] radon data (for 1 site)

Check `time`, `mf_mod` and `bc`/`bc_mod` for all cases.

rt17603:
See  #637 for an example for methane to check. That issue can be closed when this has been addressed.

rt17603:
That's great - thanks for much for looking into this in more detail Alice. 

To address your proposed solutions, there was a reason for extracting all flux data related to how the carbon dioxide methodology works and how this needs to match to dates prior to the start date so I would suggest option 1 may be the best approach to get the best of both worlds here.

Would you be confident / have time to try putting in a PR for this and I can take a look? I can always add to this if we wanted to try and add in additional details / test as well.

aliceramsden:
Ok, I did expect there might be a good reason why the whole year of flux was read in, thanks for clarifying that.

I can put together a pull request with a fix using option 1, and I'll test it out with a few different sets of inputs (I'll see how I get on with adding a coded test for this too, if I've got time).

brendan-m-murphy:
I'm not sure if this is related, but the `combine_datasets` function in inversions uses "ffill" and it tends to put nan's in the the values whose lat/lon have index = 0. So that might be another place where "nearest" makes more sense (or we need some extra code to replace that first value). (There is a similar `combine_datasets` function in the "scenario" module in OpenGHG.)

rt17603:
The 'ffill' is needed for the time dimension and I'm guessing that the reindexing is currently being done across all coordinates at the same time (time, lat, lon) this will be why this is being used. This could be updated to split out the reindex to time and the reindex to lat, lon to do what you suggested but (a) we need to make sure that's not much less efficient and (b) we must used ffill for time.

brendan-m-murphy:
In the specific case I looked at (test data added to inversions), only the time dimension needed to be filled, and the other dimensions were already aligned. By the time `fp_sensitivity` was called, there were about 100k nan's in `fp_data` that didn't need to be there (for 1 week of observations).

[...]

PR #750, opened 2023-08-15
Iss451 flux for mod mf bc

### LINKED ISSUE #637: ModelScenario for forward modelling without observations
### What is your issue?

When using ModelScenario to produce forward modelled mole fractions, I have run into two potential issues:

1. When running ModelScenario with the obs parameter set to None, calc_modelled_obs() and calc_modelled_baseline() are not run, they have to be run again manually on the ModelScenario object. For example:

    `model_scenario = ModelScenario(obs=None,
                                    footprint=footprint_dict[site],
                                    flux=flux_dict,
                                    bc=bc)`
    
    just produces the object containing the data files, without any modelled mole fractions.

2. When using the get_flux() function in _scenario.py, the entire flux file is read in, rather than the correct timeslice specified by start_date and end_date. For example:

    `get_flux_data = get_flux(species=species,
                                domain=domain,
                                source=source,
                                start_date='2018-01-01',
                                end_date='2018-02-01')`
    
    extracts the whole flux file containing a year of data, rather than the specified month. 
    
    There is then a mismatch between the output of calc_modelled_obs() from a ModelScenario object with this FluxData and the modelled obs produced by the same process in the ACRG code repository. This may be caused by how the fluxes are selected/averaged by ModelScenario before the footprint * flux calculation.

I tested the above using these inputs, but found the same issues with a range of different sites:

`site = 'TAC'`
`inlet = '185magl'`
`species = 'ch4'`
`fp_species = None`
`bc_species = 'ch4'`
`domain = 'EUROPE'`
`fp_model = None`
`met_model = 'UKV'`
`start_date = '2021-01-01'`
`end_date = '2021-02-01'`
`flux_start_date = '2018-01-01'`
`flux_end_date = '2018-02-01'`
`bc_start_date = '2018-01-01'`
`bc_end_date = '2018-02-01'`

aliceramsden:
I've changed how fluxes and boundary conditions (bc) are used when creating modelled mole fractions. If the flux or bc file is from a different year to the observations, instead of just reading in the most recently available month of fluxes/bc, fluxes/bc are read in from a month that matches the observations. This is useful when using monthly inputs for recent inverse modelling years, when more recent fluxes/bc are not available.

I had modified my copy of the ACRG repository to do this a while ago, which was why there were mismatches between the modelled mole fractions created from ACRG code and OpenGHG.

A test has been added for the monthly flux inputs, but I've run out of time to do this for the monthly bc. This could be replicated based on what I've added for the fluxes.

Similarly, I've not had time to get black, flake8 etc. working correctly for me. If the reviewer could check on these for me, that would be appreciated!

- [x] Closes #637 


aliceramsden:
I've just remembered that I haven't tested that changes are still compatible with annual flux files (where there's only one timestamp in the flux file). Theoretically the tests should catch this, but it should probably be checked manually too with some real data, before this is merged in fully.

rt17603:
@aliceramsden - I know they're related but I'm wondering if this actually closes Issue #637 rather than fully closing Issue #451?

brendan-m-murphy (changes requested):
Changes requested. This PR is conflicted with devel and its ModelScenario implementation is not safe to revive as-is. In _calc_modelled_obs_integrated it selects one flux slice using the last available flux year and only the first scenario month, then broadcasts it with nearest across the scenario; calc_modelled_baseline does the same for BC. A Jan–Mar scenario with monthly values [2, 3, 4] therefore produces [2, 2, 2], and cross-year scenarios can silently use the wrong year.\n\nThe added test uses single-month scenarios with deliberately mismatched data years, so it does not exercise this error. Current devel now routes this behavior through fp_x_flux_integrated/baseline_sensitivities; please close/supersede this PR in favour of that current path, or rebase and add multi-month and cross-year regression coverage before reconsidering.

brendan-m-murphy:
@rt17603 The PR has been rebased onto current devel and the obsolete ModelScenario implementation removed. The new single commit adds shared, guarded single-reference-year calendar alignment for integrated flux and BC, with monthly Dec→Jan, annual, and BC regressions. Focused tests, Ruff, formatting, and mypy pass locally; please re-review once the CI checks complete. The remaining policy assumption is that a non-overlapping source spanning exactly one year is a climatology.

openghg/analyse/_modelled_obs.py
```diff
@@ -23,7 +23,7 @@ def fp_x_flux_integrated(footprint: xr.Dataset, flux: xr.Dataset) -> xr.DataArra
     flux = reindex_on_dims(flux, footprint, ["lat", "lon"])
     # align separately on time
     # TODO: if method="nearest" was acceptable, then we could align all coordinates at once with reindex_like
-    flux = flux.reindex_like(footprint, method="ffill")
+    flux = reindex_time_with_climatology(flux, footprint)
```
rt17603: A few things here:
 - This function names suggests to me that this will always apply a climatology which I don't believe is the case here so I think this would be better as something like `reindex_time_to_target` (Claude suggestion).
- As this function allows a climatology to be applied this should be configurable behaviour so I think adding a flag for allowing climatology would be better. This should be an option which can be passed through from the original function calls. This can be set to True by default though.

Summary suggestion:
```suggestion
    flux = reindex_time_to_target(flux, footprint, allow_climatology=True)
```

openghg/analyse/_alignment.py
```diff
@@ -156,6 +157,36 @@ def timedelta_to_hourly_freq(td: pd.Timedelta) -> str:
 T2 = TypeVar("T2", xr.DataArray, xr.Dataset)
 
 
+def reindex_time_with_climatology(source: T1, target: T2) -> T1:
```
rt17603, line 160: A few things here:
- This function names suggests to me that this will always apply a climatology which I don't believe is the case here so I think this would be better as something like reindex_time_to_target (Claude suggestion).
- As this function allows a climatology to be applied this should be configurable behaviour so I think adding a flag for allowing climatology would be better. This should be an option which can be passed through from the original function calls. This can be set to True by default though.

From this one suggestion would be:
```suggestion
def reindex_time_to_target(source: T1, target: T2, allow_climatology: bool = True) -> T1:
```

openghg/analyse/_alignment.py
```diff
@@ -156,6 +157,36 @@ def timedelta_to_hourly_freq(td: pd.Timedelta) -> str:
...
+    source_times = pd.DatetimeIndex(source.time.values)
+    target_times = pd.DatetimeIndex(target.time.values)
+
+    if source_times.min() <= target_times.max() and target_times.min() <= source_times.max():
+        return source.reindex(time=target.time, method="ffill")
+
+    if source_times.year.nunique() != 1:
+        return source.reindex(time=target.time, method="ffill")
+
+    reference_year = source_times.year.max()
```
rt17603, line 177: Suggest adding in a clause here to continuing when `allow_climatology` is True but returning an error otherwise.

openghg/analyse/_modelled_baseline.py
```diff
@@ -29,9 +30,9 @@ def baseline_sensitivities(bc: xr.Dataset, fp: xr.Dataset, species: str | None =
         if dim in fp.dims and "units" not in bc[dim].attrs:
             bc[dim].attrs["units"] = fp[dim].attrs.get("units")
 
+    bc = reindex_time_with_climatology(bc, fp)
```
rt17603, line 33: Suggest update to more generic function name and allow_climatology flag to be passed

We could make this more complex by allowing climatology to be specified for flux and bc separately and to be resolved here or in a function higher up the stack. Depends on how much extra work this would be.

openghg/analyse/_modelled_obs.py
rt17603, line 26: Again, suggest update to more generic function name and allow_climatology flag to be passed

Again, we could make this more complex by allowing climatology to be specified for flux and bc separately and to be resolved here or in a function higher up the stack. Depends on how much extra work this would be.

tests/analyse/test_alignment.py
rt17603, line 1: Would be good to see at least one full example of the workflow with a climatology file, especially with the option to specify whether climatology should be included.

rt17603 (commented):
This update looks good - thanks for the work to rebase and get this aligned with the new split out function structure around analyse.

As you say, looks like this currently assumes if the dates within bc or flux do not match to the current year for footprints that this should be interpreted as representative of a year cycle and used as a climatology - that seems reasonable but want to make sure that change is very clear.

Regardless for this PR a few things:
1. I think we should make sure this is configurable behaviour from the ModelScenario layer (or wherever seems the most appropriate)
2. I think we should make the function name a bit more generic
3. Based on (1) it would be good to make sure there is at least one test which runs the whole ModelScenario pipeline to check this works across layers (perhaps run the negative to check an error is raised).

Issue #747, opened 2023-08-15
Check if data to be added is the same as already stored


gareth-j:
We should check if the data we're adding is the same as already stored. If a file header changes then the hash will be different but the stored data could be exactly the same. This is from #554 and was to be added to `Datasource.add_timed_data`.

```
            # Check the equality of what we've got at the moment and what
            # we're going to add, doing a simple dict eqality comparison
            # isn't enought to check the contents of the Datasets here
            if new_data == self._data:
                for _old, _new in zip(self._data.values(), new_data.values()):
                    if not _old.equals(_new):
                        break
                else:
                    logger.warning("No new data, nothing to do.")
                    return None
```

rt17603:
Do we still want to do this?

Issue #591, opened 2023-03-07
How should `Datasource` handle data and `overwrite=True`


gareth-j:
How do we want `Datasource` to handle new data being added? Do we always want data to be combined when new data is added?

rt17603:
Scenario 1
 - New data added
   - Add new data source
   - version 1 added

Scenario 2
 - New data, non-overlapping date range
   - Combine with previous data source (check overlap)
   - new version assigned, version 2

Scenario 3a
 - New data, overlapping date range / replacement date range
   - [Currently: Add data check for overlapping, try combining and drop duplicates]
   - Don't add

Scenario 3b
 - New data, overlapping date range / replacement date range + overwrite
  - "Retire previous data to previous data"

Scenario 3c
 - New data, overlapping date range / replacement date range + overwrite + update
   - "Replace overlapping date range" - delete old, insert new

rt17603:
Caveats: need to make sure NaN values and data not being present is treated differently. NaN should be replaced and then dropped rather than just ignored.

gareth-j:
I'm going to link this with #319 as the ability to pass in these flags will be helped by a rework of `assign_data`.

rt17603:
Alternatively, could set this up as `mode` option inputs e.g.
 - mode = None - default
 - mode = "overwrite" - overwrite description above
 - mode = "update" - update description above

IF overwrite / update are essentially mutually exclusive

rt17603:
Check and close #465 as appropriate in relation to this.

rt17603:
@joe-pitt - if you could take a look at this workflow let me know if this does/doesn't address the question you had around updating data in the object store.

joe-pitt:
Yes I think on_overlap input = "replace" will be useful for data that gets regularly updated (e.g. NRT data). I guess we can decide between us at some later point exactly which data should be "retired" (with a version number) and which should be replaced, but it would be good to have the functionality to do both.

rt17603:
I'm considering what should be done for Scenario 2 (Previous data source, new data added, non-overlapping date range) when the flags are passed and I think this may be better as an `if_exists` input rather than `on_overlap` so something like:
 - `if_exists = "new"` always create a new version
 - `if_exists = "replace"` replace any overlapping existing data and insert into current version

rt17603:
From discussion on 05/06/2023 with Joe Pitt:

Better to consider this as two separate questions:
1. How do we want to combine the new data with any current data?
[...]

PR #752, opened 2023-08-16
Make metadata store source of truth for Datasource UUIDs

### LINKED ISSUE #751: Make the metadata store the single source of truth for Datasource UUIDs
@joe-pitt has encountered an error on BP1 where for some reason the dictionary of Datasource UUIDs has been cleared for each of the storage classes. I'm not sure how this has happened but it prevented the removal of data from the object store using the `DataManager`.

I think we should remove the storage of Datasource UUIDs from the base storage class and just read the data in when the class is instantiated.

gareth-j:

This makes the metadata store for each storage class the source of truth for `Datasource` UUIDs. This will prevent synchronisation issues.

- [x] Closes #751 


brendan-m-murphy (dismissed):
Looks good to me!

brendan-m-murphy (dismissed):
I reverted the changes I made previously, since the long running time on ES_store was due to a flux file with 8700 versions.

gareth-j:
It's crazy there were that many versions. We should ask Eric what he was running when that happened. I wonder if the code in issue [#✈️](https://github.com/openghg/openghg/issues/747) could have prevented the creation of that many versions? I think Rachel's new versioning code might help stop issues like that.

rt17603 (commented):
Quick question before this gets merged: Are we assuming the "UUID" now is essentially the code/details used to locate the data? In which case there's no point internally storing it associated with the data?

brendan-m-murphy:
> > It's crazy there were that many versions. We should ask Eric what he was running when that happened. I wonder if the code in issue [#✈️](https://github.com/openghg/openghg/issues/747) could have prevented the creation of that many versions? I think Rachel's new versioning code might help stop issues like that.
> 
> Yeah - it would be useful to know what process caused that! In the new versioning it would raise an Error the second time it tried to add so harder to do by accident (but it would be possible with the correct flags).

I think what was happening was that some of the files being uploaded were for very short time frames, so a month of data required over 100 calls to standardise_flux. I was just poking around in ipython on BP1, I could get the details again if you like.

I think something we'll need to deal with is adding data in smaller amounts than our chunk sizes. These shouldn't be "revisions" if the new data doesn't overlap the old data. So we need to check something more specific than "this new data's time period overlaps with the chunk time period".

rt17603:
> I think something we'll need to deal with is adding data in smaller amounts than our chunk sizes. These shouldn't be "revisions" if the new data doesn't overlap the old data. So we need to check something more specific than "this new data's time period overlaps with the chunk time period".

It sounds like this is what has already been implemented in PR #653 which is currently on the `new_object_store` parallel branch - see also Issue #591

brendan-m-murphy:
> Quick question before this gets merged: Are we assuming the "UUID" now is essentially the ID used to locate the data? In which case there's no point internally storing it in a place associated with the data but better to only store it in a central location as a reference (i.e. the metastore)?

As in, the datasources don't need to save their UUID? Some of the datasource metadata was useful in figuring out what happened with the 8000 versions. So some kind of logging for datasources is useful, but maybe not the UUID and bucket, since you need to know those to load the Datasource.

rt17603:
> As in, the datasources don't need to save their UUID?

Yes - that seems like the purpose for this PR if I'm understanding the summary - the Datasource no longer internally stores it's UUID but I'm just checking that doesn't matter as the UUID is how the Datasource is retrieved anyway. Just ensuring I understand the logic being included here.

brendan-m-murphy:
> > I think something we'll need to deal with is adding data in smaller amounts than our chunk sizes. These shouldn't be "revisions" if the new data doesn't overlap the old data. So we need to check something more specific than "this new data's time period overlaps with the chunk time period".
> 
> It sounds like this is what has already been implemented in PR #653 which is currently on the `new_object_store` parallel branch - see also Issue #591

In lines 190-197 in _datasource.py in PR #653: if there are duplicate times, this should force a new version/revision, or raise an error if the person adding the data hasn't indicated that they are revising data (or maybe have an option that allows adding but not overwriting... so the new duplicates are dropped).

Also, if you do line 197 no matter what, you don't need line 190. (Not sure it's much of a performance gain...)

As it is, I think the code in PR #653 means that you must overwrite if the new data date range overlaps with the date range of an existing chunk. I think the default should be, allow this without overwrite as long as there are no duplicate timestamps. If there are duplicate time stamps, then either the revision number must increase, or the new timestamps should be dropped (although maybe just an error would be better...)

brendan-m-murphy:
...it's possible that I missed something in how the date ranges are represented in the existing data, so that the "overlapping date range" scenario only happens if you are trying to modify existing data. But I don't think that's how it works in v6.2 at least.

rt17603:
> The Datasources still know their own UUID as it's part of the JSON they save/load themselves from. This just stops the storing of the Datasource UUIDs in the storage classes themselves. Datasource UUIDs are now just stored in the metadata store.

Ok understood - thanks for clarifying.

rt17603:
> In lines 190-197 in _datasource.py in PR #653: if there are duplicate times, this should force a new version/revision, or raise an error if the person adding the data hasn't indicated that they are revising data (or maybe have an option that allows adding but not overwriting... so the new duplicates are dropped).

Yes that's what it's doing.

> Also, if you do line 197 no matter what, you don't need line 190. (Not sure it's much of a performance gain...)

Details around checking the overlaps is being changed - work in progress (see #591 for intent here)

> As it is, I think the code in PR #653 means that you must overwrite if the new data date range overlaps with the date range of an existing chunk. I think the default should be, allow this without overwrite as long as there are no duplicate timestamps. If there are duplicate time stamps, then either the revision number must increase, or the new timestamps should be dropped (although maybe just an error would be better...)

We would want to carefully consider whether we should be collating data in this way (especially without proper signalling from the user). Adding data before the start or after the end is relatively safe but collating data could fundamentally change the data timeseries included.

brendan-m-murphy:
> Yes this isn't in the current devel or master version yet as it's sitting on the parallel [`new_object_store` branch](https://github.com/openghg/openghg/tree/new_object_store) - this is part of the ["Special Release" milestone](https://github.com/openghg/openghg/milestone/13).

It seems that `get_representative_daterange_str` is defined the same in both devel and the special branch. I guess I misunderstood how chunking by year works.

So I guess what I thought was happening with that flux datasource with 8000 versions was incorrect. I suppose it's possible that when the data was split into small chunks to upload, the end points of the small chunks overlapped.

brendan-m-murphy (approved):
Looks good!

rt17603:
> > Yes this isn't in the current devel or master version yet as it's sitting on the parallel [`new_object_store` branch](https://github.com/openghg/openghg/tree/new_object_store) - this is part of the ["Special Release" milestone](https://github.com/openghg/openghg/milestone/13).
> 
> It seems that `get_representative_daterange_str` is defined the same in both devel and the special branch. I guess I misunderstood how chunking by year works.
> 
> So I guess what I thought was happening with that flux datasource with 8000 versions was incorrect. I suppose it's possible that when the data was split into small chunks to upload, the end points of the small chunks overlapped.

In the current code, I believe it's just that every time data is uploaded, if there's already a Datasource present a new version will have been created (even if it's identical to the previous version).

rt17603:
> In the current code, I believe it's just that every time data is uploaded, if there's already a Datasource present a new version will have been created (even if it's identical to the previous version).

Scratch that - that's not quite true.
 - If the file is identical (i.e. same hash) and overwrite=False it won't be re-added
 - If overwrite=True it will always add a new version even if identical
 - If overwrite=False and there's any small change in the new data then a new version will be created

I suspect it's the middle case which may be linked to the 8000 versions.

brendan-m-murphy:
> > In the current code, I believe it's just that every time data is uploaded, if there's already a Datasource present a new version will have been created (even if it's identical to the previous version).
> 
> Scratch that - that's not quite true.
> 
> * If the file is identical (i.e. same hash) and overwrite=False it won't be re-added
> * If overwrite=True it will always add a new version even if identical
> * If overwrite=False and there's any small change in the new data then a new version will be created
> 
> I suspect it's the middle case which may be linked to the 8000 versions.

Maybe, but it looks like 7394 different files were added to the flux with UUID 1349a7aa-fbb8-49c9-859f-65e40d9901e7 (which has the key 'co2_cte-hr-fossil_europe'). So most of the files were new apparently. But this is the last case? So it would create a new version anyway.

brendan-m-murphy:
These are the dates associated with the first 50 files uploaded to that flux datasource (assuming they were added in increasing time order): 

'2021-01-20T02:00:00',
 '2021-04-13T12:00:00',
 '2021-08-15T06:00:00',
 '2021-09-15T03:00:00',
 '2021-02-06T23:00:00',
 '2021-07-11T23:00:00',
 '2021-01-29T14:00:00',
 '2021-06-18T19:00:00',
 '2021-10-12T22:00:00',
 '2021-08-07T04:00:00',
 '2021-05-18T18:00:00',
 '2021-09-11T01:00:00',
 '2021-07-10T18:00:00',
 '2021-04-04T14:00:00',
 '2021-08-03T10:00:00',
 '2021-11-29T12:00:00',
 '2021-08-18T09:00:00',
 '2021-01-10T14:00:00',
 '2021-01-27T15:00:00',
 '2021-09-25T23:00:00',
 '2021-10-06T23:00:00',
 '2021-06-14T19:00:00',
 '2021-04-18T13:00:00',
 '2021-01-03T18:00:00',
 '2021-12-14T13:00:00',
 '2021-10-10T03:00:00',
 '2021-09-07T17:00:00',
 '2021-05-13T08:00:00',
 '2021-02-20T04:00:00',
 '2021-11-28T18:00:00',
 '2021-07-03T22:00:00',
 '2021-10-26T21:00:00',
 '2021-07-05T05:00:00',
 '2021-02-06T08:00:00',
 '2021-02-04T21:00:00',
 '2021-01-30T01:00:00',
 '2021-08-22T02:00:00',
 '2021-11-08T13:00:00',
 '2021-07-16T23:00:00',
 '2021-12-24T15:00:00',
 '2021-05-24T04:00:00',
 '2021-09-11T22:00:00',
 '2021-03-03T17:00:00',
 '2021-01-27T22:00:00',
 '2021-05-30T17:00:00',
 '2021-10-19T16:00:00',
 '2021-11-01T02:00:00',
 '2021-08-31T04:00:00',
 '2021-10-08T14:00:00',
 '2021-05-29T14:00:00'

The files had names like: 'paris_flux_ff_prior_EUROPE_2021-01-20T02:00:00.000000000.nc'

rt17603 (approved):

PR #755, opened 2023-08-16
Iss749 add compression

### LINKED ISSUE #454: Add option to lazy load data from object store
We could add an option to lazily load data from the object store. Large datasets such as footprints / flux aren't very big when compressed as a NetCDF. We could just return the bytes from the object store instead of doing a `load_dataset` in Datasource. We could then mayeb use an BytesIO buffer to open the data in chunks when needed.

### LINKED ISSUE #749: Add compression to footprints while waiting for zarr
### What is your issue?

Right now footprints aren't being compressed. When we switch to zarr as a storage backend, we will get compression back, but in the short term we're having trouble adding footprints to the shared object store.

We can use this work around to compress netcdf: [https://github.com/openghg/openghg_inversions/pull/19](https://github.com/openghg/openghg_inversions/pull/19).

Also, we could run a script on Blue Pebble to compress existing files once the workaround is tested.

brendan-m-murphy:

1. Compression level for saving Datasources is increased from 4 to 5.
(Compression level 4 seems to be the default.) I added a test to show
that increasing compression does actually result in small files on disk.

2. `Datasource.load_dataset` now uses `xarray.open_dataset` instead
of `xarray.load_dataset`, which should lazy-load data. This gives a 
33% decrease in the time taken to run all of the tests on my machine.

I couldn't make a useful test to demonstrate that data is lazy-loaded,
but the xarray docs state that this is the case:

https://docs.xarray.dev/en/latest/user-guide/io.html#reading-encoded-data

(They say data is "always lazy loaded from netcdf", although the docs for
`load_dataset` say the opposite.)

*If* `load_dataset` is putting all data in memory, and `open_dataset` is
lazy-loading data, this should drastically reduce the amount of memory needed
to add footprints to the object store.

- [x] Closes #749 (partially... I suggested using a script to compress anything already in the object store, but maybe it won't matter too much)
- [x] Closes #454 

### DIFF
--- a/openghg/store/base/_datasource.py
+++ b/openghg/store/base/_datasource.py
@@ -615,10 +615,12 @@ class Datasource:
                             encoding[var] = enc
 
                     try:
+#                        breakpoint()
                         data.to_netcdf(filepath, engine="netcdf4", encoding=encoding)
                     except RuntimeError:
+                        raise
                         logger.warning(
-                            "Storing footprint for date range {daterange} without compression due to netCDF4 RuntimeError."
+                            f"Storing footprint for date range {daterange} without compression due to netCDF4 RuntimeError."
                         )
                         data.to_netcdf(filepath, engine="netcdf4")
                     except ValueError as e:
--- a/openghg/store/base/_datasource.py
+++ b/openghg/store/base/_datasource.py
@@ -615,7 +615,6 @@ class Datasource:
                             encoding[var] = enc
 
                     try:
-#                        breakpoint()
                         data.to_netcdf(filepath, engine="netcdf4", encoding=encoding)
                     except RuntimeError:
                         logger.warning(
--- a/openghg/store/base/_datasource.py
+++ b/openghg/store/base/_datasource.py
@@ -1,6 +1,7 @@
 from pandas import DataFrame, Timestamp, Timedelta
 from xarray import Dataset
 from collections import defaultdict
+import re
 from typing import DefaultDict, Dict, List, Optional, Tuple, Type, TypeVar, Union
 import logging
 import numpy as np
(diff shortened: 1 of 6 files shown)


openghg/store/base/_datasource.py
```diff
@@ -577,8 +579,60 @@ def save(self, bucket: str) -> None:
...
+                        if var not in do_not_compress:
+                            enc = {k: v for k, v in data[var].encoding.items() if k in valid_encoding_keys}
+                            enc.update(comp)
+                            encoding[var] = enc
 
+                    try:
+                        data.to_netcdf(filepath, engine="netcdf4", encoding=encoding)
+                    except RuntimeError:
+                        logger.warning(
+                            "Storing footprint for date range {daterange} without compression due to netCDF4 RuntimeError."
```
gareth-j, line 621: This needs to be an f-string

gareth-j: I'm getting some of these RuntimeErrors when I run the tests, can we have a look at what's causing them?

gareth-j (changes requested):
Just a couple of questions and a quick fix for the f-string issue.

openghg/store/base/_datasource.py
brendan-m-murphy: I tried tracing what goes wrong. The variable that causes the problem is `rn_number_of_observations` for the WAO data in one of the fixtures in tests/analyse/conftest.py. This variable doesn't have chunking set... I'm not sure why it raises an issue. The error comes from the netcdf-c code base, and it's a pain to try to find where things are actually defined. For some reason the error comes from a function that has to do with chunking, but there is no chunking... so I have no idea what is wrong.

gareth-j (approved):
Looks good.

rt17603:
@brendan-m-murphy - just to check - will these updates work ok with the current object store?

brendan-m-murphy:
@rt17603 Yea I just branched off of devel. The lazy loading changing is just replacing `load_dataset` with `open_dataset`. The xarray docs say this should give you lazy loading. (I also replaced the binary IO part with just reading from a file, maybe that would be a problem for running in the cloud?) Actually the xarray docs say that they always use lazy loading, but the docs for `load_dataset` say it doesn't, while `open_dataset` does.

I haven't tested it by loading in a lot of footprints, but I could. I tried to write a test to check how much memory was being used by open vs. load dataset, but I couldn't get it to work (it's probably possible, but not straightforward because of how numpy stores things).

Issue #588, opened 2023-03-03
Add `transform_*` wrapper function akin to `standardise_*` functions currently used


rt17603:
### Is your feature request related to a problem?

The e.g. `*.read_file()` methods are currently accessed through `standardise_*` functions on the front end to improve workflow for users. We should probably also add equivalent `tranform_*` funcitons for `*.tranform_data()` methods as well. At the moment this has only been implemented for the `Emissions` object so we can at the least start there and build up as needed.

PR #760, opened 2023-08-24
Updated standardise module, applied to tests

brendan-m-murphy:
This PR refactors tests to use "standardise" functions instead of directly using BaseStore children.
This serves two purposes:
1) It means that we can change how data is standardised without modifying all the tests.
2) It shows points where BaseStore children are (currently) being used to standardise data, 
as opposed to being used to access the metastore.

Previously, `standardise_surface`, etc. could not accept a bucket as an argument, which meant that BaseStore children needed to be used in a context manager to standardise files. Now these functions do accept a bucket.

`standardise_surface`, `standardise_flux`, `standardise_footprints`, etc. now all call a generic `standardise` function that accepts a `data_type` argument.

I also added standardising function for binary data, remote data, and transformed data (these also accept a `data_type` argument). These functions are kind of ugly, but they're meant to help refactor the BaseStore children by providing an interface that hides the actual implementation details.

### DIFF
--- a/openghg/standardise/_standardise.py
+++ b/openghg/standardise/_standardise.py
@@ -1,18 +1,33 @@
 from pathlib import Path
-from typing import Dict, Literal, Optional, Union
+from typing import Dict, Literal, Optional, Union, cast
 from pandas import Timedelta
 
+from openghg.store.spec import define_data_type_classes
 from openghg.cloud import create_file_package, create_post_dict
 from openghg.objectstore import get_writable_bucket
 from openghg.util import running_on_hub
 from openghg.types import optionalPathType, multiPathType
 
 
+def standardise(
+    bucket: str, data_type: str, filepath: multiPathType, **kwargs
+) -> Optional[dict]:
+    try:
+        dclass = define_data_type_classes()[data_type]
+    except KeyError:
+        raise ValueError(f"No standarising function found for data type {data_type}.")
+    else:
+        with dclass(bucket) as dc:
+            result = dc.read_file(filepath=filepath, **kwargs)
+    return result
+
+
 def standardise_surface(
-    filepaths: multiPathType,
     source_format: str,
     network: str,
     site: str,
+    filepath: Optional[multiPathType] = None,
+    filepaths: Optional[multiPathType] = None,
     inlet: Optional[str] = None,
     height: Optional[str] = None,
     instrument: Optional[str] = None,
@@ -24,6 +39,7 @@ def standardise_surface(
     verify_site_code: bool = True,
     site_filepath: optionalPathType = None,
     store: Optional[str] = None,
+    bucket: Optional[str] = None,
 ) -> Optional[Dict]:
     """Standardise surface measurements and store the data in the object store.
 
@@ -52,14 +68,19 @@ def standardise_surface(
             Otherwise will use the data stored within openghg_defs/data/site_info JSON file by default.
         store: Name of object store to write to, required if user has access to more than one
         writable store
+        bucket: object store bucket to use; this takes precendence over 'store'
     Returns:
         dict: Dictionary of result data
     """
     from openghg.cloud import call_function
-    from openghg.store import ObsSurface
 
-    if not isinstance(filepaths, list):
-        filepaths = [filepaths]
+    if filepath is None and filepaths is None:
+        raise ValueError("One of `filepath` and `filepaths` must be specified.")
+    elif filepath is None:
+        filepath = filepaths
+
+    if not isinstance(filepath, list):
+        filepath = [filepath]
 
     if running_on_hub():
         # TODO: Use input for site_filepath here? How to include this?
@@ -84,7 +105,7 @@ def standardise_surface(
             metadata["sampling_period"] = sampling_period
 
         responses = {}
-        for fpath in filepaths:
+        for fpath in filepath:
             gcwerks = False
             if source_format.lower() in ("gc", "gcwerks"):
                 metadata["source_format"] = "gcwerks"
@@ -136,27 +157,27 @@ def standardise_surface(
 
         return responses
     else:
-        bucket = get_writable_bucket(name=store)
+        if bucket is None:
+            bucket = get_writable_bucket(name=store)
 
-        with ObsSurface(bucket=bucket) as obs:
-            results = obs.read_file(
-                filepath=filepaths,
-                source_format=source_format,
-                network=network,
-                site=site,
-                inlet=inlet,
-                height=height,
-                instrument=instrument,
-                sampling_period=sampling_period,
-                calibration_scale=calibration_scale,
-                measurement_type=measurement_type,
-                overwrite=overwrite,
-                verify_site_code=verify_site_code,
-                site_filepath=site_filepath,
-                update_mismatch=update_mismatch,
-            )
-
-        return results
+        return standardise(
+            bucket=bucket,
+            data_type="surface",
+            filepath=filepath,
+            source_format=source_format,
+            network=network,
+            site=site,
+            inlet=inlet,
+            height=height,
+            instrument=instrument,
+            sampling_period=sampling_period,
+            calibration_scale=calibration_scale,
+            measurement_type=measurement_type,
+            overwrite=overwrite,
+            verify_site_code=verify_site_code,
+            site_filepath=site_filepath,
+            update_mismatch=update_mismatch,
+        )
 
 
 def standardise_column(
@@ -172,6 +193,7 @@ def standardise_column(
     source_format: str = "openghg",
     overwrite: bool = False,
     store: Optional[str] = None,
+    bucket: Optional[str] = None,
 ) -> Optional[Dict]:
     """Read column observation file
 
@@ -195,11 +217,11 @@ def standardise_column(
         source_format : Type of data being input e.g. openghg (internal format)
         overwrite: Should this data overwrite currently stored data.
         store: Name of store to write to
+        bucket: object store bucket to use; this takes precendence over 'store'
     Returns:
         dict: Dictionary containing confirmation of standardisation process.
     """
     from openghg.cloud import call_function
-    from openghg.store import ObsColumn
 
     filepath = Path(filepath)
 
@@ -230,24 +252,24 @@ def standardise_column(
         response_content: Dict = fn_response["content"]
         return response_content
     else:
-        bucket = get_writable_bucket(name=store)
+        if bucket is None:
+            bucket = get_writable_bucket(name=store)
 
-        with ObsColumn(bucket=bucket) as obs_col:
-            result = obs_col.read_file(
-                filepath=filepath,
-                satellite=satellite,
-                domain=domain,
-                selection=selection,
-                site=site,
-                species=species,
-                network=network,
-                instrument=instrument,
-                platform=platform,
-                source_format=source_format,
-                overwrite=overwrite,
-            )
-
-        return result
+        return standardise(
+            bucket=bucket,
+            data_type="column",
+            filepath=filepath,
+            satellite=satellite,
+            domain=domain,
+            selection=selection,
+            site=site,
+            species=species,
+            network=network,
+            instrument=instrument,
+            platform=platform,
+            source_format=source_format,
+            overwrite=overwrite,
+        )
 
 
 def standardise_bc(
@@ -259,6 +281,7 @@ def standardise_bc(
     continuous: bool = True,
     overwrite: bool = False,
     store: Optional[str] = None,
+    bucket: Optional[str] = None,
 ) -> Optional[Dict]:
     """Standardise boundary condition data and store it in the object store.
 
@@ -273,11 +296,11 @@ def standardise_bc(
         continuous: Whether time stamps have to be continuous.
         overwrite: Should this data overwrite currently stored data.
         store: Name of store to write to
+        bucket: object store bucket to use; this takes precendence over 'store'
     returns:
         dict: Dictionary containing confirmation of standardisation process.
     """
     from openghg.cloud import call_function
-    from openghg.store import BoundaryConditions
 
     filepath = Path(filepath)
 
@@ -303,19 +326,19 @@ def standardise_bc(
         response_content: Dict = fn_response["content"]
         return response_content
     else:
-        bucket = get_writable_bucket(name=store)
-        with BoundaryConditions(bucket=bucket) as bcs:
-            result = bcs.read_file(
-                filepath=filepath,
-                species=species,
-                bc_input=bc_input,
-                domain=domain,
-                period=period,
-                continuous=continuous,
-                overwrite=overwrite,
-            )
-
-        return result
+        if bucket is None:
+            bucket = get_writable_bucket(name=store)
+        return standardise(
+            bucket=bucket,
+            data_type="boundary_conditions",
+            filepath=filepath,
+            species=species,
+            bc_input=bc_input,
+            domain=domain,
+            period=period,
+            continuous=continuous,
+            overwrite=overwrite,
+        )
 
 
 def standardise_footprint(
@@ -336,6 +359,7 @@ def standardise_footprint(
     high_time_resolution: bool = False,
     overwrite: bool = False,
     store: Optional[str] = None,
+    bucket: Optional[str] = None,
 ) -> Optional[Dict]:
     """Reads footprint data files and returns the UUIDs of the Datasources
     the processed data has been assigned to
@@ -359,12 +383,12 @@ def standardise_footprint(
                         Note this will be set to True automatically for Carbon Dioxide data.
         overwrite: Overwrite any currently stored data
         store: Name of store to write to
+        bucket: object store bucket to use; this takes precendence over 'store'
     Returns:
         dict / None: Dictionary containing confirmation of standardisation process. None
         if file already processed.
     """
     from openghg.cloud import call_function
-    from openghg.store import Footprints
 
     filepath = Path(filepath)
 
@@ -399,28 +423,28 @@ def standardise_footprint(
         response_content: Dict = fn_response["content"]
         return response_content
     else:
-        bucket = get_writable_bucket(name=store)
-        with Footprints(bucket=bucket) as fps:
-            result = fps.read_file(
-                filepath=filepath,
-                site=site,
-                domain=domain,
-                model=model,
-                inlet=inlet,
-                height=height,
-                metmodel=metmodel,
-                species=species,
-                network=network,
-                period=period,
-                chunks=chunks,
-                continuous=continuous,
-                retrieve_met=retrieve_met,
-                high_spatial_resolution=high_spatial_resolution,
-                high_time_resolution=high_time_resolution,
-                overwrite=overwrite,
-            )
-
-        return result
+        if bucket is None:
+            bucket = get_writable_bucket(name=store)
+        return standardise(
+            bucket=bucket,
+            data_type="footprints",
+            filepath=filepath,
+            site=site,
+            domain=domain,
+            model=model,
+            inlet=inlet,
+            height=height,
+            metmodel=metmodel,
+            species=species,
+            network=network,
+            period=period,
+            chunks=chunks,
+            continuous=continuous,
+            retrieve_met=retrieve_met,
+            high_spatial_resolution=high_spatial_resolution,
+            high_time_resolution=high_time_resolution,
+            overwrite=overwrite,
+        )
 
 
 def standardise_flux(
@@ -437,6 +461,7 @@ def standardise_flux(
     continuous: bool = True,
     overwrite: bool = False,
     store: Optional[str] = None,
+    bucket: Optional[str] = None,
 ) -> Optional[Dict]:
     """Process flux data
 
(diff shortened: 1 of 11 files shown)


openghg/standardise/_standardise.py
```diff
@@ -489,72 +508,48 @@ def standardise_flux(
...
+) -> Optional[dict]:
+    dclass = get_data_class(data_type)
+    with dclass(bucket) as dc:
+        result = dc.read_data(
+            binary_data=binary_data, metadata=metadata, file_metadata=file_metadata, **kwargs
+        )
+    return result
+
+
+def standardise_from_remote_source(bucket: str, data_type: str, data: dict, **kwargs: Any) -> Optional[dict]:
```
rt17603, line 542: Missing doc string here (and including that would make it easier to understand what this new function is doing)

openghg/standardise/_standardise.py
```diff
@@ -489,72 +508,48 @@ def standardise_flux(
...
+            high_time_resolution=high_time_resolution,
+            period=period,
+            continuous=continuous,
+            chunks=chunks,
+            overwrite=overwrite,
+        )
+
+
+def standardise_from_binary_data(
+    bucket: str, data_type: str, binary_data: bytes, metadata: dict, file_metadata: dict, **kwargs: Any
```
rt17603, line 532: Missing doc-string

openghg/standardise/_standardise.py
```diff
@@ -489,72 +508,48 @@ def standardise_flux(
...
+
+
+def standardise_from_remote_source(bucket: str, data_type: str, data: dict, **kwargs: Any) -> Optional[dict]:
+    dclass = get_data_class(data_type)
+    with dclass(bucket) as dc:
+        result = dc.store_data(data=data, **kwargs)
+    return result
+
+
+def standardise_via_transform(
```
rt17603, line 549: See Issue #588 in relation to the plans for the transform functionality. This is meant to be expanded and sit in parallel to standardise rather than going through the same interface.

tests/db_integrity/test_data_updates.py
```diff
@@ -261,14 +252,9 @@ def test_obs_data_read_header_diff():
      - BSD GCMD different data added - header changed so hash will be different but data will be the same
     Expect that GCMD (and CRDS) data can still be accessed.
     """
-    clear_test_stores()
-    # Load BSD data - CRDS data
     bsd_data_read_crds()
```
rt17603, line 255: Please add the clarifying comments be added back here.

tests/db_integrity/test_data_updates.py
```diff
@@ -511,10 +471,9 @@ def test_obs_data_representative_date_overlap():
 
     This test checks this will no longer raise a KeyError based on this.
     """
-
-    clear_test_stores()
-    bsd_data_read_crds_internal_overlap()
-    bsd_data_read_crds_internal_overlap(overwrite=True)
+    # Add same data twice, overwriting the second time
+    bsd_crds_overwrite(False)
```
rt17603, line 475: I feel the previous setup for this was more clear - this was a test for a specific bug we found rather than a generic overwrite of crds data.

tests/store/test_eulerian_model.py
```diff
@@ -1,16 +1,17 @@
...
 from xarray import open_dataset
 
 
 def test_read_file():
     test_datapath = get_eulerian_datapath("GEOSChem.SpeciesConc.20150101_0000z_reduced.nc4")
 
     bucket = get_bucket()
-    with EulerianModel(bucket=bucket) as em:
-        proc_results = em.read_file(filepath=test_datapath, model="GEOSChem", species="ch4")
+    proc_results = standardise(bucket=bucket, data_type="eulerian_model", filepath=test_datapath, model="GEOSChem", species="ch4")
```
rt17603, line 14: What we should do really is add a `standardise_eulerian` function and test this here. I'm adding this a suggestion though rather than saying this needs to be done in this PR (otherwise we should open an issue).

tests/store/test_footprints.py
```diff
@@ -121,34 +126,25 @@ def test_read_footprint_standard(keyword, value):
...
-        fps.read_file(
-            filepath=datapath,
-            site=site,
-            model=model,
-            network=network,
-            inlet=inlet,
-            domain=domain,
-            period="monthly",
-            high_spatial_resolution=True,
+    standardise(
```
rt17603, line 139: Small comment - any reason why this is laid out differently than the `test_read_footprint_standard`? I like how that test was laid out, so a small suggestion would be to match this to that to keep consistent.

tests/store/test_footprints.py
```diff
@@ -277,7 +273,7 @@ def test_read_footprint_high_spatial_resolution():
         ),
     ],
 )
-def test_read_footprint_co2(site, inlet, metmodel, start, end, filename):
+def test_read_footprint_co2(site, inlet, metmodel, start, end, filename, standardise):
```
rt17603, line 276: Same small comment as for `test_read_footprint_high_spatial_resolution` - would be nice to keep the variable definition and layout consistent.

tests/db_integrity/test_data_updates.py
```diff
@@ -6,108 +6,80 @@
...
+@pytest.fixture(autouse=True, scope="module")
+def auto_bucket():
+    """Put `bucket` in namespace.
+
+    Making bucket a module level variable doesn't work, and
+    autouse fixtures don't provide return values unless passed
+    as arguments to the test, so this seems like the best way
+    to do this.
+    """
+    global bucket
```
rt17603, line 33: Can I ask what problem this is aiming to solve? Is the idea to allow `bucket` to be accessible to all tests without having to specify it each time?

I think overall we should avoid global state where we can as it can easily cause confusion. Using a fixture with the functions directly and without the autouse may be a more clear approach here.

tests/db_integrity/test_data_updates.py
```diff
@@ -6,108 +6,80 @@
...
 from openghg.objectstore import get_bucket
-
-
 from helpers import clear_test_stores
 
 
-def flux_data_read():
-    """
-    Flux data set up.
+@pytest.fixture(autouse=True)
```
rt17603, line 14: I think we should be careful with overusing `autouse` as it can cause confusion - my feel is that for these tests in particular we'd be better off being more explicit with clearing the store at the start of each test to make it clear the steps that have been taken.

tests/db_integrity/test_data_updates.py
```diff
@@ -141,18 +112,17 @@ def bsd_small_edit_data_read():
...
+        filepath=(bsd_path3, bsd_prec_path3),
+        source_format=source_format2,
+        site=site,
+        network=network,
+        instrument=instrument,
+    )
 
 
-def bsd_diff_data_read(overwrite=False):
+def bsd_diff_data_read():
```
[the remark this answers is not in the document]
brendan-m-murphy: For the next 3-4 comments, in an earlier version I was trying to use fixtures to help set things up and it ended up being more complicated, so I tried to restore it... I'll try to put it back to closer to what it was.

tests/store/test_emissions.py
```diff
@@ -11,43 +13,67 @@
...
-        "high_time_resolution": False,
-    }
+    We only deal with emissions data in this test module, and most
+    tests just use `bucket = get_bucket()`, so this works for
+    most tests in this module.
+    """
+    return partial(standardise_flux, bucket=bucket)
+
+
+class TestCardamom:
```
rt17603, line 33: I understand why this approach has been taken to create test classes to group together functionality but if this needs to be split out in this way, this may be better split into two separate files instead, linking with the appropriate fixtures. If possible it would be preferable to keep this as a set of tests in one file, linking to fixtures if different data needs to be set up.

A few other points here:
 - Often the data is used not because it's EDGAR / Cardamom data but as a way to simulate a particular set of steps. So splitting by the two emissions types here doesn't necessarily map to the purpose of the tests.
 - As long as we understand what we're testing, having both EDGAR and Cardamom data within the same object store is not a wholly bad thing (as it could highlight some problems we hadn't anticipated) - however I do appreciate this is a less clear and predictable approach for the tests.

tests/store/test_footprints.py
```diff
@@ -1,10 +1,32 @@
...
+
+@pytest.fixture(autouse=True, scope="module")
+def auto_bucket():
+    global bucket
+    bucket = get_bucket()
+    yield bucket
+    globals().pop('bucket')
+
+
+@pytest.fixture(autouse=True)
```
rt17603, line 18: In this case, is the autouse producing anything here (based on your previous comments around the reason for global bucket)?

It looks like `standardise` is used within the tests as needed as an input?

tests/store/test_footprints.py
```diff
@@ -1,10 +1,32 @@
...
+    """Use `standardise(**kwargs)` to standardise footprints data.
+
+    We only deal with footprints data in this test module, and most
+    tests just use `bucket = get_bucket()`, so this works for
+    most tests in this module.
+
+    This was added to help refactor `BaseStore` and its children
+    by removing uses of the context manager for using `read_file`.`
+    """
+    return partial(standardise_footprint, bucket=bucket)
```
rt17603, line 29: Again, has this been created as a means to reduce repetition for creating the bucket?

CHANGELOG.md
```diff
@@ -9,6 +9,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [0.6.2] - 2023-08-07
```
rt17603, line 10: As 0.6.2 has been released, this should be included under the "Unreleased" section instead.

rt17603 (changes requested):
Great work - some very useful restructuring and clarity added to the code here with some nice features.

I have some questions about some of the new functions added and around some updates to the tests to be addressed.

tests/db_integrity/test_data_updates.py
brendan-m-murphy: It's the same as calling "bucket" as a fixture each time. I could make it function scope and then it would essentially be a local variable.

I guess it serves the same purpose as the fixture that clears test stores each time.

I could just call the fixture explicitly if you prefer that.

tests/store/test_footprints.py
brendan-m-murphy: Yes, I didn't figure out how to make an autouse fixture set a variable (e.g. bucket) until after I had updated these tests. And originally the `standardise` fixture did a lot more, because I didn't update the standardise_* functions at first, then later decided it made sense to do that.

tests/store/test_footprints.py
brendan-m-murphy: Partially because I could pass keyword=value through **kwargs, but not directly. I changed it so that only keyword=value is in kwargs.

tests/store/test_eulerian_model.py
brendan-m-murphy: I've now added this function

tests/db_integrity/test_data_updates.py
```diff
@@ -164,20 +134,42 @@ def bsd_diff_data_read(overwrite=False):
...
+        filepath=(bsd_path5, bsd_prec_path5),
+        source_format=source_format2,
+        site=site,
+        network=network,
+        instrument=instrument,
+        overwrite=True,
+    )
+
+
+def bsd_diff_date_range_read_delayed():
```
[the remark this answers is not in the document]
brendan-m-murphy: I removed this... it's a duplicate of one of the original helper functions.

tests/db_integrity/test_data_updates.py
brendan-m-murphy: I put the tests with overwrite parameters back as they were

tests/db_integrity/test_data_updates.py
brendan-m-murphy: Done

openghg/standardise/_standardise.py
brendan-m-murphy: I added a _transform.py module to the top level of openghg.transform and put `transform_emissions_data` in there, which does the same thing that this function did. (I added a doc string as well.)

openghg/standardise/_standardise.py
brendan-m-murphy: fixed

openghg/standardise/_standardise.py
brendan-m-murphy: I removed this function. It was meant to be a front end for `store_data`, but I've just changed that back to using the `ObsSurface` context manager. (Most likely I will remove that context manager when I made the Metastore class, so I'll have to change the call slightly then, but I guess that is better than having an out of place function in standardise.)

tests/store/test_emissions.py
brendan-m-murphy: I think I did this because I figured the groups made sense. I'll just drop the classes if the groups don't make sense. (Some book I have suggested using classes for groups of related tests. They said they don't typically use them for grouping fixtures, but it seemed to make sense in this case.)

tests/store/test_footprints.py
brendan-m-murphy: I think you're right, I removed the "auto_bucket" fixture and standardise fixture and just called bucket where needed.

tests/store/test_footprints.py
rt17603: This isn't a strong suggestion so if you feel like there's a good reason to lay it out how you did that's fine.

brendan-m-murphy:
I think I addressed everything you brought up. Some of the test files were a bit of a mess, I think they're better now (no autouse fixtures, and I replaced "standardise" fixtures with the proper functions from the standardise module). I added a transform module to mirror the standardise module, added docstrings, and got rid of one function in standardise that didn't fit that well.

openghg/transform/_transform.py
rt17603, line 1: In general, adding this sort of parallel functionality should probably be included on a separate branch and PR (and will close the #588 issue mentioned previously).

As it is, if we're keeping this here, we should also update the associated tutorial to now use this new function - https://github.com/openghg/openghg/blob/devel/doc/source/tutorials/local/Adding_new_data/Creating_data_from_databases.rst#1-creating-emissions-from-the-edgar-database

tests/db_integrity/test_data_updates.py
```diff
@@ -511,10 +462,9 @@ def test_obs_data_representative_date_overlap():
 
     This test checks this will no longer raise a KeyError based on this.
     """
-
-    clear_test_stores()
+    # Add same data twice, overwriting the second time
     bsd_data_read_crds_internal_overlap()
-    bsd_data_read_crds_internal_overlap(overwrite=True)
+    bsd_data_read_crds_internal_overlap(overwrite=False)
```
rt17603, line 467: This should include `overwrite=True`

tests/store/test_emissions.py
```diff
@@ -108,10 +123,33 @@ def test_read_file():
     assert metadata.items() >= expected_metadata.items()
 
 
-# TODO: Add test for adding additional years data - 2013 gpp cardomom
+    # TODO: Add test for adding additional years data - 2013 gpp cardomom
 
+@pytest.fixture
+def load_edgar(bucket):
+    def _load_edgar(version, year, database_version: bool = True):
```
rt17603, line 130: There's a few assumptions here we can't necessarily make around EDGAR data
 - For the database version there was previously a v4.3.2 and there may be other versions going forward which use similiar syntax
 - We may want to look at monthly edgar data rather than yearly (that's a case we will need to add in future)
 - We may also want to look at a different species e.g. co2 or no2

Overall, I think we'd be better off just allowing this be flexibly called rather than setting this up as a more clever fixture.

rt17603: And think this isn't an update which is needed in this PR

tests/db_integrity/test_data_updates.py
rt17603: Yes I think if we can avoid global state where we can, that would be better.

tests/db_integrity/test_data_updates.py
rt17603: @brendan-m-murphy - just a lingering comment from previous review on the use of autouse for clear_stores in test_data_updates.py to address.

rt17603 (changes requested):
Thanks for making the changes - most look good, just a lingering comment on autouse to address and a few more comments based on the updates.

You may also consider splitting out some of the added functionality into a different PR (e.g. adding the transform functionality; updating the way edgar is accessed in the test files) but we can leave this for now.

tests/store/test_emissions.py
brendan-m-murphy: This fixture is really only meant for the tests in that file. I added it because otherwise I would have had to make the standardize changes in 10 places instead of 1. I can substitute the body of the fixture back into the tests, but I still need to update those tests one way or another.

rt17603: Sure, I'm thinking about building in flexibility which will be needed in this test file (other versions / gases for edgar will likely be included in tests within this file).

A few small additions which may address this:
 - make the full string for the version be included in the input (solves the "v4.3.2" issue)
 - add species as an input (e.g. "ch4")

brendan-m-murphy:
@rt17603 I made the changes suggested in the second code review. I didn't split off the `transform` module or EDGAR test changes into separate PRs; I can do this if you like.

tests/db_integrity/test_data_updates.py
```diff
@@ -2,112 +2,74 @@
...
-
-
 from helpers import clear_test_stores
 
 
-def flux_data_read():
-    """
-    Flux data set up.
+@pytest.fixture
+def clear_stores():
```
rt17603, line 15: One final detail around this - at this point these fixtures have just ended up as function calls. Re-evaluating these changes in this state, it might make more sense to either:
1. Just add the function calls themselves back to the tests rather than using these fixtures
2. Combine these two fixtures to get and then clear the bucket (could then return bucket value to be used within the tests)

rt17603 (changes requested):
Thanks for addressing comments so far on this - looks good. One last tidy up step for test_data_updates.py to be addressed based on its current state - see comment in file for suggestions here.

tests/db_integrity/test_data_updates.py
brendan-m-murphy: All of the tests clear the stores, so they all need to call `clear_stores`, which is why it was autouse, but not all of them need `bucket` as an argument now, because Gareth suggested that I change the standardise functions to only accept 'store' as an argument, and not 'bucket'.

Gareth is also suggesting that `get_bucket()` be replaced by `get_writeable_buckets(name="user")`, so if I add the function calls back then that change becomes harder to make.

Changing the helper functions to take 'bucket' as an argument instead of inheriting through the `auto_bucket' fixture made making the changes Gareth suggested about 'store' instead of 'bucket' harder to do, so I feel like removing some of these fixtures has already made making changes harder. (I admit the global argument in auto_bucket seemed slightly dodgy, although I don't think anything bad could happen because globals are defined at the point where functions are defined, so it only affects code within the same module, which consists of tests and helper functions.)

So, I don't think either option is great, given the changes suggested by Gareth.

rt17603: The first option would be the best one to go with then. There's no reason to have `clear_stores` as a fixture if all it contains is one function call and there's no issue really with having that as a function at the start each test.

Clarity is much more important than brevity when it comes to tests (especially this set of tests). The problem with the autouse was that this reduced clarity especially when the purpose of these specific tests was to very clearly check a particular state of the object store.

Overall this could all be improved by creating a "give me an empty bucket" fixture which creates a new empty bucket and then tidies this up afterwards (which is what I was partially suggesting the fixture at the top could be turned into) but we can use what we have to work with for now.

rt17603:
> @rt17603 Gareth requested some changes on PR #767, but he meant them to go with this PR, so I've made those changes. Those changes affect the code in your final requested changed, and I don't think either option is very good now.

When asking for review I would always recommend asking at the point where you're satisfied the PR is completed and ready to be looked at. This includes the "Files changed" tab showing the changes specifically related to that PR - that way it's much easier to review.

For chained PRs asking for a review on both means we often end up with changes on both branches and more chance of merge conflicts (and a more complex history). So yes I would recommend requesting one review after the other rather than both at the same time.

tests/db_integrity/test_data_updates.py
brendan-m-murphy: Okay I will put the code back into the tests and remove the fixtures. The combined fixture wouldn't do any harm, but not all of the tests need to use a bucket, so I don't think that aids clarity.

I personally think the autouse fixture for clearing stores is more explicit in this case, since we want every test to start from a blank state; if the code is moved to the tests, then it's not obvious that all of these tests start from a blank slate (in some test modules, clearing stores isn't done uniformly).

Removing the bucket fixture isn't that big of a deal since I think the change to `get_writeable_buckets` should just be search and replace.

rt17603: Thanks for applying that change. I do understand what you're saying - you were aiming to get everything set up from a known base state of an empty object store (and I do think one resolution to consider for future would be to have a `get_empty_bucket` type input).

However, overall we should be careful with what is essentially hidden state and autouse functions and global variables can act as that - I've definitely encountered problems with that kind of setup in the past when trying to debug tests and setting up mocking for instance.

rt17603 (approved):
Looks good - happy for this to be merged.

PR #763, opened 2023-08-24
Enable read-only opening of metastore

### LINKED ISSUE #762: Stop writing out the metadata store on search
We shouldn't be writing the metastore on search. Currently each metastore is being loaded and then written out by the context manager when it goes out of scope.

gareth-j:
* **Summary of changes**

This enables the read-only opening of the metastore by passing the a `mode` keyword to `load_metastore` and the constructor of the store.

- [x] Closes #762 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Bug where the object store path being written to JSON led to an invalid path being given to some users - [PR #741](https://github.com/openghg/openghg/pull/741)
 
+### Changes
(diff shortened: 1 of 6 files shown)


rt17603 (commented):
Looks good overall - just a question about whether it would be worth raising an exception of some kind rather than printing and returning None. Once that question is addressed, happy for this to be merged (just ask for re-review).

brendan-m-murphy (changes requested):
Looks like a nice solution. I have the same comment as Rachel: maybe writing to a read only copy of the metastore.

brendan-m-murphy (approved):
Looks good to me

PR #775, opened 2023-09-12
Idea for bool -> str for current data [skip ci]

gareth-j:

Just an idea on how to use a `str` instead of an optional bool. To me it makes it clearer but I'm not sure everyone will agree.

### DIFF
--- a/openghg/util/_versions.py
+++ b/openghg/util/_versions.py
@@ -13,7 +13,7 @@ import platform
 import struct
 import subprocess
 import sys
-from typing import List, IO, Union, Tuple, Optional, cast
+from typing import List, IO, Union, Tuple
 
 __all__ = ["show_versions", "check_if_need_new_version"]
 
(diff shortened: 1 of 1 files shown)


gareth-j:
Yeah it might not be the best name, I'm just playing with ideas on how to move away from the `Optional[bool]` to a string.

rt17603:
> Yeah it might not be the best name, I'm just playing with ideas on how to move away from the `Optional[bool]` to a string.

I don't think having an Optional[bool] is actually a bad thing - for an Optional[str] I agree this could offer too many options but for an optional bool it's clear what this is doing as long as it's well documented and only provides three clear options.

rt17603:
@gareth-j - Can I close this given this has been changed in #811 ?

rt17603:
@gareth-j - Can I close this now due to the changes made in  in https://github.com/openghg/openghg/pull/811 ?

PR #787, opened 2023-09-14
Iss786 model scenario flux key

### LINKED ISSUE #786: `ModelScenario.calc_modelled_obs` throws an error if 'source' or 'sources' isn't specified
### What happened?

I created a model scenario with obs, footprint, flux, and bc data: 

`scenario = ModelScenario(site='tac', species='ch4', inlet='185m', domain='europe', start_date='2019-01-01')`

The model scenario was successfully created with all four data types:

```
INFO:openghg.analyse:Adding obs_surface to model scenario
INFO:openghg.analyse:Updating any inputs based on observation data
INFO:openghg.analyse:site: tac, species: ch4, inlet: 185m
INFO:openghg.analyse:Using height_name option(s) for footprint inlet: 185magl
WARNING:openghg.retrieve:No data found for the dates given in the user store, please try a wider search.
INFO:openghg.analyse:Adding footprint to model scenario
WARNING:openghg.retrieve:No data found for the dates given in the user store, please try a wider search.
INFO:openghg.analyse:Adding flux to model scenario
INFO:openghg.analyse:Adding boundary_conditions to model scenario
```

The attributes `scenario.obs, scenario.footprint, scenario.bc, scenario.fluxes` had the expected data.

However, `scenario.fluxes` used `None` as the key for the flux, since I didn't specify `source` or `sources` in `__init__`:

```
In [78]: scenario.fluxes
Out[78]: 
{None: FluxData(data=<xarray.Dataset>
 Dimensions:  (lat: 293, lon: 391, time: 1)
 Coordinates:
   * lat      (lat) float32 10.73 10.96 11.2 11.43 ... 78.36 78.59 78.82 79.06
   * lon      (lon) float32 -97.9 -97.55 -97.2 -96.84 ... 38.32 38.68 39.03 39.38
   * time     (time) datetime64[ns] 2019-01-01
   ...
```

This caused an error when I called `scenario.footprints_data_merge()`:

```
ModelScenario.combine_flux_sources(self, sources, cache, recalculate)
    841 time_dim = "time"
    843 sources = self._clean_sources_input(sources)
--> 844 sources_str = ", ".join(sources)
    846 # Return any matching cached data
    847 if self.flux_stacked is not None and not recalculate:

TypeError: sequence item 0: expected str instance, NoneType found
```

I expected to get combined data back from the function call.

_No response_

### Relevant log output

_No response_

The issue is on lines 817-818 of _scenario.py:

```
        if sources is None:
            sources = list(flux_dict.keys())
```

The variable `sources` is passed to `combine_flux_sources`, which tries to do `", ".join(sources)`.

### Environment

<details>
Python version 3.9
OpenGHG version 0.6.2

</details>

brendan-m-murphy:

If 'source' or 'sources' is not specified, but `ModelScenario` only finds one flux file matching the given arguments,
then `ModelScenario.add_flux` tries to infer the source, and if not found, it sets the key for the added flux to 'source_not_specified'.

A test was added to check this.

- [ ] Closes #786 (Replace xxxx with the Github issue number)

### DIFF
--- a/openghg/analyse/_scenario.py
+++ b/openghg/analyse/_scenario.py
@@ -464,6 +464,8 @@ class ModelScenario:
                 flux_source = self._get_data(flux_keywords, data_type="flux")
                 # TODO: May need to update this check if flux_source is empty FluxData() object
                 if flux_source is not None:
+                    if name is None and len(sources) == 1:
+                        name = flux_source.metadata.get("source", "no_source_specified")
                     flux[name] = flux_source
 
         elif flux is not None:
--- a/tests/analyse/test_scenario.py
+++ b/tests/analyse/test_scenario.py
@@ -1451,6 +1451,23 @@ def test_scenario_infer_flux_source_ch4():
     if only a single flux matches the given metadata
     and source is in the flux metadata.
     """
+    import tinydb
+
+    from openghg.store import data_manager
+    from openghg.objectstore import get_readable_buckets
+    from openghg.store import Emissions
+
+    # remove 'waste' flux file
+    bucket = get_readable_buckets()["user"]
+    with Emissions(bucket=bucket) as em:
+        query = tinydb.Query().fragment({'species': 'ch4', 'domain': 'europe', 'source': 'waste'})
+        result = em._metastore.search(query)[0]
+        uuid = result['uuid']
+
+    manager = data_manager(data_type="emissions", store="user")
+    manager.delete_datasource(uuid=uuid)
+
+    # create model scenario without specifying source
     start_date = "2012-08-01"
     end_date = "2012-09-01"
 
@@ -1460,7 +1477,6 @@ def test_scenario_infer_flux_source_ch4():
     network = "DECC"
 
     species = "ch4"
-    source = "anthro"
 
     bc_input = "MOZART"
 
@@ -1475,7 +1491,8 @@ def test_scenario_infer_flux_source_ch4():
         end_date=end_date,
     )
 
-    assert source in model_scenario.fluxes
+    # expect 'anthro' to be found in flux metadata:
+    assert 'anthro' in model_scenario.fluxes
 
 
 def test_modelscenario_doesnt_error_empty_objectstore():
--- a/tests/analyse/test_scenario.py
+++ b/tests/analyse/test_scenario.py
@@ -1445,6 +1445,39 @@ def test_stack_datasets_with_alignment(flux_daily, flux_daily_small_dim_diff):
     np.testing.assert_allclose(output_flux, expected_flux)
 
 
+def test_scenario_infer_flux_source_ch4():
+    """
+    Test ModelScenario can find the source of a flux
+    if only a single flux matches the given metadata
+    and source is in the flux metadata.
+    """
+    start_date = "2012-08-01"
+    end_date = "2012-09-01"
+
+    site = "tac"
+    domain = "EUROPE"
+    inlet = "100m"
+    network = "DECC"
+
+    species = "ch4"
+    source = "anthro"
+
+    bc_input = "MOZART"
+
+    model_scenario = ModelScenario(
+        site=site,
+        species=species,
+        inlet=inlet,
+        network=network,
+        domain=domain,
+        bc_input=bc_input,
+        start_date=start_date,
+        end_date=end_date,
+    )
+
+    assert source in model_scenario.fluxes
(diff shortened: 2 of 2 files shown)


gareth-j (dismissed):
Looks good!

openghg/analyse/_scenario.py
```diff
@@ -464,6 +464,8 @@ def add_flux(
                 flux_source = self._get_data(flux_keywords, data_type="flux")
                 # TODO: May need to update this check if flux_source is empty FluxData() object
                 if flux_source is not None:
+                    if name is None and len(sources) == 1:
+                        name = flux_source.metadata.get("source", "no_source_specified")
```
rt17603, line 468: As discussed on the Issue, if we're unable to extract a "source" value we should be raising an appropriate error as all flux data should have an associated source value and if it doesn't there's a separate issue.

tests/analyse/test_scenario.py
```diff
@@ -1445,6 +1445,56 @@ def test_stack_datasets_with_alignment(flux_daily, flux_daily_small_dim_diff):
     np.testing.assert_allclose(output_flux, expected_flux)
 
 
+def test_scenario_infer_flux_source_ch4():
```
rt17603, line 1448: To check - was this removing "waste" so this would be the only ch4 emissions file found and so could infer "anthro" as the source? If so, a nicer test would be to add a single flux file for a different long-lived gas (can create dummy flux data if needed here) and checking this rather than altering what's in the object store as this point.

rt17603 (changes requested):
Good to make sure this corner case is covered. There were a few things discussed on Issue #786 that need to be implemented here.

tests/analyse/test_scenario.py
brendan-m-murphy: Yes that's the reason for removing 'waste'.

The dummy fixtures used elsewhere in `test_scenario` wouldn't work since the change I made only applies when the `ModelScenario` needs to retrieve the flux; if I mock `get_flux`, I would need to validate the input somehow, since this should test a set of arguments that can actually be used to construct a model scenario.

If I change the species, I need to add matching boundary conditions, since `domain` is required for flux, and if `domain` is not `None`, then `ModelScenario.__init__` will try to fetch boundary conditions.


I decided that deleting a source would be fine since the last test clears the store anyway. (Admittedly, the last test tripped me up because I tried to add a new test after it and didn't realise that the store had been cleared at that point.)

openghg/analyse/_scenario.py
brendan-m-murphy: I added error handling now.

tests/analyse/test_scenario.py
rt17603: So you wouldn't need to add matching boundary conditions as you don't need to add boundary conditions to a ModelScenario instance at all. It can try and fail to find boundary conditions and that should be fine (if it's not that's a different problem).

rt17603: In principle, you could also just create an empty `ModelScenario` instance and test the `ModelScenario.add_flux` method directly since this is the function that has been updated.

brendan-m-murphy: Okay thanks, I'll try those ideas.

rt17603: Did you have a look into the other ideas for this here? This test does seem to check what you want it to check but I'm just considering if this could have unintended consequences for other tests later on if there's an expectation the "waste" emissions are there and they're not.

rt17603 (commented):
There was just an open question on this around the test added and whether there was another way to approach this? The test does seem to check what you want it to check but I'm just considering if this could have unintended consequences for other tests later on if there's an expectation the "waste" emissions are there and they're not.

brendan-m-murphy:
> There was just an open question on this around the test added and whether there was another way to approach this? The test does seem to check what you want it to check but I'm just considering if this could have unintended consequences for other tests later on if there's an expectation the "waste" emissions are there and they're not.

Yea I will update the tests quickly. I figured it was harmless to remove the waste emissions at this point because the last test clears the store, but actually the last test tripped me up when I tried to add a new test... so maybe it's better to use one of the ideas above.

rt17603 (approved):
Looks good to me - very small comment which could be updated - but otherwise looks good to go.

rt17603 (approved):

PR #800, opened 2023-10-02
configuration_file to specify tinydb or mongodb

### LINKED ISSUE #757: Configuration file for the object store
We could do with a configuration file for the object store. This will allow users to setup the object store and add in settings and paths to MongoDB etc. This is linked to #756.

SutarPrasad:

    Created a configuration file that specifies the version and mentions 2 database types `tinyBD` and `mongoDB`.
    The connection string for MongoDB or config file path for tiny DB configuration is not added as they are unknown.

- [x] Closes #757 (Replace xxxx with the Github issue number)
- [ ] Reading and writing functions added

### DIFF
--- a/openghg/db_config.toml
+++ b/openghg/db_config.toml
@@ -1,0 +1,6 @@
+config_version = ""
+[database]
+  type = ""
+
+  [database.mongodb]
+    uri = "mongodb connection string"
--- a/openghg/db_config.yaml
+++ b/openghg/db_config.yaml
@@ -1,8 +1,0 @@
-config_version: ""
-database:
-  type: "" #tinydb or "mongodb"
-  tinydb:
-    file_path: "tinydb config file"
-  mongodb:
-    uri: "mongodb connection string"
-  
--- a/openghg/db_config.yaml
+++ b/openghg/db_config.yaml
@@ -1,0 +1,8 @@
+config_version: ""
+database:
+  type: "" #tinydb or "mongodb"
+  tinydb:
+    file_path: "tinydb config file"
+  mongodb:
+    uri: "mongodb connection string"
+  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/openghg/openghg/compare/0.6.2...HEAD)
 
 ### Added
+- Added database configuration file to specify database type as tinydb or mongodb. - [PR #800](https://github.com/openghg/openghg/pull/800)
+
 - Added `DeprecationWarning` to the functions `parse_cranfield` and  `parse_btt`. - [PR #792](https://github.com/openghg/openghg/pull/792)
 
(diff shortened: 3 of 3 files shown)


brendan-m-murphy:
The format of the config file looks like it could work, but we need code to generate and use it. @gareth-j is this part of another issue?

openghg/db_config.yaml
```diff
@@ -0,0 +1,8 @@
+config_version: ""
```
gareth-j, line 1: I think this should be a TOML file instead of YAML as it's cleaner.

openghg/db_config.yaml
```diff
@@ -0,0 +1,8 @@
...
+database:
+  type: "" #tinydb or "mongodb"
+  tinydb:
+    file_path: "tinydb config file"
```
gareth-j, line 5: We don't need the filepath for the tinydb database as this will be in the object store. I think we'll only need config information for mongodb.

gareth-j (changes requested):
Should change to TOML and remove the tinydb section.

SutarPrasad:
> > The format of the config file looks like it could work, but we need code to generate and use it. @gareth-j is this part of another issue?
> 
> No, I was just working on the zarr stuff and realised this. @SutarPrasad we also need functions to
> 
>     * read the config file
> 
>     * check correctness of the config file
> 
>     * write to the config file (but I don't think we necessarily need a command line interface as with the user config setup)
> 
>     * documentation on how to create the file, we'll need a new section in the docs for this

Created new issue for this https://github.com/openghg/openghg/issues/814

gareth-j:
> > > The format of the config file looks like it could work, but we need code to generate and use it. @gareth-j is this part of another issue?
> > 
> > 
> > No, I was just working on the zarr stuff and realised this. @SutarPrasad we also need functions to
> > ```
> > * read the config file
> > 
> > * check correctness of the config file
> > 
> > * write to the config file (but I don't think we necessarily need a command line interface as with the user config setup)
> > 
> > * documentation on how to create the file, we'll need a new section in the docs for this
> > ```
> 
> Created new issue for this #814

I think we should keep all the read/write functionality as part of this PR. We don't want to add a TOML file to the openghg folder, it should be created in the root of the object store itself as it will be user independent.

So what we need is:

1. A function to create an empty config file in the root of an object store, where the object store path is given by the user
   1a. Optionally an interactive setup like we've got for `openghg --quickstart`, could be something like `openghg --setup-store` and have a few options for user input
3. A function to read the config file from the object store
4. Documentation added to the README and getting started documentation for setting the configuration file
5. Tests added for the reading and writing functions

openghg/db_config.toml
gareth-j, line 1: The format looks good but this file should be created by a function, not added to the `openghg` folder.

gareth-j (changes requested):
This needs a bit more work before it's ready.

gareth-j:
I think we can bump this from 0.7.0 as it's not required until we implement the mongodb code. Thoughts @rt17603  and @brendan-m-murphy ?

gareth-j:
I think this needs completely revisiting when MongoDB is being worked on. To tidy things up I'll close this.

PR #802, opened 2023-10-03
Merge `devel` into `new_object_store` - fixes needed

gareth-j:

So this caused me more issues than I wanted it to. I've had to move back to `load_dataset` instead of `open_dataset` because errors that come from xarray's internal caching structure.

The cause of the error seems to be trying to write to netcdfs that xarray has cached within its internal lru_cache and haven't been removed from the cache. This is probably due to a dangling file handler as `compute` isn't called on the opened files. If you do `xr.open_dataset(filepath).compute()` then the problem goes away. But this is just the same as doing `xr.load_dataset`.

## TODO

- there are 3 tests failing - @rt17603 can you have a look at these to see what's happened?
- the load vs open dataset thing may be fixed by the move to zarr, so maybe we can wait for that work to be merged in?


brendan-m-murphy:
@gareth-j not sure if you've already seen this: https://github.com/pydata/xarray/issues/4240

The last comment is:
> Running `xarray.backends.file_manager.FILE_CACHE.clear()` fixed the issue for me. I couldn't find any other way to stop xarray from pulling up some old data from a newly saved file. I'm using the h5netcdf engine with xarray version 2022.6.0 by the way.

Maybe try/except for saving to catch the permission denied error and clearing the cache if you get that? I've noticed a permission denied error when trying to make/modify test data in ipython.

gareth-j:
> @gareth-j not sure if you've already seen this: [pydata/xarray#4240](https://github.com/pydata/xarray/issues/4240)
> 
> The last comment is:
> 
> > Running `xarray.backends.file_manager.FILE_CACHE.clear()` fixed the issue for me. I couldn't find any other way to stop xarray from pulling up some old data from a newly saved file. I'm using the h5netcdf engine with xarray version 2022.6.0 by the way.
> 
> Maybe try/except for saving to catch the permission denied error and clearing the cache if you get that? I've noticed a permission denied error when trying to make/modify test data in ipython.

Ah I was thinking there must be a way of clearing that cache, good find! I tried calling `.close()` on the Datasets but that didn't seem to help but I may not have been catching them all or something.

tests/store/test_datasource.py
```diff
@@ -126,6 +126,8 @@ def test_versioning(capfd, bucket):
 
     min_keys = d.versions()
 
+    raise ValueError("Should this return data keys instead of datasource keys?")
```
rt17603, line 129: @gareth-j - An explicit raised `ValueError` is the cause of test errors - any idea why this was included and how this got into the code?

tests/store/test_datasource.py
```diff
@@ -536,6 +561,7 @@ def test_key_date_compare():
     with pytest.raises(ValueError):
         in_date = d.key_date_compare(keys=error_key, start_date=start, end_date=end)
 
+
```
rt17603, line 564: @gareth-j Can't add a detail on the specific line L204 as this was not changed but the test is checking for a folder called "datasource/uuid/..." but this now seems to have changed to "data/uuid/..." - is this expected behaviour for the changes made?

rt17603: Same issue for `test_in_daterange()` failing test on L493 - expected "datasource" not "data" as part of the uuid (filepath)

rt17603 (commented):
I've highlighted the issues as comments but for the 3 failing tests the issue seems to be around a change in the uuid from including "datasource/uuid" to "data/uuid" - is this expected behaviour? This needs to be resolved before this can be fixed.

Details:
 - `test_versioning` contains an explicit `raise ValueError` with a question "Should this return data keys instead of datasource keys?". The code below includes the "datasource/uuid" output.
 - `test_in_daterange` and `test_replace_version` explictly fail because of the new output as "data/uuid" rather than "datasource/uuid"

tests/store/test_datasource.py
gareth-j: Yeah I wasn't sure about those changes, I'll take a look at the git blame and see what happened.

rt17603: So this may be related to the changes in the way the Datasource was written out within the save() method? This used to do something internal but now primarily uses `openghg.objectstore` instead.

e.g. [Datasource.save() on devel](https://github.com/openghg/openghg/blob/devel/openghg/store/base/_datasource.py#L546) versus [Datasource.save() on new_object_store](https://github.com/openghg/openghg/blob/new_object_store/openghg/store/base/_datasource.py#L625)

gareth-j:
@rt17603 have you taken this over or are you waiting for me to make the changes?

rt17603:
> @rt17603 have you taken this over or are you waiting for me to make the changes?

I was leaving this for you as this seemed to be something introduced from using the underlying `openghg.objectstore` functionality. If that's the case, I think either `openghg.objectstore` needs to be updated to bring this in line with what is currently used or we use this an opportunity to re-define this since we're creating a new object store anyway.

rt17603:
@gareth-j Are we at the point where this is ready to be merged into `new_object_store`? Or do the changes still need to be added? I think it would be best to get this sorted and merged etc. so it's not lingering too long (especially if the work you're doing is branched off this etc.)

gareth-j:
OK this is ready to go in.

rt17603 (approved):
Happy for this to be merged as long as all tests pass locally.

Issue #541, opened 2023-01-25
Remove "units" from metadata.


rt17603:
Remove "units" from metadata. Currently, defined inconsistently (e.g., "nmol mol-1" in ICOS, nothing for CRDS, and "ppb" for GCWERKS). Use unit in data frame instead

_Originally posted by @mrghg in https://github.com/openghg/openghg/issues/395#issuecomment-1234263857_

rt17603:
Links to #540

rt17603:
Starting looking at this on branch: "rm_units_metadata"
 - Removed this from GCWERKS and updated impacted tests
 - Removed from global attributes and updated impacted tests
 - Need to check over other source_format options and do the same.

rt17603:
Also need to add this to icos and ceda retrieval functions, making sure the unit details are converted and integrated into the unit attributes for each species variable.

PR #806, opened 2023-10-11
Adding new filter for ObsPack data retrieval from ICOS Carbon Portal

### LINKED ISSUE #788: Retrieving European Obspack data from ICOS CP - specifying files
### What is your issue?

There have been a couple of changes on the ICOS CP that have broken the capability of openghg.retrieve.icos.retrieve_atmospheric to correctly import European Obspack data:

1) The .nc files were previously skipped through the fix in #630 . This searches dobj.meta["references"]["doi"]["titles"][0]["title"].lower() for "globalview" and skips when found, however the new .nc files no longer have "globalview" in this attribute. So a new solution is required to skip them.

2) The 2022 European Obspack data are still returned by the search, even though these data are now deprecated and superseded by the 2023 versions. So we need to be able to filter by version.

For example - try running:
cp_data = retrieve_atmospheric(site="TAC",
                               force_retrieval=True,
                               data_level=2,
                               dataset_source="European ObsPack",
                               update_mismatch="from_source",
                               species="co2",
                               inlet="54m")

This tries to retrieve three records (and fails on the third of these):
https://meta.icos-cp.eu/objects/MWN6tAJF49pvqY3B09KBkhJa
https://meta.icos-cp.eu/objects/1etvTvRIvxOIQ9Wut4RpRFUE
https://meta.icos-cp.eu/objects/MtJYBO6tu3qHn62u2oSIDIEF

We only want it to return the second of these records (although it might be useful to retain the ability to retrieve the first of these records too)

gareth-j:

Quick fix for ICOS data retrieval.

> [!NOTE]
> Tests aren't passing at the moment and I haven't got time to fix them at the moment, will keep this as a draft for now

- [x] Closes #788


tests/retrieve/icos/test_retrieve_icos.py
```diff
@@ -148,17 +148,37 @@ def test_icos_retrieve_skips_obspack_globalview(mocker, caplog):
 
     # Check attributes within stored Dataset contain extra keys
     attr1_expected_additional = {
-        "measurement_type": "co2 mixing ratio (dry mole fraction)",
+        "species": "co2",
+        "instrument": "FTIR",
+        "instrument_data": ["FTIR", "http://meta.icos-cp.eu/resources/instruments/ATC_505"],
+        "site": "WAO",
+        "measurement_type": "co2 (dry air mole fraction)",
+        "units": "µmol mol-1",
```
rt17603, line 156: We're planning to remove "units" from the global attributes at some point in relation to #541 (see also branch rm_units_metadata). We do also need to make sure the units details are converted and stored against the individual species.

Should we remove this in this PR? Or did we want to include it PR but to reference this in the linked issue so we make sure to cover this when fixing it? Thoughts?

rt17603 (commented):
Solution looks reasonable to me but I've also just asked Joe in case this may have more wide reaching impacts than we're anticipating. Also added a small comment on whether to include "units" as a global attribute.

tests/retrieve/icos/test_retrieve_icos.py
gareth-j: Yeah I can remove that now.

gareth-j: OK let's leave the units attribute as is for now.

joe-pitt (commented):
This is working nicely for me at the moment! I guess there's always a risk going forward that the labels associated with the files we want to exclude change again, but we can deal with that if/when it happens...

rt17603 (approved):
Ok, looks like a useful fix for now. A few small requests for the PR description:
 - Can this be updated to remove the note that the tests aren't passing (I'm assuming they now are?)
 - Can this be updated to summarise change made e.g. updating ICOS retrieve function to ignore all names containing "ObsPack".

PR #818, opened 2023-11-07
Dropping support for Python 3.8 and adding Python 3.12

### LINKED ISSUE #817: Drop support for Python 3.8
### What is your issue?

The last version of xarray that supports Python 3.8 was released in January 2023 (version 2023.1.0). In order to unpin pandas (and use versions >= 2.0) we need a slightly more recent version of xarray (>= version 2023.3).

The reason for this is that `xr.Dataset.resample` passes `base` as a kwarg to a pandas `Grouper` object, but this object doesn't support `base` as a kwarg in versions >= 2.0. This was fixed later by the xarray team to raise a `FutureWarning` instead of passing `base = None` to pandas (which causes a `TypeError`).

In short, if we want to unpin pandas and keep using xarray's `resample` method, then we need to drop support for Python 3.8

This issue was brought up in the course of addressing issue #618.

rt17603:

Dropping support for Python 3.8. This has become necessary because the dependencies for openghg no longer support it (e.g. `xarray` dropped support as of [v2023.02.0](https://docs.xarray.dev/en/stable/whats-new.html#v2023-02-0-feb-7-2023)).

As a part of this, this also adds a check for [Python 3.12](https://www.python.org/downloads/release/python-3120/) which was released on 02/10/2023.

- [x] Closes #817

* **Additional note**

Had to update branch protection rules on GitHub for both `devel` and `master` to remove "Run the tests (3.8)" from the list of checks and replace this with "Run the tests (3.11)".

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -22,7 +22,7 @@ jobs:
       fail-fast: false
       matrix:
         operating-system: [ubuntu-latest]
-        python-version: ["3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.9", "3.10", "3.11", "3.12"]
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,15 +25,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


brendan-m-murphy (dismissed):
Looks good to me! I'll wait for Gareth to check it before merging.

gareth-j (changes requested):
We just need to update the root `__init__.py` file as well to update the Python version check there.

Issue #815, opened 2023-11-06
Issue with deleting and re-adding ICOS data from object store (remnant of previous entry)


rt17603:
From @joe-pitt on 03/11/2023
"OK so I deleted all the ICOS obs from paris_openghg_store and tried to repopulate it with updated data, but randomly some data from the 2018 drought paper crept in. I updated my populate script to exclude this random data, deleted all the icos data again, and tried to repopulate again. However now when I try to add the data I get the message:

WARNING:openghg.store:Note: There is no new data to process.

When I look in the store there is no icos data there. I guess there is still some remenant of the data in the store somehow - any suggestions on how to force an update using the current openghg?
"

brendan-m-murphy:
I can confirm that there isn't any ICOS metadata in the ObsSurface metastore, but there are a lot of *retrieved* hashes in the metadata stored by the `ObsSurface` class when adding data to the object store.

So, whatever method was used to delete the data didn't clear the "retrieved hashes", which is why OpenGHG is saying the data has already been added.

@joe-pitt could you share the script you used, or describe the method you used to delete the data? It was successfully removed from the metastore, but not from the metadata that tracks what data has been added. (And actually, I see no reasonable way to decide how to remove these hashes unless the file hash is stored with some identifying info somewhere else.)

joe-pitt:
I used the following:

```
dm_lookup = dataobjects.data_manager(data_type="surface",
                                     store=store,
                                     datasource="icoscp")
uuids = list(dm_lookup.metadata.keys())
dm_lookup.delete_datasource(uuid=uuids)

```

If there isn't a simple way to resolve this then don't worry - I can stick to using a separate store for these obs for now, and then update the Paris store once the new functionality for updating data gets implemented.

brendan-m-murphy:
@rt17603 discussed this on Teams:

- In the near future, there will be a "force" option that ignores the stored hashes (part of the `new_object_store` branch)
- We should store some info connecting the hashes to the data sources:
    - some files will result in multiple datasources
    - multiple files can be added to the same datasource (for different time periods)
    - so, we need to associate to each file/retrieved hash the (list of) UUID(s) of the data source(s) the data was added to, as well as the time period that the data spans
    - we need to update data manager to remove hashes when a datasource is deleted

brendan-m-murphy:
We're going to add a `force` option to ignore hashes when retrieving data from ICOS, NOAA, etc. as a short term fix.

[...]

PR #819, opened 2023-11-07
Iss #815 fix for re-adding ICOS data after removing it

brendan-m-murphy:
Added `force` option to `retrieve_atmospheric` and `ObsSurface.store_data`. If `force=True`, then retrieved hashes will be ignored.

Two tests were added, one to check that if `force=False` (the default option) then `retrieve_atmospheric` issues a warning that there is no new data to process. If `force=True`, then this warning is not issued (and the data is added again, although this is not explicitly checked).

- [x]  Quick fix for #815 

### DIFF
--- a/openghg/retrieve/icos/_retrieve.py
+++ b/openghg/retrieve/icos/_retrieve.py
@@ -50,8 +50,7 @@ def retrieve_atmospheric(
                 - "never" - don't update mismatches and raise an AttrMismatchError
                 - "from_source" / "attributes" - update mismatches based on attributes from ICOS Header
                 - "from_definition" / "metadata" - update mismatches based on input metadata
-        force: If True, disregard previously retrieved hashes. Use this if you are unable to re-add
-            previously deleted data.
+        force: Force adding of data even if this is identical to data stored (checked based on previously retrieved file hashes).
     Returns:
         ObsData, list[ObsData] or None
     """
@@ -182,8 +181,7 @@ def local_retrieve(
                 - "never" - don't update mismatches and raise an AttrMismatchError
                 - "from_source" / "attributes" - update mismatches based on attributes from ICOS Header
                 - "from_definition" / "metadata" - update mismatches based on input metadata
-        force: If True, disregard previously retrieved hashes. Use this if you are unable to re-add
-            previously deleted data.
+        force: Force adding of data even if this is identical to data stored (checked based on previously retrieved file hashes).
     Returns:
         ObsData, list[ObsData] or None
     """
@@ -239,7 +237,12 @@ def local_retrieve(
         for data in standardised_data.values():
             measurement_data = data["data"]
             # These contain URLs that are case sensitive so skip lowercasing these
-            skip_keys = ["citation_string", "instrument_data", "dobj_pid", "dataset_source"]
+            skip_keys = [
+                "citation_string",
+                "instrument_data",
+                "dobj_pid",
+                "dataset_source",
+            ]
             metadata = to_lowercase(data["metadata"], skip_keys=skip_keys)
             obs_data.append(ObsData(data=measurement_data, metadata=metadata))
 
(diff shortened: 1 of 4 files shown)


openghg/retrieve/icos/_retrieve.py
```diff
@@ -49,6 +50,8 @@ def retrieve_atmospheric(
                 - "never" - don't update mismatches and raise an AttrMismatchError
                 - "from_source" / "attributes" - update mismatches based on attributes from ICOS Header
                 - "from_definition" / "metadata" - update mismatches based on input metadata
+        force: If True, disregard previously retrieved hashes. Use this if you are unable to re-add
```
rt17603, line 53: In this circumstance this comment is true but the key can also be used to replace identical data as well.
On the `new_object_store` branch this key word is defined as
> force: Force adding of data even if this is identical to data stored.

and I think that does fit with this definition as well so could combine the two in some way like this:

```suggestion
        force: Force adding of data even if this is identical to data stored. This can also be used to disregard previously retrieved hashes if unable to re-add data.
```

rt17603, line 53: In this circumstance this comment is true but the `force` key can also be used to replace "identical" data in general.

On the `new_object_store` branch this key word is defined as

> force: Force adding of data even if this is identical to data stored.

and I think that does fit with this definition as well in this case or could augment this to:

> force: Force adding of data even if this is identical to data stored (checked based on previously retrieved file hashes).

rt17603, line 53: In this circumstance this comment is true but the `force` key can also be used to replace "identical" data in general.

On the `new_object_store` branch this key word is defined as

> force: Force adding of data even if this is identical to data stored.

and I think that does fit with this definition as well in this case or could augment this to something like:

> force: Force adding of data even if this is identical to data stored (checked based on previously retrieved file hashes).

openghg/retrieve/icos/_retrieve.py
```diff
@@ -177,6 +182,8 @@ def local_retrieve(
...
```
rt17603, line 185: Same as previous comment on documentation.

openghg/store/_obssurface.py
```diff
@@ -483,18 +483,24 @@ def validate_data(data: Dataset, species: str) -> None:
...
+        required_metakeys: Optional[Sequence] = None,
     ) -> Optional[Dict]:
         """This expects already standardised data such as ICOS / CEDA
 
         Args:
             data: Dictionary of data in standard format, see the data spec under
-            Development -> Data specifications in the documentation
+                Development -> Data specifications in the documentation
             overwrite: If True overwrite currently stored data
+            force: If True, disregard previously retrieved hashes. Use this if you are unable to re-add
```
rt17603, line 498: Same as other comment on keyword definition.

openghg/store/_obssurface.py
```diff
@@ -503,7 +509,10 @@ def store_data(
         # Very rudimentary hash of the data and associated metadata
         hashes = hash_retrieved_data(to_hash=data)
         # Find the keys in data we've seen before
```
rt17603, line 511: Could we rename `seen_before` to something like `file_hashes_to_compare` given this new check and update the comment. I just think otherwise the subtly here of what is being checked may not be clear.

rt17603 (changes requested):
Looks great and some useful tests included here.

A few small comments on the definition of `force` in the doc-string and a suggested change in variable name but otherwise this is pretty much ready to merge.

openghg/retrieve/icos/_retrieve.py
brendan-m-murphy: I went with the second option, since a data source lookup could prevent adding data currently. (I think there is an overwrite option to get around this, but I'm not sure if it's implemented for `store_data`.)

openghg/store/_obssurface.py
brendan-m-murphy: Done

rt17603 (approved):

openghg/retrieve/icos/_retrieve.py
rt17603: For reference, the `overwrite` option is being superceded by the more nuanced `if_exists`, `save_current` and `force` keywords and will be deprecated. Regardless though makes sense.

PR #822, opened 2023-11-10
Added black line length = 110 to toml

brendan-m-murphy:
Black is configured using the `pyproject.toml` file (unlike flake8, which is configured in the setup.cfg file).

Putting `line-length=110` in the pyproject toml means that this setting can be found by IDEs.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -34,7 +34,7 @@ jobs:
       - name: Format code with black
         run: |
           python -m pip install black
-          black --check --line-length 110 openghg/
+          black --check openghg/
       - name: Lint with flake8
         run: |
           python -m pip install --upgrade pip wheel setuptools flake8
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -14,7 +14,6 @@ repos:
     rev: 22.8.0
     hooks:
       - id: black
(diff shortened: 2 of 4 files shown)


rt17603 (changes requested):
Looks good and this seems like a good solution.

If we're changing this to be set centrally like this we should also make sure to remove this from the other places where line length has been set explicitly:
 - e.g. [pre-commit](https://github.com/openghg/openghg/blob/black-config/.pre-commit-config.yaml#L17C9-L17C34)
 - e.g. [workflow yaml file](https://github.com/openghg/openghg/blob/black-config/.github/workflows/workflow.yaml#L37)

gareth-j:
> Looks good and this seems like a good solution.
> 
> If we're changing this to be set centrally like this we should also make sure to remove this from the other places where line length has been set explicitly:
> 
>     * e.g. [pre-commit](https://github.com/openghg/openghg/blob/black-config/.pre-commit-config.yaml#L17C9-L17C34)
> 
>     * e.g. [workflow yaml file](https://github.com/openghg/openghg/blob/black-config/.github/workflows/workflow.yaml#L37)
> 
> 
> @gareth-j - do you concur with that?
> 
> (perhaps worth checking that does all work as expected though)

Yeah that sounds good.

rt17603 (approved):

PR #823, opened 2023-11-14
Edgar parser

brendan-m-murphy:
This is a very rough version of a parser for EDGAR data in netCDF format.

It assumes that the data is annual, and that there is only one flux source per netCDF file.

It uses the scitools `iris` package to do regridding instead of `xesmf`.

Also, the organisation of the modules doesn't really make sense now, since `openghg.standardise.emissions._openghg.py` now has a generic parser, plus parsers for openghg and edgar formats.

On the other hand, the code works and can be used to regrid and add EDGAR flux files to the object store.

### DIFF
--- a/openghg/standardise/_standardise.py
+++ b/openghg/standardise/_standardise.py
@@ -463,6 +463,7 @@ def standardise_flux(
     continuous: bool = True,
     overwrite: bool = False,
     store: Optional[str] = None,
+    source_format: str = "openghg",
 ) -> dict:
     """Process flux data
 
@@ -478,6 +479,7 @@ def standardise_flux(
         continuous: Whether time stamps have to be continuous.
         overwrite: Should this data overwrite currently stored data.
         store: Name of store to write to
+        source_format: format of raw data file (e.g. "openghg" or "edgar")
 
     Returns:
         dict: Dictionary of Datasource UUIDs data assigned to
@@ -530,6 +532,7 @@ def standardise_flux(
             continuous=continuous,
             chunks=chunks,
             overwrite=overwrite,
+            source_format=source_format,
         )
 
 
--- a/openghg/standardise/emissions/__init__.py
+++ b/openghg/standardise/emissions/__init__.py
@@ -1,1 +1,1 @@
-from ._openghg import parse_openghg
+from ._openghg import parse_openghg, parse_edgar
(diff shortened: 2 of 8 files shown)


brendan-m-murphy:
Thanks @qq23840 for the fixes. 

To incorporate this into OpenGHG, I think @rt17603 said this should be part of the "transform" workflow, since parsers are meant to take one input and process it to one entry in the object store, while the transform workflow supports making multiple entries in the object store from a single input (in this case, by regridding to different domains).

This will require
1) restoring the emissions/flux parsers to their original state
2) moving the edgar "parser" code to the transform submodule.

It might be easier to do this by cherry picking code from this branch into a new branch after release 0.8.

Another issue is the regridding functions: they add a new dependency (scitools "iris" package), and we already have regridding functionality, but it requires a package (xesmf) that can't be installed with pip. xarray has improved support for iris now, so it might be worth switching to iris.

rt17603:
> Thanks @qq23840 for the fixes.
> 
> To incorporate this into OpenGHG, I think @rt17603 said this should be part of the "transform" workflow, since parsers are meant to take one input and process it to one entry in the object store, while the transform workflow supports making multiple entries in the object store from a single input (in this case, by regridding to different domains).
> 
> This will require
> 
>     1. restoring the emissions/flux parsers to their original state
> 
>     2. moving the edgar "parser" code to the transform submodule.
> 
> 
> It might be easier to do this by cherry picking code from this branch into a new branch after release 0.8.

Yes, there was actually a current workflow for adding EDGAR data from a database folder which is a bit out of date that we could emulate / incorporate this into - https://github.com/openghg/openghg/blob/devel/openghg/transform/emissions/_edgar.py. This works off a whole folder and you wanted to work from a file but the principle should be similiar.

For the `standardise` functionality, the idea was that one input file would produce the same sets of datasources whereas for `transform` this would be able to create *different* data sources dependent on the inputs. For example for an obs file for a site with data for one gas, this would always create one data source and running with the same file would always create the same data source. For an EDGAR database file the same file could be used but depending on the domain specified this could produce different data sources each time this is run. In this case specifying the domain changes the data source being created.

> Another issue is the regridding functions: they add a new dependency (scitools "iris" package), and we already have regridding functionality, but it requires a package (xesmf) that can't be installed with pip. xarray has improved support for iris now, so it might be worth switching to iris.

As we discussed earlier, there isn't a perfect solution around regridding with both `xesmf` and `iris` having limitations. If we are able to incorporate the code into this PR into the pre-existing functionality discussed above we could see if this could work with `xesmf` for now as long as we make sure to give useful advice on how to use it with conda/micromamba.

Issue #765, opened 2023-08-24
Umbrella issue: short term solution for concurrency problem


rt17603:
For the concurrent write (and potentially read) problem we can execute the following steps for a quicker fix on the path to a longer term solution (e.g. using MongoDB - #756 )

- [x] #762 - being addressed by #763
- [x] #695 
- [x] #764

chryswoods:
@gareth-j will add in support for lock files after doing work on zarr. This will be via time-limited lock-files. It is noted that this is a short-term fix that will be resolved properly by the mongodb work #756 and zarr work in the special release.

This issue relates only to the short-term fix needed for 0.7.0.

This solution may benefit from a short technical meeting focussed around this topic.

rt17603:
Closed with final PR #695

rt17603:
Closed with final PR https://github.com/openghg/openghg/issues/695

PR #837, opened 2023-11-28
Iss695 object store lockfile

### LINKED ISSUE #695: Object store lockfile
So we don't encounter issues with a single data storage class being used by multiple people at the same time we should get them to check for the existence of a lockfile at the start of their `read_file` functions. Something like

```
# Checks if the lockfile exists for this object in the bucket
self._set_lock()
# do some work and assign the data
self._release_lock()
```

brendan-m-murphy:
`ClassicMetaStore` was removed and replaced by `DataClassMetaStore`, 
since only the children of `BaseStore` use `ClassicMetaStore` directly.
`open_metastore` now uses `TinyDBMetaStore` directly.
This means that we don't need a class method for use in `BaseStore`.

File locks were added to `DataClassMetaStore`, and used in `BaseStore.assign_data`.
Tests were added to show that
1) the locks stop other instances of `DataClassMetaStore` from acquiring a lock on a locked file
2) the locks are "advisory", meaning that they don't actually stop access to the file. So, read operations
work on locked files. (This also means that we might need to add locks to other places where write operations occur.)

This could be improved by making the "read" and "read/write" modes of our custom TinyDB storage
class be enforced via locks. Thus any write operation much acquire a lock.

I didn't do this initially, since acquiring a lock in `BaseStore.assign_data` means the lock is held for
less time than if it were acquired when a child class of `BaseStore` is instantiated (e.g. from `standardise`).

- [ ] Closes #695 #765 

### DIFF
--- a/openghg/objectstore/metastore/_classic_metastore.py
+++ b/openghg/objectstore/metastore/_classic_metastore.py
@@ -27,13 +27,13 @@ from contextlib import contextmanager
 import json
 from typing import Literal, Optional
 
-import tinydb
-from tinydb.middlewares import Middleware
-
-from openghg.objectstore import exists, get_object, set_object_from_json
+from filelock import FileLock
+from openghg.objectstore import exists, get_object, set_object_from_json, get_object_lock_path
 from openghg.objectstore.metastore._metastore import TinyDBMetaStore
 from openghg.types import MetastoreError
 from openghg.util import hash_string
+import tinydb
+from tinydb.middlewares import Middleware
 
 
 object_store_data_classes = {  # TODO: move this to central location after ObjectStore PR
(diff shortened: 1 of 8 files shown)


brendan-m-murphy:
> So I'm getting a `FileNotFoundError` when running the tests locally.

@gareth-j I added the fix you suggested. I think it helps; I created a `tinydb.Storage` subclass with file locking, and got `FileNotFound` errors that went away with your fix. I think maybe without creating the directories when getting the lock file path, you won't get an error if a read is made the metastore before using `assign_data`, since an initial read will create the directories.

I'm not sure why it worked for me and the CI, but not you.

(I haven't pushed the locking `tinydb.Storage` subclass yet, because I'm not sure if we want it. I was getting conflicting results testing the locking storage in isolation vs. testing it via `Datamanager`, so I'm not confident that it is working.)

openghg/objectstore/metastore/_classic_metastore.py
```diff
@@ -25,15 +25,15 @@
...
-from typing import Literal, Optional, TypeVar
+from typing import Any, cast, Literal, Optional, Type
 
-import tinydb
-from tinydb.middlewares import Middleware
-
-from openghg.objectstore import exists, get_object, set_object_from_json
+from filelock import FileLock
+from openghg.objectstore import exists, get_object, set_object_from_json, get_object_lock_path
 from openghg.objectstore.metastore._metastore import TinyDBMetaStore
```
gareth-j, line 32:
```suggestion
from openghg.objectstore.metastore import TinyDBMetaStore
```

gareth-j (approved):
Looks good!

gareth-j (approved):
Looks good!

PR #842, opened 2023-12-08
Ensure parent folder is created for lockfile path

gareth-j:

I had a lot of errors caused by it trying to create a lockfile in a folder that didn't exist yet. This is a workaround for that and makes all my tests pass. Happy for the same thing to be done somewhere else though.


gareth-j:
Closing as implemented in #837.

PR #850, opened 2023-12-14
Update changelog ready for release

gareth-j:

Update the changelog ready for release.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,45 +5,34 @@ All notable changes to OpenGHG will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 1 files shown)


rt17603 (approved):
Looks good - only small comment was whether we wanted to add the "###Added", "###Fixed" and "###Changed" headers under the Unreleased section as well?

gareth-j:
> Looks good - only small comment was whether we wanted to add the "###Added", "###Fixed" and "###Changed" headers under the Unreleased section as well?

We haven't been doing that previously I don't think. We've just added them as needed.

rt17603:
> > Looks good - only small comment was whether we wanted to add the "###Added", "###Fixed" and "###Changed" headers under the Unreleased section as well?
> 
> We haven't been doing that previously I don't think. We've just added them as needed.

Yeah you're right but I believe last time these headings were created differently in the subsequent release until they got corrected to match to the previous details. Not a big deal just thought that may help keep the log consistent.

PR #877, opened 2024-01-25
Load Matt's combined .nc files

joe-pitt:
Made some changes so that we can load Matt's combined (i.e. more than one instrument) .nc files:

- Get the correct instrument name from the filename
- Added relevant entries to json file - setting sampling_period to "varies"
- Handle this in the model scenario code
- Adjust the timestamp based on the sampling_period read from the file, rather than from the json
- Change the GCMD sampling_period in the json to 1s (this has always been an instantaneous measurement)
- Change tests accordingly

Some tests are failing, but the same tests fail for me on the agage_new_format branch. So I think we can merge these changes into that branch, then fix remaining issues when we merge agage_new_format back to devel.

### DIFF
--- a/openghg/analyse/_scenario.py
+++ b/openghg/analyse/_scenario.py
@@ -629,6 +629,9 @@ class ModelScenario:
             sampling_period = obs_attributes["sampling_period"]
             if sampling_period == "NOT_SET":
                 infer_sampling_period = True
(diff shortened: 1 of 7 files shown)


qq23840 (approved):
All looks good to me!

PR #892, opened 2024-02-05
Check current chunks

### LINKED ISSUE #888: Store the chunking data in the metastore, error if chunks are different
We need to be able to store and retrieved the chunk sizes of previously added data for the Datasource. We need to be able to do a metastore lookup at the start of the standardise function to check if there's a current Datasource.

gareth-j:

This adds checking of current chunking to the `LocalZarrStore` class. When data is added to a current version the chunking of the incoming data is checked and compared with that in the currently stored data. If it's different then the incoming data is rechunked to match that of the stored data and the user is informed of this.

- [x] Closes #888

### DIFF
--- a/openghg/store/storage/_localzarrstore.py
+++ b/openghg/store/storage/_localzarrstore.py
@@ -329,10 +329,18 @@ class LocalZarrStore(Store):
             return {}
 
         if incoming_actually_chunked != stored_actually_chunked:
-            logger.warning(
-                f"Chunking of incoming data {incoming_actually_chunked} does not match stored data {stored_actually_chunked}\n"
-                "Using stored chunking schema."
-            )
+            if not incoming_actually_chunked:
+                msg = (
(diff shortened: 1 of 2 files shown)


SutarPrasad (changes requested):
Looks good.
I think if the chunks parameter is not passed but we have already chunked stored data then, in that case, it should showcase that `Chunking Mismatch and chunking the data based on the present chunk schema`

gareth-j:
> Looks good. I think if the chunks parameter is not passed but we have already chunked stored data then, in that case, it should showcase that `Chunking Mismatch and chunking the data based on the present chunk schema`

OK I've improved the error messages.

SutarPrasad:
> > Looks good. I think if the chunks parameter is not passed but we have already chunked stored data then, in that case, it should showcase that `Chunking Mismatch and chunking the data based on the present chunk schema`
> 
> OK I've improved the error messages.

Hi Gareth,
Currently its not chunking on present chunk sizes of the store if we don't supply the `chunks` parameter.

PR #899, opened 2024-02-07
Allowing period to be specified when this cannot be inferred from the data

### LINKED ISSUE #887: On standardise functions `continuous=False` doesn't work
At the moment passing `False` to `continuous` doesn't stop it failing, look at why.

rt17603:

This fixes a bug so that a `period` can be specified when `continuous=False` is included for the `infer_date_range` function. This is used within the `openghg.standardise`: `standardise_footprint`, `standardise_flux` and `standardise_bc` functions.

*Note: This is a short term fix to address the immediate issue of struggling to add a period when this cannot be inferred from the data (which will be more of an issue with the new zarr store set up as well). Issue #911 has been opened for longer term logic changes.*

- [x] Closes #887  (Replace xxxx with the Github issue number)


openghg/store/_infer_time.py
```diff
@@ -9,7 +9,7 @@
 logger = logging.getLogger("openghg.store")
 logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
 
-TupleTimeType = Tuple[Union[int, float], str]
+TupleTimeType = Tuple[Union[int, float, None], Union[str, None]]
```
brendan-m-murphy, line 12: Maybe this should be a dataclass with a classmethod that does what `parse_period` does?

rt17603: Could you add that as a suggestion to Issue #911 ?

brendan-m-murphy: Okay, I've done that. The comment here is more around readability: `TupleTimeType` is somewhat complicated and the parts have a clear meaning. You could also just use a named tuple, e.g. `TimePeriod = namedtuple("TimePeriod",  ["value", "unit"])`, although maybe the `NamedTuple` class from `typing` would be better:

```
from typing import NamedTuple

class TimePeriod(NamedTuple):
   value: Union[int, float, None] = None
   unit: Optional[str] = None

input_freq  = parse_period(period) if period else TimePeriod()
```

brendan-m-murphy: This doesn't really affect the use of `parse_period` since `tp = TimePeriod(1, "H")` and `u, v = tp` give `u == 1` and `v == "H"`.

brendan-m-murphy: Also, `TimeTupleType` has different definitions in `util._time` and `store._infer_time`, so they should be synchronised. I guess I would suggest defining a `typing.NamedTuple` subclass to replace `TupleTimeType` in `util._time` and importing that in `store._infer_time`, although importing the type alias could work.

brendan-m-murphy (changes requested):
A few changes for clarity:
1) Update the docstring of `infer_date_range` to say that that the specified period will be used if a period cannot be inferred and continuous is False. Also mention that the inferred period will be used instead of a specified period, if it can be inferred. (The docstring has details about what will happen if there is only one time point, but doesn't elaborate if there is more than one time point.)

2) I would suggest using `typing.NamedTuple` in place of the type alias `TupleTimeType`. I think it would make the code clearer, and you've already make several changes to make code around these tuples easier to understand (e.g. the `null_freq` variable.) Also, the `TupleTimeType` in `util._time` isn't the same as the alias in `store._infer_time`, so at a minimum, these should be the same (e.g. import the alias from `util._time` into `store._infer_freq`).

Otherwise, the updates looks good.

openghg/store/_infer_time.py
rt17603: Ok - I've integrated in the suggested `TimePeriod` class into `openghg.types._types` for this and used this in the two places this is defined. This can be revisited when addressing Issue #911 if needed.

rt17603:
Failing test to do with checking hashes should be fixed by #929 . Other checks pass for this.

brendan-m-murphy (approved):
Looks good!

rt17603:
Failing check is from test unrelated to this PR. This has been fixed but was merged into `new_object_store` branch rather than `devel` which is why this fix can't be pulled onto this branch at the moment.

PR #917, opened 2024-02-19
Ensure dataset chunking matches what's already there

gareth-j:

This makes sure incoming data with different chunks is actually stored with the original schema. In the current version if data is passed in with different chunks then the rechunk in `LocalZarrStore.add` doesn't get rid of the chunks in different coordinates.

### DIFF
--- a/openghg/store/storage/_localzarrstore.py
+++ b/openghg/store/storage/_localzarrstore.py
@@ -331,6 +331,12 @@ class LocalZarrStore(Store):
                 )
 
             logger.warning(msg)
+
+            # For coordinates we haven't chunked over we'll use the full size
+            for k in dataset.coords:
+                if k not in stored_actually_chunked:
+                    stored_actually_chunked[k] = dataset.sizes[k]
+
             return stored_actually_chunked
 
         return {}
--- a/openghg/store/storage/_localzarrstore.py
+++ b/openghg/store/storage/_localzarrstore.py
@@ -333,7 +333,7 @@ class LocalZarrStore(Store):
             logger.warning(msg)
 
             # For coordinates we haven't chunked over we'll use the full size
-            for k in dataset.coords:
+            for k in dataset.dims:
                 if k not in stored_actually_chunked:
                     stored_actually_chunked[k] = dataset.sizes[k]
 
--- a/tests/standardise/test_standardise.py
+++ b/tests/standardise/test_standardise.py
@@ -408,4 +408,5 @@ def test_standardise_footprint_different_chunking_schemes(caplog):
     search_results = search(data_type="footprints", model="chunk_model", store="user")
     fp_data = search_results.retrieve_all()
 
+    # Check that the chunking scheme is what was specified with the first standardise call
     assert dict(fp_data.data.chunks) == {"time": (2, 2, 2), "lat": (12,), "lon": (12,), "height": (20,)}
--- a/tests/standardise/test_standardise.py
+++ b/tests/standardise/test_standardise.py
@@ -299,44 +299,6 @@ def test_standardise_footprint():
     assert "tmb_europe_test_model_10m" in results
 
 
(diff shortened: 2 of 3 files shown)


rt17603:
> I get the following failing tests when I run this locally for me:
> 
> ```
> =============================================================================================== 
> short test summary info 
> ===============================================================================================
> FAILED tests/store/storage/test_localzarrstore.py::test_update - FileNotFoundError: [Errno 2] No such file or directory: '/home/rt17603/openghg/tests/data/footprints/TAC-100magl_UKV_EUROPE_201607.nc'
> FAILED tests/store/storage/test_localzarrstore.py::test_match_chunking - AssertionError: assert not {'time': 31}
> ============================================================== 2 failed, 432 passed, 30 skipped, 12 xfailed, 2 xpassed, 680 warnings in 90.84s (0:01:30) ==============================================================
> ```
> 
> Full details attached: [2024-02-19_checkChunkingChecks_tests.log](https://github.com/openghg/openghg/files/14334035/2024-02-19_checkChunkingChecks_tests.log)

The first failing test seems to also be present in the `Iss561/addZarrStore` branch.

tests/standardise/test_standardise.py
```diff
@@ -370,3 +371,41 @@ def test_cloud_standardise(monkeypatch, mocker, tmpdir):
...
+        height=height,
+        domain=domain,
+        store="user",
+        chunks={"time": 2, "lat": 5, "lon": 5},
+    )
+
+    search_results = search(data_type="footprints", model="chunk_model", store="user")
+    fp_data = search_results.retrieve_all()
+
+    assert dict(fp_data.data.chunks) == {"time": (2, 2, 2), "lat": (12,), "lon": (12,), "height": (20,)}
```
rt17603, line 411: Is the idea here that the `lat`, `lon` chunking hasn't been applied when more data is added? If so, would be good to add a comment / doc string to make that clear.

tests/store/storage/test_localzarrstore.py
```diff
@@ -200,12 +200,16 @@ def test_match_chunking(store):
...
-    data_a_chunked = data_a.chunk(chunks)
     store.add(version="v0", dataset=data_a)
+
+    data_a_chunked = data_a.chunk(chunks)
     chunking = store.match_chunking(version="v0", dataset=data_a_chunked)
 
-    assert not chunking
+    # As the data we originally put in wasn't chunked then we get the full size of the time coordinate
+    # which is 31 here
+    assert chunking == {'time': 31}
```
rt17603, line 212: Will this test still be valid when the default chunking is included? Or would this need to be updated?

rt17603 (approved):
A few small clarifications requested but otherwise looks good.

tests/store/storage/test_localzarrstore.py
gareth-j: Yeah it'll still be valid as this is a layer lower than the default chunking. This is just making sure the correct chunking is used.

openghg/store/storage/_localzarrstore.py
```diff
@@ -331,6 +331,12 @@ def match_chunking(self, version: str, dataset: xr.Dataset) -> Dict[str, int]:
                 )
 
             logger.warning(msg)
+
+            # For coordinates we haven't chunked over we'll use the full size
+            for k in dataset.coords:
```
brendan-m-murphy, line 336:
```suggestion
            for k in dataset.dims:
```
Chunks and sizes are dicts keyed by dims. Usually our coord names mirror the dim names, but it's possible that they differ.

brendan-m-murphy (approved):
Looks good to me except the one suggestion I made

tests/standardise/test_standardise.py
gareth-j: Done

PR #920, opened 2024-02-20
Set default chunk values

### LINKED ISSUE #908: Default chunks for the zarr store
### What is your issue?

Need to have the default chunk value set for the Zarr storage move. 
Especially for `co2` data files it is required to at least chunk the time dimension.

As of the current investigation having the default chunk value set to `chunks = {time:100}` is the optimum value for the `standardise_*` functions.

gareth-j:

Set default chunk values and checking functionality to the base storage class and data type storage classes.

- [x] Closes #908 

## TODO

- [x] Update the max chunk size to 300 MB and use MB units instead of bytes to make it easier to read


openghg/store/_footprints.py
```diff
@@ -313,8 +314,13 @@ def read_file(
         if not filepath:
             return {}
 
-        if chunks is None:
-            chunks = {}
+        chunks = self.check_chunks(
```
brendan-m-murphy, line 317: we need to use the species to update `high_time_resolution` etc before this step

openghg/store/_footprints.py
```diff
@@ -617,3 +622,36 @@ def validate_data(
...
+
+        Args:
+            high_time_resolution : Set footprint variable to be high time resolution.
+            high_spatial_resolution : Set footprint variables include high and low resolution options.
+            short_lifetime: Include additional particle age parameters for short lived species.
+        Returns:
+            dict: Chunking schema for footprint data.
+        """
+        if high_spatial_resolution or short_lifetime:
+            raise NotImplementedError(
```
rt17603, line 643: Could we add to this message something like "using default footprint chunking"?

openghg/store/storage/_chunking.py
```diff
@@ -0,0 +1,22 @@
...
+    """
+
+    # TODO - open to alternative names here
+
+    variable: str
+    chunks: Dict[str, int]
+    secondary_dims: List[str]
+    # max chunk size in bytes (250 MB)
+    # TODO - do we want to make this configurable?
+    max_chunk_size: int = 262144000
```
rt17603, line 22: As Brendan mentioned - we may need to up this to the equivalent of 300MB to be able to manage high time resolution footprint chunks.

rt17603 (commented):
Looks good - wanted to put in a few comments for now but will do one final look over next week.

openghg/store/storage/_chunking.py
gareth-j: Done.

rt17603 (approved):
Looks good to me - useful documentation updates as well.

One general comment is that we should probably be adding doc-strings for our tests as it helps to identify the purpose of the tests (and can be useful if it stops working).

PR #930, opened 2024-02-22
Mypy fix for local_zarr_store

brendan-m-murphy:

`xr.Dataset.dims` returns a `Mapping` with `Hashable` keys, while
our chunk `dict` specify strings for keys. To appease mypy, I converted
the Hashable keys to strings. (It is conceivable that a dim is some
Hashable with no custom str method, so the call to `str` would return
the `repr` instead, which might be ugly... but at least it will never
cause an error.)


rt17603 (approved):
Thanks for putting this in - I think this seems like a reasonable approach to fix this for now so looks good!

PR #937, opened 2024-02-29
Fix for `test_hash_retrieved_data`

gareth-j:
Fixes the hash test error in `test_hash_retrieved_data`


rt17603 (approved):
Good idea to get this fix into `devel` directly.

PR #940, opened 2024-03-01
Fix for standardisation test error

gareth-j:
* **Summary of changes**
This fixes the test failure in `test_standardise_obs_openghg` that was being caused by the same file being processed previously. I've added a new function `clear_test_store` which I've used to clear the test store before this test is run so the file can be standardised properly. This function just clears the store that's specified instead of deleting all three test stores.

- [ ] Closes #932 

### DIFF
--- a/tests/helpers/__init__.py
+++ b/tests/helpers/__init__.py
@@ -2,6 +2,7 @@ from .cfchecking import check_cf_compliance
 from .helpers import (
     all_datasource_keys,
     call_function_packager,
+    clear_test_store,
     clear_test_stores,
     get_bc_datapath,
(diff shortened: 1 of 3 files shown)


rt17603:
Should we merged this into `devel` so this can go into `0.7.1` release and then can be merged into `new_object_store`?

rt17603:
> Should this fix be merged into `devel` directly so this can go into `0.7.1` release and then can be merged from `devel` into `new_object_store`?

Running the tests again on `devel` and `new_object_store` this doesn't seem to be affecting `devel` - so merging this into `new_object_store` directly seems fine.

PR #941, opened 2024-03-01
Adding pytest-timeout to stop any tests taking longer than 300s

gareth-j:

This adds the [`pytest-timeout`](https://github.com/pytest-dev/pytest-timeout) plugin which will stop any individual tests after they've run for a specified time, here 60s.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -7,3 +7,6 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 110
+
+[pytest]
+timeout = 60
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -9,4 +9,4 @@ build-backend = "setuptools.build_meta"
 line-length = 110
 
 [pytest]
-timeout = 60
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -7,3 +7,6 @@ build-backend = "setuptools.build_meta"
 
 [tool.black]
 line-length = 110
+
+[pytest]
+timeout = 60
```
rt17603, line 12: If this is 60 seconds, I think many of our tests would legitimately take longer than that? If we're setting a timeout we probably want to make this very long as a failsafe to end tests which clearly aren't going to finish running.

rt17603: From timing this locally,
$ pytest --durations=10
the longest test on my system is ~10s but I still don't think it hurts to set a longer extreme timeout of (e.g. 5 mins)

rt17603 (approved):
Seems like a useful failsafe to include - would you mind changing the title and PR text to reflect the new timeout value (or remove number from title).

PR #942, opened 2024-03-01
Try skipping intem test

gareth-j:

Try and skip the INTEM standardisation test, this seems to be where the CI gets stuck.

### DIFF
--- a/tests/standardise/test_standardise.py
+++ b/tests/standardise/test_standardise.py
@@ -319,6 +319,7 @@ def test_standardise_flux():
     assert "co2_gpp-cardamom_europe" in proc_results
 
 
+@pytest.mark.skip(reason="Testing removal of this to find stuck test")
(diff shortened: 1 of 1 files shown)


tests/standardise/test_standardise.py
```diff
@@ -319,6 +319,7 @@ def test_standardise_flux():
     assert "co2_gpp-cardamom_europe" in proc_results
 
 
+@pytest.mark.skip(reason="Testing removal of this to find stuck test")
```
rt17603, line 322: We do want to be careful skipping tests but if we expect this to be fixed in v0.8.0 (new_object_store) this should included in the reason string. We'd also want to add an issue to make sure this gets added back in the next version - could even put this issue on the Special Release milestone.

rt17603:
@gareth-j - can we close this now (I think you said it's not relevant after new_object_store goes in)?

PR #943, opened 2024-03-01
Remove codecov

gareth-j:
* **Summary of changes**

This removes codecov as it's adding itself to PRs where it's not wanted and its coverage isn't correct for us.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -55,14 +55,7 @@ jobs:
       - name: Test with pytest
         run: |
           pip install --upgrade pytest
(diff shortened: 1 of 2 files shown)


rt17603 (approved):
This is fine for now as codecov is being strange but could you add an issue to attempt to add this back as it is useful to have if we can get it to me more usable for us,

gareth-j:
> This is fine for now as codecov is being strange but could you add an issue to attempt to add this back as it is useful to have if we can get it to me more usable for us,

I've opened an issue to look at either adding it back in or to look at alternatives: https://github.com/openghg/openghg/issues/945

PR #947, opened 2024-03-03
Add TOB entry to process_gcwerks_parameters.json

joe-pitt:
Add an entry for Taunus (TOB) site in process_gcwerks_parameters.json


gareth-j (approved):
I think this is good to go in.

PR #962, opened 2024-03-20
Temporary check for old object stores when using `0.8.0`

gareth-j:
* **Summary of changes** 

This adds some functionality to the `read_local_config` function to check for old non-Zarr object stores and remove them from the config dictionary that OpenGHG will use internally when doing lookups etc. This does **not** remove them from the config file on disk. 

It could be some useful temporary functionality to allow users to have multiple virtual environments and versions of OpenGHG in use whilst populating their new store / new shared stores are populated.

This also means they can have new and old stores in their config file whilst the changeover happens.

### DIFF
--- a/openghg/objectstore/_local_store.py
+++ b/openghg/objectstore/_local_store.py
@@ -82,7 +82,7 @@ def get_writable_bucket(name: Optional[str] = None) -> str:
     if not writable_buckets:
         raise ObjectStoreError("No writable object stores found. Check configuration file.")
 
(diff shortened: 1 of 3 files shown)


SutarPrasad (commented):
Just a question? Do we want it to still allow the record processing if the user has an old store? For now, it throws the warning and then processes the record in the presence of the old store.

SutarPrasad (approved):
Looks good working as expected on local.

PR #965, opened 2024-03-21
Iss568 update edgar transform

### LINKED ISSUE #568: Make sure transform functions can work with EDGAR v7.0 and v8.0 data
### What is your issue?

Releases of [EDGAR v7.0](https://edgar.jrc.ec.europa.eu/dataset_ghg70) (now archived) and [EDGAR v8.0](https://edgar.jrc.ec.europa.eu/dataset_ghg80) are now available. This needs to be tested against our current `parse_edgar` function within the `openghg.transform.emissions` sub-module. (Linked through `openghg.store.Emissions.transform_data()`.

SutarPrasad:

- [x]       This PR covers updates to the transform function to process v8.0 edgar files.
- [x]       These include handling changes related to data variables renamed from emi_{species_label}(until v7) to fluxes(v8)
- [x]       Raising appropriate error if desired data var name is not present in the files
- [x]       Added relevant tests for these changes

- [x] Closes #568  (Replace xxxx with the Github issue number)
- [ ] ~~Documentation and tutorials updated/added~~


openghg/transform/flux/_edgar.py
```diff
@@ -1,21 +1,162 @@
...
+    timestamp_now,
+)
+
+
 logger = logging.getLogger("openghg.transform.flux")
 logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
 
+
 ArrayType = Optional[Union[ndarray, xr.DataArray]]
+Path = Union[pathlib.Path, zipfile.Path]
```
brendan-m-murphy, line 74: I know I made this this type alias, but it is only used twice and I think it's kind of confusing since we usually import `Path` from `pathlib`. I marked one of the changes that needs to happen if this is removed. The other is on line 539, where `Union[Path, str]` needs to change to `Union[pathlib.Path, zipfile.Path, str]`.
```suggestion
```

openghg/transform/flux/_edgar.py
```diff
@@ -1,21 +1,162 @@
...
+
+_edgar_known_versions = ("v432", "v50", "v60", "v70", "v80")
+
+
+# TODO: make this work for...
+# - yearly - "v6.0_CH4_2015_TOTALS.0.1x0.1.nc"
+# - sectoral - "v6.0_CH4_2015_ENE.0.1x0.1.nc"
+# - monthly sectoral - "v6.0_CH4_2015_1_ENE.0.1x0.1.nc", "v6.0_CH4_2015_2_ENE.0.1x0.1.nc", ...
+def assemble_edgar_metadata(
+    filepath: Path,
```
brendan-m-murphy, line 85: This changes goes with removing the type alias
```suggestion
    filepath: Union[pathlib.Path, zipfile.Path],
```

openghg/transform/flux/_edgar.py
```diff
@@ -301,21 +331,20 @@ def parse_edgar(
...
-    # file clean up and remove temporary directory and file.
-    if zipped:
-        temp_extract_folder.cleanup()
-
     # Check for "time" dimension and add if missing.
     flux_ndim = flux_values.ndim
     time_name = "time"
     if time_name in flux_da:
         time = flux_da[time_name].values
+        flux = flux_values  # TODO: this was missing... is this correct?? otherwise 'flux' might be undefined
```
brendan-m-murphy, line 339: @rt17603 can you confirm that this addition is correct? Previously `flux` wasn't defined if `time_name in flux_da` is True, and `flux` is needed later. (I'm guessing this means that we haven't seen an EDGAR file with a `time` variable yet.)

openghg/transform/flux/_edgar.py
```diff
@@ -587,95 +565,69 @@ def _extract_file_info(edgar_file: Union[Path, str]) -> Dict:
...
-    # e.g. "v432_CH4_2010_9_IPCC_6A_6D.0.1x0.1.nc" --> "IPCC_6A_6D.0.1x0.1"
-    try:
-        source_resolution = "-".join(filename_split[index_remaining:])
-    except (IndexError, ValueError):
-        raise ValueError(f"Unable to extract source: {filename}")
+    sector = file_info.pop("sector") or ""
+    sector = sector.replace("_", "-")
+
+    source = "_".join([co2_options, sector])
+    source = source.strip("_")  # in case co2_options or sector is ""
```
brendan-m-murphy, line 611: @rt17603 this results in slightly different source names than the old code in the case where there is a CO2 specific string like "org_short-cycles_C" but no sector string like "TOTALS". The old code would make the source "org_short-cycles_" and the new code would make it "org_short-cycles". Is this okay>

tests/transform/flux/test_edgar.py
```diff
@@ -257,6 +280,28 @@ def test_parse_edgar_no_domain():
                 "resolution": "0.1x0.1",
             },
         ),
+        (
```
brendan-m-murphy, line 283: I would add a few options with more complicated species and source names, e.g. 
`("v7.0_FT2021_N2O_2021_TNR_Aviation_CDS.0.1x0.1.zip", {"version": "v7.0", "species": "NO2", "year": 2021, "source": "TNR-Aviation-CDS", resolution: "0.1x0.1"})` and `("v8.0_FT2022_GHG_HFC-43-10-mee_1994_TOTALS_flx.nc", {"version": "v8.0", "species": "HFC-43-10-mee", "year": 1994, "source": "TOTALS", resolution: None})`

brendan-m-murphy (changes requested):
Looks good! I suggested some extra test cases, as well as removing the type alias `Path`, since it is kind of confusing (I know I added it, but I'm changing my mind). I also asked Rachel to look at a few changes.

openghg/transform/flux/_edgar.py
brendan-m-murphy, line 339: @qq23840 added some code to my previous draft EDGAR parser PR to fill NaNs in the flux data with zeroes. That code used a different library to regrid the data, so I'm not sure if it is necessary here. But we should check if regridding to the East Asia domain creates NaNs.

openghg/transform/flux/_edgar.py
rt17603: Yes that seems fine to me.

openghg/transform/flux/_edgar.py
```diff
@@ -63,210 +204,96 @@ def parse_edgar(
...
-
-    # TODO: May want to split out into a separate function, so we can use this for
-    # - yearly - "v6.0_CH4_2015_TOTALS.0.1x0.1.nc"
-    # - sectoral - "v6.0_CH4_2015_ENE.0.1x0.1.nc"
-    # - monthly sectoral - "v6.0_CH4_2015_1_ENE.0.1x0.1.nc", "v6.0_CH4_2015_2_ENE.0.1x0.1.nc", ...
-
-    if isinstance(date, int):
-        date = str(date)
+        # check if _readme.html is in folder_filelist, and if so, try to extract version
+        if (readme := datapath / "_readme.html").exists():
```
rt17603, line 229: If I was being a bit nitpicky - I think we should try and avoid more complex syntax like the walrus operator, `:=`, in favour of being a bit more verbose for clarity. I understand it can make syntax a bit neater but I just think it can make code a bit more confusing sometimes (and so can make it a bit harder to contribute).

Also Line 280 and 596.

openghg/transform/flux/_edgar.py
rt17603: This seems fine to me (looking at the file history this code may have evolved and so this omission was missed).

openghg/transform/flux/_edgar.py
```diff
@@ -63,210 +204,96 @@ def parse_edgar(
...
+
+    # get dataset
+    # using .open("rb") for pathlib.Path and zipfile.Path compatibility
+    edgar_ds = xr.open_dataset(edgar_file.open("rb"))
+
+    # Expected name e.g. "emi_ch4", "emi_co2" for version <= 7; "fluxes" for version 8
+    name = "fluxes" if version.startswith("v8") else f"emi_{species_label}"
+
+    # just in case we didn't get the name right...
+    if name not in (dvs := edgar_ds.data_vars):
```
brendan-m-murphy, line 280: Related to Rachel's suggestion above, I think this can just be
```suggestion
    if name not in edgar_ds.data_vars:
        if version.startswith("v8"):
            raise ValueError(...)
```

openghg/transform/flux/_edgar.py
```diff
@@ -63,210 +204,96 @@ def parse_edgar(
...
+    version = edgar_file_info["version"]
+
+    # get dataset
+    # using .open("rb") for pathlib.Path and zipfile.Path compatibility
+    edgar_ds = xr.open_dataset(edgar_file.open("rb"))
+
+    # Expected name e.g. "emi_ch4", "emi_co2" for version <= 7; "fluxes" for version 8
+    name = "fluxes" if version.startswith("v8") else f"emi_{species_label}"
+
+    # just in case we didn't get the name right...
```
brendan-m-murphy, line 279: I think this needs to change a bit: if `name` is not in the data vars, but the version doesn't start with `v8`, then no error is raised, but line 290 will fail. I think the `if` on line 281 needs to be followed by an `else` that assigns `name` to something (e.g. `edgar_ds.data_vars[0]`) or raises an error, but this time without the "flx_nc" vs. "emi_nc" part of the message. (Also, on 281, we don't need to check if `name not in dvs`, so we can remove the "warlus" statement that defines `dvs` and just make line 280 `if name not in edgar_ds.data_vars:`)

SutarPrasad: done

openghg/transform/flux/_edgar.py
SutarPrasad: done

openghg/transform/flux/_edgar.py
```diff
@@ -277,11 +277,13 @@ def parse_edgar(
     name = "fluxes" if version.startswith("v8") else f"emi_{species_label}"
 
     # just in case we didn't get the name right...
```
brendan-m-murphy, line 279:
```suggestion
    # check that data variable `name` is in `edgar_ds`
```

brendan-m-murphy (commented):
All the latest updates look good to me. We just need to decide what to do about the longitude conversion issue.

tests/util/test_domain.py
```diff
@@ -54,10 +54,25 @@ def test_create_coord():
...
+        (np.array([360.0]), np.array([0.0])),
+        (np.array([181.0]), np.array([181.0])),
+        (np.array([-180.0, 0.0, 179.9]), np.array([180.0, 0.0, 179.9])),
+        (np.arange(1, 361, 1), np.arange(1, 361, 1)),
+    ],
+)
+def test_convert_longitude_scale(lon_in, expected_lon_out):
+    # Your conversion logic here
+    lon_out = convert_lon_to_360(lon_in)
+    assert(lon_out, expected_lon_out)
```
brendan-m-murphy, line 73:
```suggestion
    np.testing.assert_allclose(lon_out, expected_lon_out)
```
I think `assert(x, y)` will just return true if both `x` and `y` evaluate to true?

openghg/util/_domain.py
```diff
@@ -150,12 +150,30 @@ def convert_longitude(longitude: ArrayLikeMatch) -> ArrayLikeMatch:
...
+    This does *not* reorder the values.
+
+    Args:
+        longitude: Valid longitude values in degrees.
+    Returns:
+        ndarray / DataArray : Updated longitude values in the same order.
+    """
+
+    # Check range of longitude values and convert to 0 to 360.
+    longitude = (longitude + 360) % 360
```
brendan-m-murphy, line 172:
```suggestion
    longitude = longitude % 360
```
Since 360 % 360 is 0, we can leave that part out.

brendan-m-murphy (changes requested):
Looks good! Just a few minor changes

tests/util/test_domain.py
SutarPrasad: done

brendan-m-murphy (approved):
Looks good!

PR #972, opened 2024-04-12
Store original version of file to be standardised

### LINKED ISSUE #296: Store a compressed version of the original file
Store this and then the Datasources can remember the name of the original file they were created from.

gareth-j:
* **Summary of changes**

This adds the functionality to store the original file in the object store.

- [ ] Closes #296 

### DIFF
--- a/openghg/objectstore/_local_store.py
+++ b/openghg/objectstore/_local_store.py
@@ -6,7 +6,7 @@ import os
 import threading
 from pathlib import Path
 import shutil
-from typing import Dict, List, Literal, Optional, Union
+from typing import Dict, List, Optional, Union
 import logging
 from openghg.types import ObjectStoreError
 from openghg.util import read_local_config
@@ -346,14 +346,16 @@ def set_compressed_file(bucket: str, key: str, filepath: Path) -> None:
     Returns:
         None
     """
-    filename = f"{bucket}/{key}._data.gz"
+    filename = Path(f"{bucket}/{key}._data.gz")
+    filename.parent.mkdir(exist_ok=True)
 
     # We shouldn't hit this but this but it might help us catch some logic errors
     if exists(bucket=bucket, key=key):
         raise ObjectStoreError("A compressed version of this file already exists.")
 
-    with open(filepath, "rb") as f_in, gzip.open(filename, "wb") as f_out:
-        shutil.copyfileobj(f_in, f_out)
+    with open(filepath, "rb") as f_in:
+        with gzip.open(filename, "wb") as f_out:
+            shutil.copyfileobj(f_in, f_out)
 
 
 def get_compressed_file(bucket: str, key: str, out_filepath: Path) -> None:
--- a/openghg/objectstore/_local_store.py
+++ b/openghg/objectstore/_local_store.py
@@ -1,11 +1,12 @@
 from __future__ import annotations
 import glob
+import gzip
 import json
 import os
 import threading
 from pathlib import Path
 import shutil
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Literal, Optional, Union
 import logging
 from openghg.types import ObjectStoreError
 from openghg.util import read_local_config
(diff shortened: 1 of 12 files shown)


openghg/objectstore/_local_store.py
```diff
@@ -100,11 +101,12 @@ def get_writable_bucket(name: Optional[str] = None) -> str:
 
 def get_tutorial_store_path() -> Path:
     """Get the path to the local tutorial store
+    in the user's home directory.
 
     Returns:
         pathlib.Path: Path of tutorial store
     """
-    return get_user_objectstore_path() / "tutorial_store"
+    return Path.home() / "openghg_tutorial_store"
```
rt17603, line 109: Was this related to a different issue?

openghg/store/base/_base.py
```diff
@@ -243,12 +308,13 @@ def assign_data(
                 datasource.add_data(
                     metadata=meta_copy,
                     data=dataset,
+                    data_type=data_type,
+                    file_hashes=file_hashes,
```
rt17603, line 312: Just thinking about the places where `file_hashes` has had an empty dict passed (e.g. in _flux.py with the transform function) - would it be better to make this optional rather than required and then have to pass a dummy input?

openghg/store/base/_datasource.py
```diff
@@ -364,10 +372,37 @@ def add_timed_data(
...
+
+        if "original_file_hashes" not in self._metadata:
+            self._metadata["original_file_hashes"] = {version: filename_only}
+        else:
+            if version in self._metadata["original_file_hashes"]:
+                self._metadata["original_file_hashes"][version].update(filename_only)
+            else:
+                self._metadata["original_file_hashes"][version] = filename_only
+
     def delete_all_data(self) -> None:
```
rt17603, line 402: When we delete data would we also want to delete the org files and references to the file hashes? I'm just thinking that when we've tried deleting data and then adding again, we seem to be retaining a memory of that data at the moment.

tests/store/test_datasource.py
```diff
@@ -550,13 +616,38 @@ def test_add_data_with_overlap_check_stored_dataset(bucket, datasets_with_overla
...
-    d.add_data(metadata=attributes, data=data_a, data_type="surface", new_version=False, if_exists="combine")
-    d.add_data(metadata=attributes, data=data_b, data_type="surface", new_version=False, if_exists="combine")
-    d.add_data(metadata=attributes, data=data_c, data_type="surface", new_version=False, if_exists="combine")
+    d.add_data(
+        metadata=attributes,
+        data=data_a,
+        file_hashes=test_hash_a,
+        data_type="surface",
+        new_version=False,
+        if_exists="combine",
```
rt17603, line 633: This may be unrelated to this PR but at the moment the `if_exists="combine"` mode isn't doing anything so any ideas what this is doing here?

rt17603 (commented):
Looks great overall - a few questions to consider (and one typo).

openghg/objectstore/_local_store.py
gareth-j: Yeah I was having issues with a store within a store. I'll change this back and I've opened an issue - #998

openghg/store/_flux.py
```diff
@@ -321,6 +323,9 @@ def transform_data(
         # Find all parameters that can be accepted by parse function
         all_param = list(inspect.signature(parser_fn).parameters.keys())
 
+        # TODO - QUESTION - how to hash this data?
```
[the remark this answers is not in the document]
gareth-j: Done with a0c9e1d81c361d6204b3956b9b8d223d01aaeec2

openghg/store/base/_base.py
gareth-j: I've updated the other locations where an empty dictionary was being passed in. It was actually quite easy to get the data required so that worked out.

tests/store/test_datasource.py
gareth-j: These are marked xfail, I just updated the tests so they could be used when combine is valid again.

openghg/store/base/_datasource.py
gareth-j: Oops! The file hashes and removed from the storage class and the original files are deleted now, added with 8ae5852589e840fb97beec129b48a3c3be6cbb04.

rt17603: @brendan-m-murphy raised the valid point that hashes may be related more than to one datasource so I think it's worth considering what happens if:
1. Datasource and associated hash is removed but hash is related to multiple datasources
2. Try to add back data using original file but this won't add because it finds *some* of the data already present
This can be a separate issue and we could put this in and then deal with that later if needed but it is something we will need to work out how to deal with at some point.

gareth-j: Yeah good point, I've opened https://github.com/openghg/openghg/issues/1005 for this.

brendan-m-murphy (changes requested):
I think we need a way for users to find out what datasources are associated with a file, because currently this information isn't available anywhere, except in the data manager warning message when deleting one datasource that came from a file associated to multiple UUIDs. There is code in `BaseStore.check_hashes` that comes close to doing this, so maybe we can start there.

rt17603:
From discussion in meeting - to address issue of one file hash = multiple uuids even when only asking to delete 1 uuid, would suggest we should actually delete all uuids created by the same file (otherwise couldn't re-add the data)
 - If this operation is attempted
    - raise Error, state that this would delete multiple uuids. If this is what user wants print the name of a flag to pass to say this is ok.
    - if flag passed, delete all uuids associated with that hash

rt17603:
Other discussion point:
 - See if we can separate code to collect together the uuids associated with a hash into its own function as could be useful in other places.

Issue #654, opened 2023-05-23
Decide how to handle data for the same base keyword combination (e.g. site/species/height) from different sources (e.g. NRT, ObsPack)


rt17603:
### What is your issue?

We need to decide how to further sub-categorise data associated with the same set of nominal keywords from different sources.

For example: we will have data from Mace Head (same species and inlet) which may be continuous Near Real time, from an ObsPack, from the ICOS Carbon Portal.

Suggestions include:
 - ~~a `source` field~~ a `dataset_source` field - could be relatively free form but with suggested consistency for known values such as "noaa_obspack", "icos_portal", "raw_data" (keeping lower case for now for consistency with other metadata)
     - *From feedback - the `dataset_source` field is already included as an option within the `retrieve_atmospheric` input. We can determine how this can be folded back throughout in a more generic way for other source formats and `standardise_surface` input keywords.*
 - a `label` field - potentially completely free-form (could be more verbose) e.g. "noaa obspack L2 data"

Note that some of this has already been discussed and covered regarding data level (L1, L2 classification) #551  and a `data_level` (and `data_sub_level`) keyword will also be made available independently of this.

Also c.f. discussion on internal versioning which tracks the history of a specific DataSource #597

joe-pitt:
I'm a bit confused as to how these will be different from the data_source and dataset_source attributes that we already have?

rt17603:
For reference `data_source` at the moment links to where the data was retrieved from and is only used for the `retrieve_atmospheric` via `local_retrieve` and is set to `"icoscp"`.

rt17603:
From discussion on 05/06/2023:

- Can make sure `dataset_source` is available as a general keyword when adding new data (`standardise_surface` at least). In addition to current values of "InGOS" or "European ObsPack" this could include e.g. "CEDA 2023.06" in relation to CEDA versioning.
    - This keyword can be optional and so would not always need to be included. In the case where new data is added that doesn't come from multiple dataset sources, this can be omitted (if new data were to be added later which *does* need a distinction, it would be possible to update the metadata on the previous data).
- `data_source` links to where the data was retrieved from and is also useful but could be renamed to `retrieved_from` e.g. "icoscp", "ceda". For internal processing we could either include "internal" or just omit this.

Other points:
- Suggest that if identifying metadata can be included which distinguishes between different data sets this is used (in preference to relying on versioning)
    - If a priority needs to be included for the data, the Ranking can be included for this
[...]

Issue #551, opened 2023-02-02
Distinguish between "levels" of data release


rt17603:
### Is your feature request related to a problem?

At the moment, in addition to versions of a data file, there can be several functional levels of data depending on the stages in the processing.

This could include:
 - "raw" data with no calibration applied (NRT?)
 - partially finalised data but not fully signed off (or would this be closer to NRT?)
 - public access data from e.g. CEDA repository (QA/QC?)

This data will need to be distinguished and sometimes compared.

c.f. ICOS data level

### Describe the solution you'd like

Some way to distinguish between these different "data levels" using the metadata e.g. add `data_level` key or otherwise to observation metadata.

Features needed:
- This will need to be decided in some way in advance (e.g. 2 = public facing data, 1 = local processing)
- Could allow for flexibility (e.g. 1.1, 1.2, 1.3 etc. or could this be covered by existing versioning?)
- Can be related to where the data originates OR an input e.g. 2 = from external resource / matches to data level on that platform.
- Can tie into ranking where data level could be used as a distinguisher to specify a preference between "degenerate" data

joe-pitt:
I like the idea of flexibility. It probably makes sense to keep the "L1" and "L2" definitions the same as ICOS. So having the ability to specify sublevels (e.g. L1.1, L1.2) could help with this? Then all L2 data could match a publicly-accessible data repository (and specify the source in the metadata?).

rt17603:
@joe-pitt I suppose the main benefit of the "sub-levels" would be to allow easy comparison between pre-finalised versions of the data? Is that the use case you would have in mind for this?

joe-pitt:
Yes - to allow comparison, and also to allow the NRT data to be L1, the final archived data to be L2, and the partially-finalised data to have some level in between. From a pure openghg perspective, it generally seems useful to allow for this flexibility in level naming (i.e. level numbers don't have to be integers). Then in the "acrg implementation" of openghg (i.e. the object store that we maintain on BP1) we can decide exactly how we want to define each sub-level (but in principle another group could maintain a local object store and define different sub-levels).

rt17603:
Great, that makes sense. We could either do this (A) as proposed above with non-integer levels in some way or (B) have an additional label (e.g. `data_sub_level`). Either way this could be an extra way to distinguish which can be set by the data owners.

Another question around this is if we would want a search for "L1" to return ALL data labelled with L1 (even if this has a sub-level) OR only the data labelled as "L1" with no sub-level.

[...]

PR #983, opened 2024-04-26
Metadata keys - 1: Read and writing required metadata keys

### LINKED ISSUE #978: Read `required_keys` from config file
As part of the move towards more flexible metadata handling we should be able to read the keys we use to create unique Datasources from a configuration file.

1. Store the default required_keys in a TOML file that's created in the config folder in the object store root
2. Allow users to create a new file with the updated keys
3. When we check the required keys we'll read either the default or the other file in the folder, raising an error if there's more than 2 files

We could have a simple folder structure in the object store root directory

```
config/
    db/                             <- can contain database specific config files / keys etc
    identifying_keys/      <- contain the files we use to create unique Datasources
    version.json              <- version of OpenGHG this config setup was created with
```

I don't think we should have a command line interface to set this up as it then makes it too easy for the file to be modified accidentally.

### Round 2

A slightly cleaner option

```
config/
    db/                             <- can contain database specific config files / keys etc
        config.toml            <- database config file, currently empty
        metadata_keys.toml     <- read the required keys from here
    version.json              <- version of OpenGHG this config setup was created with
```

Folder structure created by OpenGHG on first run. The keys config file will then be read on standardisation.

gareth-j:
* **Summary of changes** 

This adds the functionality to read and write required metadata keys for each storage class type in the object store.

- [ ] Closes #978


doc/source/tutorials/local/Managing_object_store/Setting_metadata_keys.rst
```diff
@@ -0,0 +1,151 @@
...
+
+    .. code-block:: python
+
+    
+    result = standardise_footprint(filepath=fp_filepath,
+                                    site="TAC",
+                                    domain="EUROPE",
+                                    model="NAME",
+                                    height="100m",
+                                    met_model"UKV",
```
SutarPrasad, line 140: met_model="UKV",

doc/source/tutorials/local/Managing_object_store/Setting_metadata_keys.rst
```diff
@@ -0,0 +1,151 @@
...
+                                    optional_metadata={"project": "project_a"})
+
+    result = standardise_footprint(filepath=fp_filepath_b,
+                                    site="TAC",
+                                    domain="EUROPE",
+                                    model="NAME",
+                                    height="100m",
+                                    met_model"UKV",
```
SutarPrasad, line 148: met_model="UKV",

gareth-j:
Tidy the defaults file to either be an easier to read TOML file or move to JSON.

doc/source/tutorials/local/Managing_object_store/Setting_metadata_keys.rst
gareth-j: Fixed

doc/source/tutorials/local/Managing_object_store/Setting_metadata_keys.rst
gareth-j: Fixed

openghg/store/base/_base.py
```diff
@@ -160,6 +160,42 @@ def check_hashes(self, filepaths: multiPathType, force: bool) -> Tuple[Dict[str,
...
+        if optional_metadata is not None:
+            common_keys = set(required) & set(optional_metadata)
+
+            if common_keys:
+                raise ValueError(
+                    f"The following optional metadata keys are already present in required keys: {', '.join(common_keys)}"
+                )
+
+            if optional:
+                for key in optional_metadata:
```
brendan-m-murphy, line 193: We could use kwargs here instead of `optional_metadata`?

brendan-m-murphy (commented):
My main question is around how we pass in the optional keys. `optional_metadata` makes sense for this, but I thought we were using that to avoid confusing "distinguishing" keys from "informational" keys (and so "optional" distinguishing keys would be passed as kwargs).

openghg/data/config/objectstore/defaults.json
```diff
@@ -0,0 +1,222 @@
...
+            },
+            "icos_data_level": {
+                "type": [
+                    "str"
+                ]
+            },
+            "data_type": {
+                "type": [
+                    "str"
+                ]
```
rt17603, line 218: We should be adding the new discussed keywords for "surface" here as well for surface (required with a default) - from #551 and #654 .

rt17603:
Summary for updates needed:
- kwargs gets added at the standardise level
- kwargs gets used as additional required keys
  - with a check that this maps to what is in the required keys config file
  - raise Error if not matching
- check optional_metadata isn't in required keys
  - raise Error if spotted

Just need to check these updates have been applied.

openghg/store/base/_base.py
```diff
@@ -160,6 +160,49 @@ def check_hashes(self, filepaths: multiPathType, force: bool) -> Tuple[Dict[str,
...
+        """
+        try:
+            metakeys = get_metakeys(bucket=self._bucket)[self._data_type]
+        except KeyError:
+            raise ValueError(f"No metakeys for {self._data_type}, please update metakeys configuration file.")
+        return metakeys
+
+    # TODO - move this kind of metadata handling into centralised location
+    def _add_additional_metadata(
+        self, data: Dict, additional_kwargs: Dict, optional_metadata: Optional[Dict]
```
brendan-m-murphy, line 177:
```suggestion
        self, data: Dict, additional_kwargs: Dict, optional_metadata: Optional[Dict] = None
```

brendan-m-murphy (commented):
Some general questions:

1. How exactly are we defining the difference between required and optional "metakeys"? Will "required" keys always be in the standardise function signature as positional or keyword args? Are "required" keys not meant to be modified? Do we need to specify default values for "optional" metakeys?
2. Should "optional_metadata" be used for creating datasources?  Or should it be purely informational? 

The code here is good, I'm just not sure on exactly what we want to happen. Right now, `optional_metadata` and `kwargs` passed to `read_file` are combined with the other metadata passed by `standardise`, then the metadata for datasource lookup is extracted. So, either `optional_metadata` or `kwargs` can be used to pass extra "distinguishing keys", as well as "informational keys". So there isn't any feed back to the user about which keys are used for what, and the interface doesn't make a distinction.

rt17603:
From discussion in Issues meeting:
2024-06-06_Metadata_keys_interpretation

Paraphrased summary of suggestions for workflow:
 - When combining metadata from different sources
    - Combine metadata for required and optional keys (including from **kwargs) - these will be used as the distinguishing keys to create the datasource
       - Check this against required keys in config file - error is they don't match
       - Check this against optional keys in config file - suggestion for now: raise warning if there are unrecognised keys but allow
   - Use these keys to lookup the datasource
   - Add informational metadata afterwards - includes informational metadata from the function signature and anything passed using `optional_metadata` input.
 - (If needed) Add "informational" key options to config file

PR #993, opened 2024-05-10
Empty stores are now treated as zarr stores

brendan-m-murphy:

Previous if a store was empty, the code that checks if it is a zarr store raised an error because no data was found.

This update changes this to log an info item stating that the store is empty and we are treating it as a zarr store.

This fixes a bug that prevented a zarr store from being initialised.


rt17603 (approved):
Looks good!

rt17603 (approved):

PR #1004, opened 2024-05-15
Centralise assign attributes and simplify parse function

### LINKED ISSUE #986: Centralise final metadata update steps to the store level rather than within parse functions
### What is your issue?

For the `ObsSurface` storage class in particular, if we want to add new data keys these need to be specifically added to every `parse_*` function call which can be time consuming, inefficient and prone to error. To make this easier it would be useful to allow partially completed metadata to be returned by the `parse_` function, containing just the keys which have been extracted in some way from accessing the data. Any additional keys e.g. user specified, or defaults could be added at the `ObsSurface.read_file` level (or in some other centralised way).

One issue to consider is that within every(/most) `parse_*` functions at the moment all the metadata values are used because each function calls the `assign_attributes` function. This function uses the metadata to update the attribute values associated with the data and does a comparison between the attributes and the metadata to make sure this is consistent. One way to get around this would be to move this function (or some variant of it) to be centralised in some way.

Suggested steps:

- [x] Split out functionality which formats the variables from `assign_attributes`. From these details create a new library of options which can be used within the (surface) parse functions (and update parse functions to use this new workflow). - completed with #1102 
- [x] Complete the metadata by adding any additional keys after the `parse_*` function call (needs to be generically done as output from each `parse_*` function will have different metadata missing) - completed within #1111 
- [ ] Move `sync_surface_metadata` steps within updated `assign_attributes` function (or functionality within this) to be used on the `ObsSurface` class
   - This would be applicable to *every* parse function within [standardise/surface](https://github.com/openghg/openghg/tree/devel/openghg/standardise/surface)
- [ ] Simplify the inputs to parse to reduce this to only the necessary keys for updating the data (returns partial metadata) - would need to be done differently for different parse functions but a similiar principle should apply. - split out into #1036

SutarPrasad:
 
- [x] Removed assign_attributes() function from parse_functions.
- [x] Added assign_attributes() function to ObsSurface class.
- [x] Removed parsed_surface_metachecker() from parse function tests.
- [x] Added parsed_surface_metachecker() to test_obssurface.py (More tests are required and in the process to identify its execution.
- [x] Changed the way parsed_surface_metachecker() function reads the input data.
- [x] Removed attribute checking call from parse and added to test_obssurface.py.
- [ ] Remove unwanted arguments from parse_* function signatures.

- [ ] Closes #986 (Replace xxxx with the Github issue number)


SutarPrasad:
https://github.com/openghg/openghg/issues/1022 current PR resolves this bug

rt17603:
Changes look good so far for this.

As discussed, this will be impacted by PR #983 and the changes will need to be integrated for this including the data keywords. 

Suggestions would be to:
 - When integrating in the new keywords from PR #983 don't add these to the parse_* functions but start to add the ability for these to be integrated into the metadata (/ attributes) at the ObsSurface level.
 - Keep this PR related to centralised assign_attributes (as has been done) and defer additional simplification of the parse_* functions to a later PR.

PR #1032, opened 2024-06-05
Iss1031 missing footprint data

### LINKED ISSUE #1031: Standardising data with a loop causes missing data at random.
### What happened?

Footprints for several sites in the shared store are missing data at random. (For instance, the inert footprint at JFJ is missing 8 days of data in April, 2021.)

The original files are not missing data.

Missing data shouldn't be added when files are added to the zarr store.

_No response_

### Relevant log output

_No response_

The problem is intermittent. If you add a few months of a footprint repeatedly, sometimes there will be times with missing data. This seems to happen about 40% of the time.

### Environment

<details>

</details>

brendan-m-murphy:

Added "synchronizer" to `to_zarr` calls to fix race condition, which was causing missing data to appear when standardising multiple files from a loop.

- [x] Closes #1031


rt17603 (approved):
Looks great - thanks very much for debugging and fixing this.

PR #1041, opened 2024-06-27
Iss978/keys config part1

### LINKED ISSUE #978: Read `required_keys` from config file

brendan-m-murphy:

This adds the functionality to read and write required metadata keys for each storage class type in the object store.

Metadata is currently processed as follows:
-   categorising metadata is determined solely by the ("metakeys" in the) config file
-   all of `optional_metadata` is combined with the metadata output by the parser (`optional_metadata` overwrites what comes out of the parser)
-   any "optional metakeys" found in `optional_metadata` will be used for datasource lookup (so remainder of `optional_metadata` is "informational")
-   an error is raised by `BaseStore.get_lookup_keys` if any key in `optional_metadata` is a required metakey (as defined by the config)
-   For flux and flux timeseries, the keyword arguments from `standardise_...` that are optional *for the parser* will overwrite items with the same keys in `optional_metadata`

In summary: informational metadata and "optional" categorising metadata can both be passed through the `optional_metadata` argument, but `optional_metadata` cannot conflict with the "required metakeys" defined in the config. Categorising metadata is determined (solely) by the metakeys config file.

- [x] Closes #978


tests/store/test_footprints.py
```diff
@@ -687,50 +688,70 @@ def test_optional_metadata():
...
+    # TODO - implement this in a different way
+    default_keys = get_metakey_defaults()
+
+    default_keys["footprints"]["optional"] = ["project", "special_tag"]
+
+    with patch("openghg.store.base._base.get_metakeys", return_value=default_keys):
+        yield
+
+
+def test_standardise_footprints_different_datasources(mock_metakeys):
```
rt17603, line 720: To make our own life easier later on, we should probably mark this test in some way / add a comment that we expect this functionality to change - at this moment this explicitly uses `optional_metadata` to include new distinguishing keys which we'll want to move away from as we go forward.

rt17603 (commented):
This looks good to me for Part 1 of these changes - small requested tidy up in the config file to keep this consistent but otherwise seems ready to go.

openghg/data/config/objectstore/defaults.json
```diff
@@ -0,0 +1,258 @@
...
+                "type": [
+                    "str"
+                ]
+            },
+            "icos_data_level": {
+                "type": [
+                    "str"
+                ]
+            },
+            "data_type": {
```
rt17603, line 251: Just spotted this after submitting review - I don't think we should have `data_type` here as a required metadata key?

brendan-m-murphy: The "required keys" for obs surface currently on devel are:
``` 
required_keys = (
                "species",
                "site",
                "sampling_period",
                "station_long_name",
                "inlet",
                "instrument",
                "network",
                "source_format",
                "data_source",
                "icos_data_level",
                "data_type",
            )
```
We probably don't want data type in there, but I guess it's been there for a while.

tests/store/test_footprints.py
brendan-m-murphy: I wrote down the name of the tests, I'll try to add it to the notes in the umbrella issue. This test will probably fail or be found when `optional_metadata` is renamed though.

openghg/data/config/objectstore/defaults.json
rt17603: Ok yeah I see it -https://github.com/openghg/openghg/blob/devel/openghg/store/_obssurface.py#L380. Thanks for pointing that out.

For the purposes of keeping this a neutral change then we can leave it for now but would you mind opening an issue for this? Should be a quick fix to remove and I don't believe would have much impact given data types are separated anyway but should do in a separate PR if we're being strict about it.

openghg/data/config/objectstore/defaults.json
```diff
@@ -0,0 +1,258 @@
...
+                "type": [
+                    "str"
+                ]
+            },
+            "data_source": {
+                "type": [
+                    "str"
+                ]
+            },
+            "icos_data_level": {
```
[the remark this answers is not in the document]
brendan-m-murphy: Sorry, I forgot to say below that "icos_data_level" has also been in there for a while. There are a lot of "required" keys for obs surface, but the min number of required keys is 5, I believe, so most won't be used.

openghg/data/config/objectstore/defaults.json
brendan-m-murphy: Okay, I added an issue: https://github.com/openghg/openghg/issues/1045

openghg/data/config/objectstore/defaults.json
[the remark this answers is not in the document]
rt17603: Ok that's fine - I've added a comment to #551 (may need to create a separate issue) so we can keep track of this.

rt17603 (approved):
Along with the local tests to ensure this can work with a current object store, this looks ready to go in.

PR #1043, opened 2024-06-28
Pinned numpy to < 2.0

brendan-m-murphy:

Numpy had a major release whle this PR was
in progress. We need to make changes to be
compatible with numpy 2.0.

We also need backwards compatibility with numpy 1.26 for a while because pymc/pytensor hasn't upgraded to numpy 2.0 yet, and openghg_inversions depends on openghg and pymc.

We have also pinned netCDF4 to versions <=1.6.5, since they seem to be having problems moving to numpy 2.0 (and our tests were failing with netCDF4 version 1.7.1).

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -41,7 +41,6 @@ jobs:
           flake8 openghg/ --count --statistics
       - name: Install openghg
         run: |
-          pip install -r requirements.txt
           pip install -r requirements-dev.txt
           pip install .
       - name: Run mypy
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -41,6 +41,7 @@ jobs:
           flake8 openghg/ --count --statistics
       - name: Install openghg
         run: |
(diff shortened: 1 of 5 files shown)


.github/workflows/workflow.yaml
```diff
@@ -41,6 +41,7 @@ jobs:
           flake8 openghg/ --count --statistics
       - name: Install openghg
         run: |
+          pip install -r requirements.txt
```
rt17603, line 44: I think (as you pointed out) as these dependencies are essentially included within the setup.py we shouldn't need to add this line. The `pip install .` below should essentially cover including details from `requirements.txt`.

rt17603 (approved):
Looks good to me - I've added an approval so this can get merged but may be worth checking the comment I made as I'm not sure we need to add the requirements.txt line in the CI.

.github/workflows/workflow.yaml
brendan-m-murphy: Okay I took it out.

rt17603 (approved):

PR #1049, opened 2024-07-04
Alignment of spatial data 

### LINKED ISSUE #1046: Use openghg_defs to set lat/lon coordinates for data with a domain
### What is your issue?

To avoid alignment issues, it would be useful to set the lat/lon coordinates for data with a domain when the data is added to an object store.

qq23840:
Fixes issue where spatial data (e.g. footprints, flux files) with slightly different lat/lon coordinates would be added, causing inability to align later in the process. 

`standardise_footprint` and `standardise_flux` workflows now contains an `align_lat_lon` function (found in utils) which 
1. Checks if the domain passed matches a domain in openghg_defs, and ignores if otherwise
2. Checks if the lat/lon values are 'close' (i.e. within 5%) to the desired lat/lon values from openghg_defs, and raises an exception if not (since the wrong domain may have been specified)
3. As long as the number of lat/lon values matches, replaces those in the data being standardised with the 'correct' values from openghg_defs. 

NB: currently, the function is called in `openghg.standardise.footprint._acrg_org.py`, `openghg.standardise.flux._openghg.py`, and `openghg.standardise.flux._intem.py`. Ideally, it would be higher up the chain so that it doesn't need to be repeated in every parser function.

- [x] Closes #1046 and https://github.com/openghg/openghg_inversions/issues/147
- [ ] Added any new requirements to `requirements.txt` and `recipes/meta.yaml` NOT APPLICABLE


tests/store/test_footprints.py
```diff
@@ -727,7 +727,7 @@ def test_standardise_footprints_different_datasources(mock_metakeys):
     file2 = get_footprint_datapath("TAC-100magl_UKV_TEST_201608.nc")
 
     site = "TAC"
-    domain = "EUROPE"
+    domain = "EUROPE-test"
```
brendan-m-murphy, line 730: Elsewhere the domain `test_europe` is used. Does this need to be the same?

tests/db_integrity/test_data_updates.py
```diff
@@ -809,7 +809,7 @@ def test_metadata_update():
             "TAC-185magl_UKV_EUROPE_TEST_201405.nc",
             "TAC-185magl_UKV_co2_EUROPE_TEST_201405.nc",
             "TAC",
-            "EUROPE",
+            "EUROPE-NON-STANDARD",
```
brendan-m-murphy, line 812: Why does the domain need to change for this test? Is that so that alignment code is not called?

tests/standardise/test_standardise.py
```diff
@@ -309,6 +311,35 @@ def test_standardise_footprint():
     assert "tmb_europe_test_model_10m" in results
 
 
+def test_standardise_align_footprint():
```
brendan-m-murphy, line 314: Could you add a short docstring here describing what the test is doing?

qq23840:
Moved the function to the `store._footprints.Footprints` class (and corresponding place for flux) to make more generic.

brendan-m-murphy (commented):
The code itself looks great, I'm just a bit confused about the changes to the tests. Could you explain the changes to the domain names in the tests? I'm not sure where the best place to do this is; perhaps just an inline comment near the domain names (like "EUROPE-NON-STANDARD"), or maybe in the change log?

We might want to think about a README file for our test data so we can keep track of the idiosyncrasies of the test data.

tests/db_integrity/test_data_updates.py
rt17603: I can chip in on this from a discussion yesterday - really in the tests we should only be using "EUROPE" for the domain name when we mean the exact "EUROPE" domain. So if there are cases with cut down domains we should be giving them a different name really.

rt17603 (commented):
Looks really good to me - thanks for putting this in. Just a few small comments to address from me.

tests/store/test_footprints.py
qq23840: See Rachel's comment below

tests/standardise/test_standardise.py
qq23840: Done in a recent commit

openghg/util/_domain.py
```diff
@@ -272,3 +272,70 @@ def cut_data_extent(
...
+        coord: coordinate to check. Currently 'lat' or 'lon' only.
+
+    Returns:
+        xarray.DataArray / xarray.Dataset: data with coordinate ranges aligned to openghg domain.
+    """
+
+    known_domains = list(get_domain_info().keys())
+
+    if domain.upper() not in known_domains:
+        return data
```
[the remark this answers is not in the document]
qq23840: Good idea - I've added this now

tests/standardise/test_standardise.py
```diff
@@ -309,6 +311,35 @@ def test_standardise_footprint():
...
+        force=True,
+        overwrite=True,
+        store="user",
+    )
+
+    data = get_footprint(site=site, network=network, height=height, domain=domain, model=model)
+
+    true_lats, true_lons = find_domain(domain=domain)
+
+    assert np.allclose(data.data.lat.values, true_lats, rtol=0, atol=1e-15)
```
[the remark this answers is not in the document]
qq23840: I agree with the bit about them being equal. The xarray functions only take in Datasets/DataArrays etc. as inputs, but the true_lats come out as a np.array, so I will just put in the first change?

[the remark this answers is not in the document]
rt17603: There's a numpy equivalent function that can be used as well: https://numpy.org/doc/stable/reference/generated/numpy.testing.assert_equal.html

rt17603 (approved):
Looks great! Happy for this to be merged.

Issue #987, opened 2024-05-02
Overview task: Ensure satellite and EM27 data can be used for inversions


rt17603:
### Is your feature request related to a problem?

We need to be able to add and run inversions with column data and so need to be able to add these to the object store.

#### Satellite

We already have satellite data and should ensure we can able to run with the same inversion set up as previous.

#### Site-column - new

For the GEMMA project we need to be able to add EM27 data to an object store so this can be run using the [openghg_inversions](https://github.com/openghg/openghg_inversions/) code.

### Describe the solution you'd like

EM27 data is Fourier Transform Infrared Spectrometer (FTIR) data, which is column data. This is the same type of data as measured by satellites so this could be standardised using the current `ObsColumn` interface with some tweaks to keep this consistent.

We should also ensure any logic added here also still works with the satellites data, though we can split into a separate issue if needed.

**Suggested approach**:

 - Satellite (known workflow): #1226 
 - Site-column (new workflow): #1229

rt17603:
Suggested approach (may want to split out into separate issues 

- [ ] Make sure EM27 data can be standardised and added to the object store
   - Review an internal `ObsColumn` format (currently based on gosat)
   - Ensure parse framework is present for the `ObsColumn` data type
   - Create `parse_` function(s) appropriate for this
      - EM27 (is this the network name?)
      - TCCON (if needed) - Related to Issue #855 (though this would currently be coming in through a different interface if retrieved from CEDA)
 - [ ] Ensure logic within [`get_gosat` code](https://github.com/ACRG-Bristol/acrg/blob/develop/acrg/obs/read.py#L636) from acrg code is implemented in openghg (can be equivalent e.g. within [`get_obs_column` function](https://github.com/openghg/openghg/blob/devel/openghg/retrieve/_access.py#L469)
 - [ ] Ensure ModelScenario uses time matched footprints for column data* (see e.g. [acrg code - resample is avoided](https://github.com/ACRG-Bristol/acrg/blob/develop/acrg/name/name.py#L780))
   - * for satellite data NAME footprints are created to exactly match the time points, however for ground-based column data we should check how these are being generated and if the footprints are being generated in the same way.
  - [ ] Check `openghg_inversions` can run with this data

elenafillo:
[...]

PR #1050, opened 2024-07-04
added satellite processing steps to get_obs_column

elenafillo:
Added logic from acrg get_gosat code, implemented within the get_obs_function. get_obs_function now retrieves standardised satellite data from a store and processes it to obtain mf_prior_factor, mf_prior_upper_level_factor, mf and mf_repeatability, based on required parameter max_level. If passed max_level is above available max level, it reverts to available max level

- [x] Solves one of the checkboxes on #987 (Replace xxxx with the Github issue number)
- [x] [Tests added and passed] Updated test_get_obs_column test

### DIFF
--- a/openghg/retrieve/_access.py
+++ b/openghg/retrieve/_access.py
@@ -3,6 +3,9 @@ import logging
 from io import BytesIO
 from typing import Any, Dict, List, Optional, Union
 from openghg.types import SearchError
+import logging
+
+logger = logging.getLogger("openghg.retrieve.access")
 
 from openghg.dataobjects import (
     BoundaryConditionsData,
@@ -511,7 +514,7 @@ def get_obs_column(
     )
 
     if max_level > max(obs_data.data.lev.values) + 1:
-        print(
+        logger.warning(
             f"passed max level is above max level in data ({max(obs_data.data.lev.values)+1}). Defaulting to highest level"
         )
         max_level = max(obs_data.data.lev.values) + 1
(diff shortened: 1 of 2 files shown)


openghg/retrieve/_access.py
```diff
@@ -509,6 +510,69 @@ def get_obs_column(
...
+            obs_data.data = obs_data.data.drop_vars(dv)
+    for coord in drop_coords:
+        if coord in obs_data.data.coords:
+            obs_data.data = obs_data.data.drop_vars(coord)
+
+    obs_data.data = obs_data.data.sortby("time")
+
+    obs_data.data.attrs["max_level"] = max_level
+    if species.upper() == "CH4":
+        obs_data.data.mf.attrs["units"] = "1e-9"
```
rt17603, line 566: Ultimately we will want the "units" attribute included but we should probably find a way to extract this from the data rather than setting this. Probably worth commenting out for now (for CH4 and CO2) but could you open a quick issue for this?

tests/retrieve/test_access.py
```diff
@@ -307,14 +307,23 @@ def test_get_obs_surface_cloud(mocker, monkeypatch):
 
 
 def test_get_obs_column():
```
rt17603, line 309: Could we split this out into two tests? - keep `test_get_obs_column` with the necessary tweaks (e.g. `xch4` is not `mf`) but then have a second test called something like `test_get_obs_column_max_level` (wordy!) which just tests the use of `max_level`. The benefit of doing that is if e.g. just the max_level feature starts failing it would be easier to pinpoint the issue.

rt17603 (commented):
Looks great - really useful to have this in! I've added a few very small comments but after those are addressed, happy to approve.

rt17603 (approved):
Looks good!

PR #1051, opened 2024-07-04
Add data_level and data_sublevel keywords for ObsSurface

rt17603:

When standardising observation surface data, this adds `data_level` and `data_sublevel` keys as options to define the data.

Notes:
 - Adds these keys as required keys for data_type="surface"
     - Uses `"not_set"` as the default when they are not specified
 - data_level keyword is checked to ensure this is only "0", "1", "2", or "3" (also allows decimals)
 - This is a basic implementation of this which is explicit about which keys are (a) not being passed to the parse function and then (b) being added to the metadata afterwards. The next step would be to improve this to make it more dynamic (next step for this is #1069)

- [x] Addresses Issue #551

Not done:
- Documentation and tutorials updated/added
- Added any new requirements to `requirements.txt` and `recipes/meta.yaml`


openghg/util/_data_level.py
```diff
@@ -0,0 +1,58 @@
...
+        str: Formatted
+    """
+
+    msg_expected = "Expect: '0', '1', '2', '3' (or decimal to indicate sub-level)"
+
+    if data_level is not None:
+        data_level = str(data_level)
+        try:
+            number = float(data_level)
+        except (ValueError, TypeError):
```
brendan-m-murphy, line 48: Maybe it's worth using `is_number` from `util`, like 

``` 
if not is_number(data_level):
   raise ValueError(...)
number = float(data_level)
```

openghg/util/_strings.py
```diff
@@ -170,3 +170,28 @@ def extract_float(s: str) -> float:
         return float(m.group(0))
 
     raise ValueError(f"No float values found in '{s}'")
+
+
+def check_and_set_null_variable(param: Union[str, None], null_value: Optional[str] = None) -> str:
```
brendan-m-murphy, line 175: This is a nice additional. Eventually it should go with the rest of the metadata formatting stuff, but I'll just make a note of that.

openghg/store/base/_base.py
```diff
@@ -160,6 +160,24 @@ def check_hashes(self, filepaths: multiPathType, force: bool) -> Tuple[Dict[str,
 
         return seen, unseen
 
+    def add_metadata_keys(self, data: Dict, additional_metadata: dict) -> dict:
```
brendan-m-murphy, line 163: Maybe call this something else since we're updated the metadata dictionary, not just the keys. Maybe just `update_metadata`?

brendan-m-murphy (changes requested):
Looks good! Just a few small suggestions.

openghg/standardise/_standardise.py
```diff
@@ -90,6 +92,14 @@ def standardise_surface(
             extract this from the file.
         height: Alias for inlet.
         instrument: Instrument name
+        data_level: The level of quality control which has been applied to the data.
```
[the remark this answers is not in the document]
rt17603: I've toyed with with the idea of using Enumerate / Literal for these types of inputs but generally came to the conclusion that I didn't want to make the interface for complex for the user - fine if we could allow the user to pass a string still though. I'm going to resolve this for now, if that's ok, but something to come back to with the metadata formatting.

openghg/store/base/_base.py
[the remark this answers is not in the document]
rt17603: You're right we should be consistent - I included `Dict` for this as that was what the `data` input to `assign_data` was so I'm actually going to update it to:
```suggestion
    def add_metadata_keys(self, data: Dict, additional_metadata: Dict) -> Dict:
```

openghg/util/_data_level.py
```diff
@@ -0,0 +1,58 @@
...
+    Args:
+        data_level: Specified data level for data
+    Returns:
+        str: Formatted
+    """
+
+    msg_expected = "Expect: '0', '1', '2', '3' (or decimal to indicate sub-level)"
+
+    if data_level is not None:
+        data_level = str(data_level)
```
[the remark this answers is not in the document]
rt17603: It was actually for a silly corner case which does come up where even though our type hinting asks for a string it doesn't enforce it so sometimes other types do get passed - this was to account for the integer case.
```suggestion
        data_level = str(data_level)  # in case a int/float is passed
```

openghg/util/_strings.py
rt17603: Yeah agreed - I was struggling to decide where to put it so kept with strings for now but will be better to include somewhere central when we pull together formatting.

openghg/store/base/_base.py
rt17603: Should be done

openghg/util/_data_level.py
rt17603: Should be done

openghg/store/base/_base.py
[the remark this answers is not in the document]
brendan-m-murphy: Since python 3.9 "Dict" etc has been deprecated: https://peps.python.org/pep-0585/

I guess it doesn't really matter because at some point we'll probably want to use a code formatter to update all the type hints (e.g. https://pypi.org/project/pyupgrade/).

openghg/util/_data_level.py
```diff
@@ -0,0 +1,58 @@
...
+from openghg.types import MetadataFormatError
+
+__all__ = ["format_data_level"]
+
+logger = logging.getLogger("openghg.util")
+logger.setLevel(logging.INFO)  # Have to set level for logger as well as handler
+
+
+@overload
+def format_data_level(
```
brendan-m-murphy, line 12: You could add int and float to the accepted inputs, the output will be str in any case. (With corresponding change in the actual implementation.) Not a big deal, I will probably look for things like that in the metadata formatting updates.

brendan-m-murphy (approved):
Looks good! I still have suggestions (opinions?) but they're just about type hints so I don't think it's worth holding up the PR. (PS let me know if you need a re-approval if you commit one of the dict vs Dict suggestions.)

openghg/store/base/_base.py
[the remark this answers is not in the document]
rt17603: Ok - agreed this should be updated then in the longer term. For now I'll keep it as is to be consistent with the other uses of `data` but yes this should all get updated in one go really to replace `Dict` with `dict` where this is used.

PR #1060, opened 2024-07-12
Iss860 icos retrieval jp

joe-pitt:
Merged devel into iss860_icos_retrieval. I had to resolve two conflicts that need checking:

1. devel has the following code to exclude various things we don't want:
```
    # We want to get the PIDs of the data for each species here
    species_upper = [s.upper() for s in species]
    # For this see https://stackoverflow.com/a/55335207
    search_str = r"\b(?:{})\b".format("|".join(map(re.escape, species_upper)))
    # Now filter the dataframe so we can extract the PIDS
    # We filter out any data that contains "Obspack" or "csv" in the specLabel
    # Also filter out some drought files which cause trouble being read in
    # For some reason they have separate station record pages that contain "ATMO_"
    filtered_sources = data_pids[
        data_pids["specLabel"].str.contains(search_str)
        & ~data_pids["specLabel"].str.contains("Obspack")
        & ~data_pids["specLabel"].str.contains("csv")
        & ~data_pids["station"].str.contains("ATMO_")
    ]

    if filtered_sources.empty:
        species_lower = [s.lower() for s in species]
        # For this see https://stackoverflow.com/a/55335207
        search_str = r"\b(?:{})\b".format("|".join(map(re.escape, species_lower)))
        # Now filter the dataframe so we can extract the PIDS
        filtered_sources = data_pids[
            data_pids["specLabel"].str.contains(search_str)
            & ~data_pids["specLabel"].str.contains("Obspack")
            & ~data_pids["specLabel"].str.contains("csv")
        ]
```
The equivalent code in iss860_icos_retrieval is commented out, so I've left it that way, but I'm not totally sure if the changes in this part of the code on devel since iss860_icos_retrieval branched off are accounted for in iss860_icos_retrieval.

2. I stuck with the iss860_icos_retrieval changes to the icos tests (i.e. test_icos_retrieve_skips_obspack_globalview has extra code at the start, and is marked xfail)

I also edited the following in retrieve.py to account for the different timestamp format for flask data:
```
        try:
            dataframe = dataframe.rename(columns=rename_cols).set_index("timestamp")
            dataframe.index.name = "time"
            dataframe.index = to_datetime(dataframe.index, format="%Y-%m-%d %H:%M:%S")
        except KeyError:
            # flask data has sampling start instead of timestamp
            dataframe = dataframe.rename(columns=rename_cols).set_index("samplingstart")
            dataframe.index.name = "time"
            dataframe.index = to_datetime(dataframe.index, unit="ms")
```


joe-pitt:
I've done a quick fix using try except for now - not sure this is the best approach...

brendan-m-murphy (approved):
Looks good!

Issue #513, opened 2022-12-13
Improve memory usage of ModelScenario


gareth-j:
We're looking at improving the memory usage of `ModelScenario`.
We could try using zarr stores to cache the loaded data and the intermediate data created by some of the methods.

chryswoods:
This will be completed by the quick fix to reduce memory consumption in the code as it is now (@brendan-m-murphy). How this relates to the special release will be considered later.

gareth-j:
We need to sit down and look at this and profile the usage when using co2 data. Talk to Eric for input on this.

PR #1062, opened 2024-07-15
Model scenario speed improvements

brendan-m-murphy:

The main algorithmic change is to remove the loop in favour of vectorised code.

Another change is to use dask via xarray. This is implicit in the code since we load the data from zarr (as opposed to explicitly chunking the xarray DataArrays).

Some test runs computing modelled obs for high time resolution CO2 footprints/fluxes suggests:
- `footprints_data_merge` takes about 1/3 of the time to run
- `footprints_data_merge` uses 1/3 to 2/3 less memory

- [x] Partly addresses #513


tests/analyse/test_scenario.py
rt17603, line 1: This is very optional but I'm wondering if it would be useful to include a few tests for different flux / footprint frequency combinations with dummy data?

rt17603 (approved):
Looks good to me! I've included a few small comments and a question on whether this would benefit from a few more tests but none are necessary for this go in.

openghg/analyse/_scenario.py
```diff
@@ -1287,102 +1285,78 @@ def _calc_modelled_obs_HiTRes(
...
+            # set H_back coordinates using highest_res_H frequency
+            h_back_type = fp_high_time_res.H_back.dtype
+            flux_hf_rolling = flux_hf_rolling.assign_coords(
+                {"H_back": np.arange(0, max_h_back, highest_res_H, dtype=h_back_type)[::-1]}
+            )
 
-        # Extract footprints array to use in numba loop
-        fp_HiTRes = da.array(fp_HiTRes)
+            # select subsequence of H_back times to match high res fp (i.e. fp without max H_back coord)
+            flux_hf_rolling = flux_hf_rolling.sel(H_back=fp_high_time_res.H_back)
```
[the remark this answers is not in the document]
brendan-m-murphy: The re-ordering is probably just incidental. The coordinates need to be assigned in the reverse order on lines 1311-1312 because the rolling window goes backwards in time. The footprint `H_back` coordinate is in increasing order, so that re-orders the `H_back` coordinate of the rolling window.

The main purpose of this part of the code is to select of a subset of the times, like you say in part (b).

I didn't think to mention the reordering in the comments because with xarray you don't need to worry about that.  (But the reverse order on lines 1311-1312 is important.)

I guess this could be confusing if a subset of `H_back` doesn't need to be selected. Also I guess the comment below the signature is slightly inaccurate since the number of times points could change at line 1316.

The selection part could be moved nearer to the calculation on line 1343. (I guess another option could be to fill values in the footprint to match the flux, if the flux is higher frequency. I'm not sure if that is better or worse than dropping the extra flux values in this case.

openghg/analyse/_scenario.py
```diff
@@ -1287,102 +1285,78 @@ def _calc_modelled_obs_HiTRes(
...
+        )
 
         if output_TS:
-            timeseries = DataArray(timeseries, dims=("time"), coords={"time": time_array})
+            timeseries = fpXflux.sum(["lat", "lon"])
+
+        # TODO: Add details about units to output
 
         if output_fpXflux and output_TS:
             timeseries.compute()
```
brendan-m-murphy, line 1362: We might not want to compute the `fpXflux` here since it will be used for later calculations in inversions (once inversions switches to using this code to get `fpXflux`) and it would probably be best to delay computation as long as possible.

openghg/analyse/_scenario.py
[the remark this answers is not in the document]
rt17603: Thanks for clarifying and that all sounds fine - I think initially I just wondered on this because of the clause "(i.e. fp without max H_back coord)" but flux_hf_rolling already didn't include the max H_back. However, re-reading the comment I think I now understand the meaning better with "select subsequence" as the key part of the comment. Before it seemed like there was an expectation that the max H_back coord would be within flux_hf_rolling but I think I understand the "i.e." clause better now.

I don't think any changes are needed (I think it's just complex!).

openghg/analyse/_scenario.py
rt17603: Did you want to add an issue for this?

brendan-m-murphy: I don't think it's possible to get `fpXflux` without calling this "private" method directly right now. If we don't have an issue for that, we should make one, since this code gets duplicated in inversions for that reason. So... delaying compute could be done now, or I could mention it in that issue.

openghg/analyse/_scenario.py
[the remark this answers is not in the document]
brendan-m-murphy: Ah yea, the "i.e." was just a reminder of what "high res fp" is supposed to mean. That should be clearer with the new footprint format.

openghg/analyse/_scenario.py
brendan-m-murphy: No, `timeseries.compute()` will do that. For inversions, we don't actually need the timeseries, we need the `fpXflux` dataset. (Similarly for `bc_mod`, we actually need the dataset used to compute that.)

brendan-m-murphy: If you want both `timeseries` and `fpXflux` and don't mind using the memory to store `fpXflux`, then you could use `.compute` or `.persist` on `fpXflux`.

rt17603: I don't think there was a strong reason why `fpXflux` was computed (or `timeseries` really) at that point before - if we wanted more control over that we could allow for that to be flagged and by default not computed.

Also if it would be more useful to split this function out into more steps so it could be accessed in a more fine-grained way for inversions that would be fine too.

Issue #921, opened 2024-02-21
Add back `elevate_inlets` functionality


gareth-j:
Add ability to elevate inlets as per the original `recombine_datasets` function, see `elevate_duplicate_attrs`, this
can be moved closer to the user facing code where the Dataset is created.

rt17603:
For reference, this was originally included to allow footprints / observations with different inlet values to be concatenated but to still keep details about the original inlet value.

PR #1064, opened 2024-07-17
Search for range

brendan-m-murphy:

Adds the ability to search for an inlet based on a range of values.

To facilitate this, two new options were added to `TinyDBMetastore.search`:
1. searching by "test functions": find results where `f(record[k])` is true, where `record` is a dict from the metastore, `f` is the "test function", and `k` is a metadata key
2. searching by "negative lookup": if `key = NOT_SET_FLAG` is passed to search, then find results where `k` is not in `record`. (This isn't necessary for the range search, but we'll want this in the near future.)

- [x] Progress towards #921 


openghg/objectstore/metastore/_metastore.py
```diff
@@ -176,7 +181,45 @@ def _get_query(self, metadata: MetaData) -> tinydb.queries.QueryInstance:
         """
         return tinydb.Query().fragment(self._format_metadata(metadata))
 
-    def search(self, search_terms: Optional[MetaData] = None) -> QueryResults:
+    def _get_test_query(self, search_tests: dict[str, Callable]) -> tinydb.queries.QueryInstance:
```
SutarPrasad, line 184: `search_tests` replaced with `search_ranges`. For uniformity and code readability. Although the user won't see it.

openghg/objectstore/metastore/_metastore.py
```diff
@@ -176,7 +181,45 @@ def _get_query(self, metadata: MetaData) -> tinydb.queries.QueryInstance:
...
+
+    def search(
+        self,
+        search_terms: Optional[MetaData] = None,
+        search_tests: Optional[dict[str, Callable]] = None,
+        negative_lookup_keys: Optional[list[str]] = None,
+    ) -> QueryResults:
         """Search metastore using a dictionary of search terms.
 
         Args:
```
SutarPrasad, line 225: docstrings needs update

openghg/objectstore/metastore/_metastore.py
```diff
@@ -189,8 +232,17 @@ def search(self, search_terms: Optional[MetaData] = None) -> QueryResults:
...
+            _query = _query & _test_query
+
+        if negative_lookup_keys:
+            _neg_query = self._get_negative_lookup_query(negative_lookup_keys)
+            _query = _query & _neg_query
+
+        return list(self._db.search(_query))
 
     def insert(self, metadata: MetaData) -> None:
         """Add new metadata to the metastore.
```
SutarPrasad, line 248: docstrings update

openghg/objectstore/metastore/_metastore.py
```diff
@@ -256,5 +308,5 @@ def delete(self, metadata: MetaData, delete_one: bool = True) -> None:
                 be deleted.
```
SutarPrasad, line 308: same as above

SutarPrasad (changes requested):
Some formatting changes, apart from that looks good

openghg/objectstore/metastore/_metastore.py
rt17603: I agree perhaps this could have a more descriptive name but doesn't this input allow for more generic cases than just ranges though? In principle this could be any function so perhaps `search_by_function` (unless that seems too wordy)? The original `search_tests` is also fine by me though if we can't think of an alternative.

openghg/retrieve/_search_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+
+def parse_search_kwargs(search_kwargs: dict) -> list[dict]:
+    """
+    Process search kwargs into list flat dictionaries with the correct combinations of search queries.
+
+    To set this up for keywords with multiple options, lists of the (key, value) pair terms are created
+    - e.g. for species = ["ch4", "methane"], time_resolution = {"time_resolved": "true", "high_time_resolution: "true"}
+    - multiple_options is [[("species", "ch4"), ("species", "methane")], [("time_resolved": "true"), ("high_time_resolution": "true")]]
```
rt17603, line 106: This may lend itself to something in the doc-string and perhaps keep the explicit tuple grouping as a comment in the body of the code. Something perhaps like:
```suggestion
    - We expect this to create search options to look for:
       - "species" as "ch4" OR "methane" AND either "time_resolved" as "true" OR "high_time_resolution" as "true"
```

openghg/retrieve/_search_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    - we then expect searches for all permutations across both lists.
+
+    Args:
+        search_kwargs: dictionary of search terms
+
+    Returns:
+        list of flat dictionaries containing all combinations of search terms from (nested) input search terms
+    """
+    multiple_options = []
```
rt17603, line 115:
```suggestion
    # multiple_options will contain tuple pairs for the options we wish to search for e.g.
    # e.g. for species = ["ch4", "methane"], time_resolution = {"time_resolved": "true", "high_time_resolution: "true"}
    # - multiple_options is [[("species", "ch4"), ("species", "methane")], [("time_resolved": "true"), ("high_time_resolution": "true")]]
    multiple_options = []
```

openghg/retrieve/_search_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+            del _search_terms[k]
+
+    return {
+        "search_terms": _search_terms,
+        "search_tests": search_tests,
+        "negative_lookup_keys": negative_lookup_keys,
+    }
+
+
+def parse_search_kwargs(search_kwargs: dict) -> list[dict]:
```
rt17603, line 100: A small thought - should we reserve the term `parse` in our codebase for the standardise parsers? In which case could this be `format_search_terms` (or something else)?

openghg/retrieve/_search_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+        return x  # key != "inlet"
+
+    def test_func(x: Any) -> bool:
+        """Return True if start <= formatter(x) <= stop."""
+        return _in_interval(formatter(x), formatter(s.start), formatter(s.stop))  # type: ignore
+
+    return test_func
+
+
+def process_special_queries(search_terms: dict, neg_lookup_flag: Any = "NOT_SET_FLAG") -> dict:
```
rt17603, line 70: We could just set the default for `neg_lookup_flag` to be `np.nan` if that's what we decided. We've basically ended up with 2 situations where the metadata key has not been specified (1) where the key is present but the value is "not_set" and (2) where the key isn't present. The `neg_lookup_flag` covers case (2) so we may want to be a bit careful to not conflate the two cases.

rt17603 (commented):
Looks good to me - a few small comments but then this looks ready to go!

openghg/retrieve/_search_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    Args:
+        x: value to test for inclusion in the interval [start, stop]
+        start: start of the interval
+        stop: end of the interval
+
+    Returns:
+        True if start <= x <= stop, False otherwise
+    """
+    if start is None and stop is None:
+        return False
```
brendan-m-murphy, line 27: This case should return `True` to be consistent with other uses of `None` in this this function

openghg/retrieve/_search_helpers.py
brendan-m-murphy: I suppose we might want to be able to specify this somewhere (possibly in the config?). I can set it to np.nan for now. I just chose something arbitrary because I wasn't sure what we wanted (but didn't want to use `NOT_SET`).

openghg/retrieve/_search_helpers.py
```diff
@@ -0,0 +1,138 @@
...
+    return {
+        "search_terms": _search_terms,
+        "search_tests": search_tests,
+        "negative_lookup_keys": negative_lookup_keys,
+    }
+
+
+def parse_search_kwargs(search_kwargs: dict) -> list[dict]:
+    """
+    Process search kwargs into list flat dictionaries with the correct combinations of search queries.
```
brendan-m-murphy, line 102:
```suggestion
    Process search kwargs into list of flat dictionaries with the correct combinations of search queries.
```

openghg/retrieve/_search_helpers.py
rt17603: Yeah makes sense and I agree we should specify this somewhere. I think that's fine for now though.

openghg/objectstore/metastore/_metastore.py
brendan-m-murphy: I changed it to "search tests" because the that is the name of the method used by the TinyDB Query: `Query().test(test_function)`, but maybe `search_by_function` is better.

openghg/objectstore/metastore/_metastore.py
brendan-m-murphy: I just updated the local variable names here, so I'm not sure if the docstring needs to change. (Actually... I didn't intend to change `query` to `_query` everywhere... my editor's "rename variable" function changed more than I wanted it to...)

openghg/retrieve/_search_helpers.py
brendan-m-murphy: I changed it to `flatten_search_terms`. Maybe `process_search_terms` is better? I guess flattening and formatting both happen...

openghg/objectstore/metastore/_metastore.py
brendan-m-murphy: Done

SutarPrasad (approved):
Looks good

PR #1066, opened 2024-07-18
Iss921 reinstate elevate inlets

brendan-m-murphy:

### DIFF
--- a/openghg/retrieve/_access.py
+++ b/openghg/retrieve/_access.py
@@ -28,6 +28,7 @@ multDataTypes = Union[
 
 
 def _get_generic(
+    combine_multiple_inlets: bool = True,
     elevate_inlets: bool = False,
     ambig_check_params: Optional[list] = None,
     **kwargs: Any,
@@ -45,6 +46,10 @@ def _get_generic(
     """
     from openghg.retrieve import search
 
+    if elevate_inlets is True:
+        combine_multiple_inlets = True
+        logger.warning("`elevate_inlets` is deprecated, use `combine_multiple_inlets` instead.", DeprecationWarning)
+
     results = search(**kwargs)
 
     keyword_string = _create_keyword_string(**kwargs)
@@ -61,15 +66,18 @@ def _get_generic(
         logger.exception(err_msg)
         raise SearchError(err_msg)
     elif isinstance(retrieved_data, list) and len(retrieved_data) > 1:
-        param_diff_formatted = _metadata_difference_formatted(data=retrieved_data, params=ambig_check_params)
-        err_msg = f"""
-        Multiple entries found for input parameters for {keyword_string}.
-        Parameter differences:
-        {param_diff_formatted}
-        Please supply additional parameters or set ranking.
-        """
-        logger.exception(err_msg)
-        raise SearchError(err_msg)
+        if combine_multiple_inlets:
+            pass
+        else:
+            param_diff_formatted = _metadata_difference_formatted(data=retrieved_data, params=ambig_check_params)
+            err_msg = f"""
+            Multiple entries found for input parameters for {keyword_string}.
+            Parameter differences:
+            {param_diff_formatted}
+            Please supply additional parameters or set ranking.
+            """
+            logger.exception(err_msg)
+            raise SearchError(err_msg)
     elif isinstance(retrieved_data, list):
         retrieved_data = retrieved_data[0]
 
--- a/openghg/retrieve/_access.py
+++ b/openghg/retrieve/_access.py
@@ -3,7 +3,7 @@ import logging
 from io import BytesIO
 from typing import Any, Dict, List, Optional, Union
 
-from openghg.dataobjects._basedata import _BaseData
+from openghg.dataobjects._basedata import _BaseData  # TODO: expose this type?
 from openghg.dataobjects import (
     BoundaryConditionsData,
     FluxData,
--- a/openghg/retrieve/_access.py
+++ b/openghg/retrieve/_access.py
@@ -288,6 +288,7 @@ def get_obs_surface_local(
     retrieved_data = _get_generic(
         sort=True,
         elevate_inlets=False,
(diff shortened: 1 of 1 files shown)


openghg/retrieve/_access.py
```diff
@@ -1,20 +1,18 @@
 import json
 import logging
 from io import BytesIO
-from typing import Any, Dict, List, Optional, Union
-from openghg.types import SearchError
-import logging
-
-logger = logging.getLogger("openghg.retrieve.access")
```
rt17603, line 8: Did this line get removed?

openghg/retrieve/_access.py
```diff
@@ -23,29 +21,33 @@
...
 multDataTypes = Union[
-    List[BoundaryConditionsData], List[FluxData], List[FootprintData], List[ObsColumnData], List[ObsData]
+    list[BoundaryConditionsData], list[FluxData], list[FootprintData], list[ObsColumnData], list[ObsData]
 ]
 
 
 def _get_generic(
-    sort: bool = True,
+    combine_multiple_inlets: bool = False,
     elevate_inlets: bool = False,
```
rt17603, line 30: I think we can probably just remove `elevate_inlets` entirely and replace with `combine_multiple_inlets` if that's easier. I wouldn't say anyone using the code or other code is relying on it at the moment.

openghg/retrieve/_access.py
```diff
@@ -23,29 +21,33 @@
...
         elevate_inlets: Elevate the inlet attribute to be a variable within the Dataset
         ambig_check_params: Parameters to check and print if result is ambiguous.
         kwargs: Additional search terms
     Returns:
         dataclass
     """
     from openghg.retrieve import search
-    from openghg.store.spec import define_data_type_classes
+
+    if elevate_inlets is True:
```
rt17603, line 46: If `elevate_inlets` gets removed and replaced with` combine_multiple_inlets` this warning can probably be removed.

openghg/retrieve/_access.py
```diff
@@ -234,7 +229,9 @@ def get_obs_surface_local(
         species: Species identifier e.g. ch4 for methane.
         start_date: Output start date in a format that Pandas can interpret
         end_date: Output end date in a format that Pandas can interpret
-        inlet: Inlet height above ground level in metres
+        inlet: Inlet height above ground level in metres; use `slice(lower, upper)` to
```
rt17603, line 232: Just to make it a bit more clear that this can still just be a single value suggesting a small change to this of:
```suggestion
        inlet: Inlet height above ground level in metres; This can be a single value or `slice(lower, upper)` can be used to
```

openghg/retrieve/_access.py
```diff
@@ -290,8 +290,8 @@ def get_obs_surface_local(
     # # Get the observation data
     # obs_results = search_surface(**surface_keywords)
     retrieved_data = _get_generic(
-        sort=True,
         elevate_inlets=False,
```
rt17603, line 293: Again, if removing `elevate_inlets` can remove from here as well.

openghg/retrieve/_access.py
```diff
@@ -773,10 +766,7 @@ def _scale_convert(data: Dataset, species: str, to_scale: str) -> Dataset:
 
     converter = scale_converter.loc[row]
 
-    if to_scale == converter["scale1"]:
-        direction = "2to1"
-    else:
-        direction = "1to2"
+    direction = "2to1" if to_scale == converter["scale1"] else "1to2"
```
rt17603, line 769: Not a big deal but I actually think the original version of this is more clear - I know its more lines but sometimes it's ok to be a bit more verbose for the sake of readability.

rt17603 (commented):
Looks great to me - thanks for putting this in! A few very small comments, the main one being that I think we can just remove `elevate_inlets` for now rather than worrying about including a DeprecationWarning.

openghg/retrieve/_access.py
brendan-m-murphy: I think ruff suggested this.. I was experimenting with it. It seems like it would be an upgrade to black and flake8. I'll keep note of the flags it uses for this kind of change. Anyway, happy to revert it.

rt17603: Ah ok makes sense - I'd prefer the former but may just be personal preference. Happy to leave this up to you.

openghg/retrieve/_access.py
brendan-m-murphy: Yes, there is a line below it `logger = logging.getLogger("openghg.retrieve")` that redefines `logger`. I think the one I deleted was added by Elena at the code retreat. I guess the way we structure the imports might make it confusing to log messages from `openghg.retrieve.access`, but I suppose there was a reason for adding the more specific logger. (It's possible I didn't realise they were different and thought they were just duplicates.)

openghg/retrieve/_access.py
brendan-m-murphy: Okay, I'll remove it.

openghg/retrieve/_access.py
brendan-m-murphy: Okay, I'll update the docstrings this way.

openghg/retrieve/_access.py
brendan-m-murphy: Done

openghg/retrieve/_access.py
brendan-m-murphy: Since the condition isn't too long, I think I prefer the one line version... it makes it clear that the main point is to define `direction`. Sometimes the one line version is unreadable though, so I'll avoid automatically changing these.

openghg/retrieve/_access.py
rt17603: Yes I see that in the unchanged bit now - I think leaving this as logging from `openghg.retrieve` is fine for now on this.

rt17603 (approved):
Looks good!

PR #1069, opened 2024-07-19
Update flexibility of collecting parameters to pass to parse function

### LINKED ISSUE #1071: Stop passing unecessary metadata to parse functions and add to metadata within `ObsSurface`
When using `standardise_surface` (and other data types) to add details to the metadata this has to be passed to the defined `parse_*` function. This could be improved by allowing extra metadata to be added at the store level after the parse step has occurred. PR #1051 started to implement this idea but did this specifically for `data_level` and `data_sublevel` keys. This issue is to aim to do this more generically based on the config file definitions.

Suggested plan:
 - Collect input keys and pass to parse_* function
    - Collect keys from read_file input (`locals()` function could be used for this) - `fn_input_parameters`
    - Check against parse function signature, pass matching + any additional keys defined internally this needs (e.g. filepath)
    - Remove keys passed to the parse function from collection of input keys `fn_input_parameters` to create `additional_input_parameters` to avoid trying to add the same keys to the metadata twice.
 - Define additional keys to add to the metadata which are not passed by the user and not created by the parse function - `additional_metadata`.
 - Input `optional_metadata` is checked to ensure this doesn't include required keys from config and then added to `additional_metadata`.
 - Check the metadata which has been returned from the parse function
    - Collect required keys for the data type from config. Does the metadata contain these required keys?
        - Add required keys from collected keys passed to read_file function (`input_parameters`) ; error if not possible
    - Check config for optional keys.
        - Check collected keys (`input_parameters`) and add optional keys to metadata (if present)
    - Add any remaining keys within `additional_metadata` to the `metadata`
    - If any key values are outside a tolerance between multiple sources of metadata - raise an error
       - rel_tolerance for numbers of 1e-5; case-insensitive matches for strings; ignore None; ignore "not_set" if other value is present.
 - Construct distinguishing keys
    - Required (config) + Optional (map metadata against config to find appropriate keys)
 - Use distinguishing keys for lookup

Caveat: Will need to make sure the key names for inputs to read_file match the key names for inputs to the parse functions. 
 - PR #1108 (data_filepath --> filepath) and #1109 (sampling_period_seconds --> sampling_period) addressed this where inconsistent for `ObsSurface`.

rt17603:

This PR is to add to flexibility around input parameters being added to the metadata to move away from needing this to be pre-defined. This also updates how keys from various sources are used when creating the overall metadata.

### Functional updates

Determines which keys have been passed into the `.read_file()` method and matches this against:
1. the parse function (where relevant)
2. the config keys associated with the data class to determine how to construct the metadata

The config keys are used again to determine the lookup keys from the collated metadata - this is used when checking for current data sources.

#### Metadata merging

This deals with multiple sources of metadata:
 - metadata returned from the parse functions / constructed within the data class
    - the values will vary per datasource being added but the keys should match between these
 - input parameters - live data passed into read_file function - when merging the metadata only the input parameters NOT passed to the parse function are included.
 - additional metadata - defined within read_file function and related to the specific data class - not determined by the user.
    - The `optional_metadata` input parameter (dictionary) is merged into this after being checked against the required keys list to ensure this is not attempting to add any required keys.

Issue #1071 describes the logic flow for this.

### New sub-modules created

As a part of this added the following util sub-modules:
  - function_inputs - creating inputs for functions based on function call
  - metadata_util
     - dealing with dictionaries and includes merging logic related to metadata
     - include definitions of null and "not_set" values for metadata
 
### Points to review

Thoughts/Questions:
 - Definitions around null metadata values have been made more explicit and moved from `openghg.store.spec` to `openghg.util._metadata_util` - are we happy with this location for these details?
    - Note: This was mainly to try and ensure `util` functions aren't getting dependencies on `store` sub-modules but perhaps `store.spec` is a reasonable place for this.
 - At the moment this creates a hard fail when an incompatible match between the metadata sources is found - do we need to try and find a way to soften this (e.g. user flag) so this doesn't catch people out and not allow them to add data?
 - Is there a better, simple way to create filtered dictionaries that don't copy the original dictionary via a dictionary comprehension?

- [x] Closes #1071 

Not needed - internal changes not interface changes


openghg/store/_boundary_conditions.py
```diff
@@ -151,6 +152,8 @@ def read_file(
         if chunks is None:
             chunks = {}
 
+        fn_input_parameters = {**locals()}  # Make a copy of parameters passed to function
```
brendan-m-murphy, line 155: maybe do `locals().copy()` so it doesn't need explanation?

openghg/store/_obssurface.py
```diff
@@ -384,14 +364,17 @@ def read_file(
                 for key, value in data.items():
                     data[key]["data"] = value["data"].chunk(chunks)
 
-            # Mop up and add additional keys to metadata which weren't passed to the parser
-            data = self.update_metadata(data, additional_metadata)
+            # Check to ensure no required keys are being passed through optional_metadata dict
```
brendan-m-murphy, line 367: Lines 367-377 look like they implement roughly the same logic as before, but it's not obvious that they do. Refactoring `get_lookup_keys` should probably be a different PR. To pass metadata around the parser, it seems like it would be enough to merge `additional_input_parameters` into the data that comes out of the parser, then continue as before.

brendan-m-murphy (commented):
It seems like the scope of this PR is much wider than just changing how parameters are passed to the parser. I was expecting the "additional input metadata" to be merged right after the parser function was called, and then for the rest of the logic (flawed as it may be...) to be the same.

I'm guessing it would be a pain to separate these changes, but the changes to the optional metadata handling etc. are fairly big and bring a lot of helper code with them.

openghg/store/_obssurface.py
```diff
@@ -271,6 +261,15 @@ def read_file(
         inlet = clean_string(inlet)
         inlet = format_inlet(inlet)
 
+        # Would like to rename `data_source` to `retrieved_from` but
+        # currently trying to match with keys added from retrieve_atmospheric (ICOS) - Issue #654
+        data_source = "internal"
```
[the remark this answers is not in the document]
rt17603: The comparison is against `data_source` for ICOS which at the moment is set to "icoscp" - this is included so data pulled from icos and added internally can be easily distinguished.

rt17603:
> It seems like the scope of this PR is much wider than just changing how parameters are passed to the parser. I was expecting the "additional input metadata" to be merged right after the parser function was called, and then for the rest of the logic (flawed as it may be...) to be the same.
> 
> I'm guessing it would be a pain to separate these changes, but the changes to the optional metadata handling etc. are fairly big and bring a lot of helper code with them.

I've split out the first step as best I can into a separate PR #1111 - in doing so I recalled the reasons behind some of the additional changes that were made e.g. the fact that optional keys could only be set using `optional_metadata` necessitated these details  being added at the store level to then be passed to `get_lookup_keys`. This only impacted flux as this was the only one making use of optional keys (but would impact all when optional keys were used). In the PR mentioned I had to get around this by adding code to `Flux` which we would want to remove again once the full solution was implemented.

rt17603:
I've split out the step to update how `optional_metadata` is used into a follow on PR  #1112 so this functionality can be compared.

rt17603:
I've included a third PR #1113 which incorporates the updates to metadata merging - this was mainly tidy up (but trying to do this in a more formal way) compared to the previous very simple version of just using the `dict.update` method. This builds on PR #1111 and #1112 .

rt17603:
The functionality within this PR was split out into PR #1111 , #1112 and #1113 which have now been merged.

PR #1070, opened 2024-07-29
Add a FLEXPART option to standardise_footprint

### LINKED ISSUE #905: New footprint formats
### What is your issue?

New footprints will now be processed with a different format and filename.  An example footprint file is located in on BP1 under _shared/temp_anita_rachel/BSD-250magl_NAME_UKV_EUROPE_inert_201412.nc_

openghg needs to be able to ingest these new files and remain compatible with the old format.

hdelongueville:

Updates:
    - Add a paris parser to standardise/footprint
    - Update check_chunks in _base.py

    - [x] Closes #905


brendan-m-murphy:
Related to #905

rt17603 (commented):
Thanks for putting this together - looks good so far. I can finish this off from here and will make this more generic than just flexpart as this should also cover the new NAME format as well.

hdelongueville:
Awesome thanks!

openghg/standardise/footprint/_paris.py
```diff
@@ -0,0 +1,175 @@
...
+    timestamp_now,
+    check_function_open_nc,
+)
+from openghg.store import infer_date_range, update_zero_dim
+from openghg.types import multiPathType
+
+logger = logging.getLogger("openghg.standardise.footprint")
+logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
+
+
```
brendan-m-murphy, line 20: Here or in the init file, you could add `parse_flexpart = parse_paris` and `parse_name = parse_paris`. I think that is all that is needed to "alias" this function.

openghg/standardise/footprint/_paris.py
```diff
@@ -0,0 +1,175 @@
...
+    if high_time_resolution:
+        warnings.warn(
+            "This argument is deprecated and will be replaced in future versions with time_resolved.",
+            DeprecationWarning,
+        )
+        time_resolved = high_time_resolution
+
+    xr_open_fn, filepath = check_function_open_nc(filepath)
+
+    with xr_open_fn(filepath) as temp:
```
brendan-m-murphy, line 70: Using a closed dataset automatically re-opens it: https://github.com/pydata/xarray/issues/4131, so we don't need to use a context manager here. `fp_data = temp.load()` might let the file stay closed, I'm not sure. If you want to ensure the file is closed, you could put all of the code using `fp_data` inside the context manager.

openghg/standardise/footprint/_paris.py
```diff
@@ -0,0 +1,175 @@
...
+    metadata["height"] = inlet
+    metadata["species"] = species
+
+    if met_model is not None:
+        metadata["met_model"] = met_model
+
+    if network is not None:
+        metadata["network"] = network
+
+    # Check if time has 0-dimensions and, if so, expand this so time is 1D
```
brendan-m-murphy, line 114: If a footprint doesn't have a time coordinate, should we raise a value error?

openghg/standardise/footprint/_paris.py
rt17603: I'm not sure about aliasing this for name, in particular, as this is still a specific format from NAME that isn't necessarily universal yet. For our NAME footprints in particular we still have previous format files so either "acrg_org" or "paris" may be valid depending on the file itself. We could perhaps add some fail cases to advise which format to use to help (e.g. check for "srr" or "fp" and mention using the other format).

For flexpart to keep this consistent with the original PR I can add that but, again, in future we may want to think about whether that's universal.

openghg/standardise/footprint/_paris.py
```diff
@@ -0,0 +1,175 @@
...
+            metadata["min_latitude_high"] = round(float(fp_data["lat_high"].min()), 5)
+
+        except KeyError:
+            raise KeyError("Expected high spatial resolution. Unable to find lat_high or lon_high data.")
+
+    metadata["time_resolved"] = str(time_resolved)
+    metadata["high_spatial_resolution"] = str(high_spatial_resolution)
+    metadata["short_lifetime"] = str(short_lifetime)
+
+    metadata["heights"] = [float(h) for h in fp_data.height.values]
```
brendan-m-murphy, line 155: This metadata ("heights", "variables") comes from the data itself, so should be set in the parser, but "time_resolved", "high_spatial_resolution", "short_lifetime" don't need to be here. 

Also there is a lot of repeated code between this and acrg parser (much of it todo with metadata). Maybe open an issue about refactoring the footprint parsers. I suspect removing unnecessary metadata parsing will remove some of this repeated code. For other parts, it should be moved to helper functions to be shared between parsers.

openghg/store/_footprints.py
```diff
@@ -350,39 +349,13 @@ def read_file(
...
-                    short_lifetime = True
-
-        chunks = self.check_chunks(
-            filepaths=filepath,
-            chunks=chunks,
-            high_spatial_resolution=high_spatial_resolution,
-            time_resolved=time_resolved,
-            short_lifetime=short_lifetime,
-        )
+        if time_resolved and sort:
```
brendan-m-murphy, line 355: As far as I understand it, `open_mfdataset` will sort automatically. So if multiple files are passed, they will be sorted by `open_mfdataset` and if one file is passed, there is no need to sort it. So the `sort` option and this warning should probably be removed.

openghg/store/_footprints.py
```diff
@@ -423,11 +395,22 @@ def read_file(
 
         footprint_data = parser_fn(**input_parameters)
 
+        chunks = self.check_chunks(
+            ds=list(footprint_data.values())[0]["data"],
```
brendan-m-murphy, line 399: It almost certainly doesn't matter here, but `next(iter(footprint_data.values()))["data"]` is more efficient because it doesn't convert the iterator from `.values()` into a list. Anyway, we probably want to use the more efficient pattern consistently (we seem to use it in most places).
```suggestion
            ds=next(iter(footprint_data.values()))["data"],
```

brendan-m-murphy (changes requested):
Looks good! A few small suggestions.

openghg/store/_footprints.py
rt17603: This all links to the fact that `sort` is used within `Datasource.add_timed_data` function 

```
225        if sort and drop_duplicates:
226            data = data.drop_duplicates(time_coord, keep="first").sortby(time_coord)
227        elif sort:
228            data = data.sortby(time_coord)
```

so the presence of `sort` does mean an action is being performed. That action may not be necessary so potentially could be removed but I wouldn't want to do that in this PR. In which case I do think the warning is telling us something at the moment.

openghg/store/_footprints.py
rt17603: Ok, makes sense - I committed your suggest for this.

openghg/standardise/footprint/_paris.py
rt17603: Added an else to current if statement for this to raise a `ParseError`

openghg/standardise/footprint/_paris.py
rt17603: The main thing here is just trying to ensure we're not making this slower again so I just want to make sure we've recreated whatever this was doing with the `chunks` lines. If this wasn't closing the file anyway, that's fine and we can keep this as a simple line without the context manager.

rt17603: Ok I've updated both `paris` and `acrg_org` format to use the same method for opening the file (without the context manager).

openghg/store/_footprints.py
brendan-m-murphy: I opened an issue for this: https://github.com/openghg/openghg/issues/1081. In this case, we know that the data either consists of one dataset or was opened by `open_mfdataset`, so we could pass `sort=False` to assign data... anyway we don't have to do this now.

openghg/standardise/footprint/_paris.py
brendan-m-murphy: For this PR, the code in the new parser is very close to the code in the existing parser, so it's just another place where probably code will need to be maintained in parallel. Since we're planning to change the parsers soon, we don't have to cover in this PR. (I realise you didn't write the bulk of this code, but we should try to avoid copying and pasting large sections of code so that we don't get into the `Footprint`, `Flux`, `Surface`, etc... issue again.)

Here's an issue for it: https://github.com/openghg/openghg/issues/1082

rt17603: Yes I understood what you were saying, it's more that I don't think we'd want to spend more time on this PR now given we were trying to just get this code in and not let it sit on a branch people were already using. I agree on avoiding repeated code in general but at this stage I was saying it was better to get this in with an improvement on a separate PR if needed - at least this can be tracked on an issue now if needed. I'm going to resolve this.

PR #1074, opened 2024-08-01
Iss1024 standardisation files are sorted if contains dates

### LINKED ISSUE #1024: Standardisation - ensure lists of files are sorted if they contain dates
We should ensure that any files that contain dates are sorted before the standardisation process starts. This is already done in `Footprints.read_file` but will help us avoid costly sort operations.

- Do this within the standardise function
- Allow us to switch this off by passing a flag]
- Which file formats should we do this with?

SutarPrasad:

- [x] Removed check for `isinstance of list` for `filepath variable` from read_file functions of footprints and obs_surface
- [x] Added files to be sorted in the standardise function based on filenames
- [x] Removed use of filepaths and kept only one variable calle filepath in standardise_surface

- [x] Closes #1024 (Replace xxxx with the Github issue number)


openghg/standardise/_standardise.py
```diff
@@ -40,6 +40,24 @@ def standardise(data_type: str, filepath: multiPathType, store: Optional[str] =
         logger.info("Compression disabled")
         compressor = None
 
+    if data_type == "surface" and kwargs["source_format"].lower() != "gcwerks":
```
rt17603, line 43: A few things on this:
1. If we're only doing this for specific data types could we move this to within the `standardise_surface` and `standardise_footprint` functions instead? Would be nice to keep `standardise` as generic as possible.
2. Since this code is repeated, would be good to create a util function for this to simplify these lines.

rt17603 (changes requested):
Looks mostly completed to me but just a few tidy up suggestions which I think would keep this compartmentalised and keep some generality for standardise.

tests/standardise/test_standardise.py
```diff
@@ -551,3 +551,23 @@ def test_standardise_flux_timeseries():
     )
 
     assert "ch4_crf_uk" in flux_results
+
+def test_standardise_sorting():
```
rt17603, line 555: This test may need a few more checks if this is to check that the data has been stored in a sorted order - e.g. searching and retrieving the data with `sort=False` to ensure sorting isn't imposed when the data is returned.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Adds `data_level` and `data_sublevel` as additional keys which can be used to distinguish observation surface data - [PR #1051](https://github.com/openghg/openghg/pull/1051)
 
+- Added feature of file sorting at standardise level before processing - [PR #1074](https://github.com/openghg/openghg/pull/1074)
```
rt17603, line 22:
```suggestion
- Added feature of file sorting at standardise level before processing and remove `filepaths` input option - [PR #1074](https://github.com/openghg/openghg/pull/1074)
```

openghg/util/_util.py
```diff
@@ -235,3 +236,20 @@ def multiple_inlets(site: str) -> bool:
...
+    Sorting time on filename basis
+
+    Args:
+        filepath: Path to the fil
+
+    Returns:
+        list[Path]: List of sorted paths
+    """
+
+    filepath = sorted([Path(f) for f in filepath])
```
rt17603, line 252: This doesn't account for a `str` being passed at the moment (would loop over the characters in a `str`) - I think the if statement you had in a previous version did though? You may also want to look at using `multiPathType` to keep this consistent.

rt17603 (commented):
A few small, fiddly things to just check here (including expanding on the test and making sure the `str` case is covered for `sort_by_filename`)

rt17603 (approved):
Two small comments but otherwise ready to go in - let me know if you need another explicit approve to be able to merge

rt17603 (approved):

PR #1083, opened 2024-08-05
Tidy up required keys around data_source and dataset_source

### LINKED ISSUE #654: Decide how to handle data for the same base keyword combination (e.g. site/species/height) from different sources (e.g. NRT, ObsPack)

rt17603:

We want to ensure that obs surface data being added can be distinguished by the source of the data. Within retrieved data (which follows a different path to standardise) there are additional keywords including: `data_source` (e.g. "icos", "ceda") and `dataset_source` ("InGOS", "European ObsPack") which it would be useful to have available more generally.

This PR formalises this to allow `dataset_source` to be specified by the user and `data_source` to be set to "internal" when using the `standardise_surface` interface. These keys are part of the set used to distinguish between different stored Datasources.

*Note: Would like `data_source` to be renamed to `retrieved_from` but kept this the same as ICOS (retrieve_atmospheric) for the moment - happy to change this if we can make sure this is backwards compatible.*

- [x] Closes #654 

Not needed:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Code to combine multiple data objects - [PR #1063](https://github.com/openghg/openghg/pull/1063)
 - A new object store config file to allow customisation of metadata keys used to store data in unique Datasources - [PR #983](https://github.com/openghg/openghg/pull/983)
 - New keywords `data_level` and `data_sublevel` as additional keys which can be used to distinguish observation surface data - [PR #1051](https://github.com/openghg/openghg/pull/1051)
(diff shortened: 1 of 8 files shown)


SutarPrasad (approved):
looks good

Issue #1067, opened 2024-07-19
Retire `icos_data_level` and replace with `data_level`


rt17603:
At the moment, when ICOS data is retrieved this is stored with a metadata `icos_data_level` keyword. This can be replaced with the more general `data_level` keyword which will be used for all source_formats.

Need to ensure backwards compatability with previous data added to the object store.

See also #551 for previous discussion on this.

rt17603:
Initial steps to complete this within PR #1084

Issue #612, opened 2023-03-30
Latest version of icoscp uses additional authentication


joe-pitt:
### What is your issue?

As of icoscp version 0.1.18 additional authentication has been implemented via a module called cpauth (as described here: https://icos-carbon-portal.github.io/pylib/modules/#authentication). We either need to modify the ICOS data retrieval functions in openghg accordingly, or fix the version at 0.1.17 in the yaml file.

brendan-m-murphy:
The "advanced" option at the end of this part of their documentation: https://icos-carbon-portal.github.io/pylib/icoscp/authentication/ seems to suggest that if we setup a config file that contains your icos username and password, that this could be used to login in automatically.

I'll try to get this to work.

joe-pitt:
As a temporary fix, I've had success in upgrading to icoscp v0.2.0 and adding the following to the start of my scripts:
```
from icoscp_core.icos import bootstrap
from icoscp import cpauth

cookie_token = ""  # Get this every day by logging into your ICOS CP account
meta, data = bootstrap.fromCookieToken(cookie_token)
cpauth.init_by(data.auth)
```

rt17603:
And I note on the [icoscp Authentication page](https://icos-carbon-portal.github.io/pylib/icoscp/authentication/) this notes you can get your authentication token (daily) from the My Account section: 

> The token can be obtained from the "My Account" page ([ICOS](https://cpauth.icos-cp.eu/)), which can be accessed by logging in using one of the supported authentication mechanisms (username/password, university sign-in, OAuth sign in).

joe-pitt:
Just a heads up that icoscp-0.2.0 had some issues described here: https://github.com/ICOS-Carbon-Portal/pylib/issues/194. This has been fixed in icoscp-0.2.1 but in that release pandas has also been pinned <2, which conflicts with openghg. I've opened an issue on icoscp for this: https://github.com/ICOS-Carbon-Portal/pylib/issues/199. In the meantime my obs scripts seem to be running fine if I manually upgrade pandas to v2.2.3.

rt17603:
So put in another draft PR #1447 for this - basically just to update the docs and drop the 0.17.0 pin for icoscp but we should discuss if we want to incorporate PR #1139 into this (and update the docs to match to this) as this seems to include a helper function for using the ICOS authentication steps. Either/or really but would be good to put something in for this.

PR #1084, opened 2024-08-07
Pin icoscp and fix tests based on current retrieval

rt17603:

The external `icoscp` package introduced new authentication as of 0.18.0 (see #612 ) which hasn't been integrated into openghg yet which means we need to pin to the previous version for now and ensure the current tests pass. This includes the following updates:
 - Pinned the icoscp package to the previous version in requirements.txt and environment.yml
 - Updated the tests/retrieve/icos tests to ensure they pass with the current configuration
 - ~Removed the icos mark for the tests/retrieve/icos tests which have mocked the retrieval so they are always run rather than just when the `--run-icos` flag is used with pytest. All retrieve icos tests were previously skipped due to this mark.~ *Would cause people's tests to fail if they don't have icoscp already pinned so will avoid for now.*

Note: that some of these tests have been built against external services (ICOS carbon portal accessed by icoscp) and so will need to be periodically reviewed / switch to using mockers exclusively for this.

This is needed to be able to continue with #1067 .

- [x] Related to #612 but doesn't fully close this

Not needed


SutarPrasad (approved):
seems straightforward, good to go

rt17603:
Will aim to merge once #1085 is resolved and tests are passing.

tests/retrieve/icos/test_retrieve_icos.py
```diff
@@ -39,7 +39,7 @@ def test_icos_retrieve_skips_datalevel_1_csv_files():
     assert first_obs.metadata == expected_metadata
 
 
-@pytest.mark.icos
+# @pytest.mark.icos
```
rt17603, line 42:
```suggestion
@pytest.mark.icos
```

brendan-m-murphy (approved):
Looks good!

Issue #1035, opened 2024-06-13
Umbrella: Short term updates to organise metadata


rt17603:
Tasks to complete:
- [x] Adding in config framework for metadata - #978 
- [x] Adding data_level and data_sublevel keywords - #551
- [x] Ensuring dataset_source/data_source keys can be added and used to distinguish - #654 
- [x] Ensure icoscp is set up correctly (pinned and tests passing) - PR #1084 
- [x] Remove `icos_data_level` and just used `data_level` within required keys - #1067 (PR #1087)
- [x] Remove keys which aren't needed from surface required keys - PR #1088
- [x] Added ability to perform negative lookup - PR #1064
    - Next step would be to integrate this into `standardise` framework

Moved to next milestone:
- ~Allow metadata keys to be added outside the parse framework at the store level class - #1071~ - moved to umbrella #1092
- ~Move functionality which require all metadata out of parse functions and into store level classes - #986~ - moved to umbrella #1092
- ~TUTORIAL's : https://github.com/openghg/openghg/issues/982~ - moved to 0.10.0 release

brendan-m-murphy:
# Table of Contents

1.  [Background on metadata](#org0179e33)
    1.  [What do we use metadata for?](#orgbac4070)
    2.  [Where does metadata come from?](#org97ccb20)
2.  [Improving how we use metadata](#orgeba5722)
    1.  [What is our goal?](#org9f23b45)
    2.  [Requirements](#org06747f9)
    3.  [It would be nice if&#x2026;](#orge169316)
    4.  [Problems](#org8e49bb1)
3.  [Improving how metadata is handled in the code](#orgc227a67)
4.  [Refactorings](#orga0f8f4d)
    1.  [`BaseStore` children `read_file` after parser applied](#orgfc955ae)
    2.  [`read_file` kwargs, most named arguments go through `standardise`?](#org8dc57f3)
    3.  [standardise functions](#org8ec12be)
5.  [Sort out Gareth&rsquo;s PR 983](#org2cd7a21)
    1.  [add code for reading categorising keys from config](#org6baf40c)
    2.  [remaining metadata tasks](#orgc3f0c3d)
    3.  [fix how optional/informative metadata is processed](#orgac1a7d7)
    4.  [summarise existing state](#orga03e37d)
6.  [make methods (and CLI?) for modifying config file](#orgf50c35c)
7.  [extract repeated code for metadata handling from subclasses of `BaseStore`](#org3c5cc25)
[...]

PR #1088, opened 2024-08-07
Streamlining keywords within surface required config definition

rt17603:

The required keys within the config file should be a mapping onto keys which are used to distinguish datasources within the object store. This is used as part of the data lookup steps when adding new data so we know whether to add data to an existing datasource or to create a new datasource.

This PR removes some additional keywords from the config file which do not need to be used as these distinguishing keys. The keywords removed are:
 - `station_long_name` - the `site` code should cover the same purpose and, in general, the same site code should have the same `station_long_name`
 - `data_type` - the datasources are already split by `data_type` as part of the structure of the object store so this does not need to be used as a distinguishing key. This is also not included as a required key for any of the other data types than surface.
 
Within Issue #1035 we have started to outline the plans for how metadata is used when adding new data. This PR is also a step towards being able to apply the changes within PR #1069 which relies on the above logic to start to generalise this workflow.

- [x] Related to #1071 but does not close this

Not needed


SutarPrasad (approved):
straightforward as discussed, tests are passing, approved

PR #1090, opened 2024-08-08
Fix for failing footprint chunk test

### LINKED ISSUE #1085: Footprint chunk test failing
### What happened?

We seem to be getting the following failed test when running the CI at the moment. One of the more recent PRs #1070 did update how the chunking was used when standardising footprint data but this test does pass locally.
```
FAILED tests/standardise/test_standardise.py::test_standardise_footprints_chunk - AssertionError: assert {'height': (2...e': (1, 1, 1)} == {'height': (2...time': (2, 1)}
```

_No response_

_No response_

### Relevant log output

_No response_

_No response_

### Environment

<details>

</details>

rt17603:

With the recent upgrade of dask to version 2024.8, the chunks returned when retrieving the data from an object store do not match to what we previously expected. This is due to a sort operation which is carried out when all data is retrieved. The sort operation is partially controlled by a `sort` input flag but is performed by default.

The fact that the chunks returned don't exactly match to the stored chunks is not necessarily a problem in itself but it is necessary to test that we are able to add our own custom chunks and this is stored correctly. To fix this test and allow this check to be made the following updates were made:
 - Exposed the `sort` keyword when creating BaseData through retrieve interface. This allows the sortby operation to be avoided.
    - Note: added warning that `sort=False` is not applied if there is a start_date and end_date range supplied.
 - Update the `tests/standardise/test_standardise.py::test_standardise_footprints_chunk` test to make sure the data isn't sorted when being retrieved so the chunks can be checked.

- [x] Closes #1085  

Not needed

### DIFF
--- a/openghg/dataobjects/_basedata.py
+++ b/openghg/dataobjects/_basedata.py
@@ -37,7 +37,7 @@ class _BaseData:
             version: Version of data requested from Datasource
             start_date: Start date of data to retrieve
             end_date: End date of data to retrieve
-            sort: Sort the resulting Dataset by the time dimension, defaults to False
+            sort: Sort the resulting Dataset by the time dimension, defaults to True
             elevate_inlet: Force the elevation of the inlet attribute
             attrs_to_check: Attributes to check for duplicates. If duplicates are present
                 a new data variable will be created containing the values from each dataset
@@ -83,6 +83,11 @@ class _BaseData:
             self.data = self._zarrstore.get(version=version)
             if slice_time:
                 # If slicing by time, this must be sorted along the time dimension
+                if sort is False:
+                    logger.warning(
+                        "Ignoring sort={sort} input as it is necessary to sort the data when extracting a start and end date range."
+                    )
+
                 self.data = self.data.sortby("time")
                 sorted = True
 
--- a/openghg/dataobjects/_basedata.py
+++ b/openghg/dataobjects/_basedata.py
@@ -85,7 +85,7 @@ class _BaseData:
                 # If slicing by time, this must be sorted along the time dimension
                 if sort is False:
                     logger.warning(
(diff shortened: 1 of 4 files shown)


openghg/dataobjects/_basedata.py
```diff
@@ -83,6 +83,11 @@ def __init__(
             self.data = self._zarrstore.get(version=version)
             if slice_time:
                 # If slicing by time, this must be sorted along the time dimension
+                if sort is False:
+                    logger.warning(
+                        "Ignoring sort={sort} input as it is necessary to sort the data when extracting a start and end date range."
```
rt17603, line 88:
```suggestion
                        f"Ignoring sort={sort} input as it is necessary to sort the data when extracting a start and end date range."
```

openghg/dataobjects/_basedata.py
```diff
@@ -83,6 +83,11 @@ def __init__(
             self.data = self._zarrstore.get(version=version)
             if slice_time:
                 # If slicing by time, this must be sorted along the time dimension
+                if sort is False:
```
brendan-m-murphy, line 86: Here you could add `if not sorted: ...` and maybe at the beginning we could add `sorted = np.all(self.data.time[:-1] <= self.data.time[1:])`

brendan-m-murphy (approved):
Looks good! I suggested a check to see if the data is sorted already, which would mean avoiding sorting when we don't need to. But that doesn't have to go in this PR.

openghg/dataobjects/_basedata.py
rt17603: I had a quick go at adding it but it's proving a little more complex as self.data isn't initially specified and there are multiple ways to add the data so data can be None initially.

Issue #1091 opened for this going forward.

PR #1094, opened 2024-08-13
Tidy up: allowing parser functions to be found using data_type input

rt17603:

On the road to being able to pull out and centralise data_type `.read_file` functionality, this PR creates two additional functions which links the names of the data_types to the currently used parser definition classes and replaces the specific import within the relevant store classes. This is so that only the `data_type` input is needed to ensure the correct details can be accessed within the `read_file` and `transform_data` methods.

The two functions are:
 - `openghg.store.spec.define_standardise_parser`
 - `openghg.store.spec.define_transform_parser`

At the moment this impacts:
 - `ObsSurface`
 - `Flux` (standardise and transform)
 - `FluxTimeseries`
 - `Footprints`
 - `ObsColumn`

*Note: We could move these definitions from where they currently sit within `openghg.types._enum.py` and, longer term, may decide to not represent these as classes.*

- [x] Related to steps for #1071 but does not close this

**Do we want this?**

Not needed


SutarPrasad (commented):
Code, Looks good, we should have tests for these functions. Ensure code coverage and also it will serve as a guide for further parser additions.

brendan-m-murphy (approved):
This looks good. As with #1095, we might want to revisit the location (or structure) of these functions, but it doesn't have to happen now.

rt17603:
> Code, Looks good, we should have tests for these functions. Ensure code coverage and also it will serve as a guide for further parser additions.

In this case these functions are largely just defining details so we could test but is potentially less essential in this case and will be checked as part of the overall test base.

PR #1104, opened 2024-08-21
Iss1103: Updating required key logic for ObsColumn data class to allow for better flexibility between site/satellite split

### LINKED ISSUE #1103: Update ObsColumn required keys to better map to site/satellite flexibility for this data class
At the moment, [`ObsColumn`](https://github.com/openghg/openghg/blob/devel/openghg/store/_obscolumn.py) is set up to support column data from both satellite and site data and the allowed keywords reflect this. Within the current [`defaults.json` file](https://github.com/openghg/openghg/blob/devel/openghg/data/config/objectstore/defaults.json#L42), which sets the required keywords for the "column" data_type, this currently requires all keys related to site AND satellite when we should only need the keys associated with site OR satellite. As a first step to make this more flexible we can reduce the number of required keys at the outset, allowing more to be optional and then consider if we need to add in more complex logic to check this going forward (or split this out into multiple data classes).

Suggestion keywords:
 - required: `species`, `platform` (to state this is "site" OR "satellite")
 - optional: `site`, `satellite`, `selection`, `domain`, `network` 

Can add extra logic around `site` OR `satellite` not `site` AND `satellite` in the code if needed.

rt17603:

Updated required and optional keys for "column" data type (`ObsColumn` data class). This includes all the same keys as previous but rearranged into required and optional definitions to better to match to how we expect this to be used for satellite OR site column data. Also added `platform` as a required key (should be set to "site" or "satellite" though this isn't imposed much at the moment) - this was already a key that could be added through the `standardise_*` framework but was not included within the config.

The defaults.json config now includes:
 - required: species, platform [added]
 - optional: satellite, selection, domain, site, network

Includes:
 - Added in check within `ObsColumn.read_file` for `satellite` OR `site` but not `satellite` AND `site`
 - Removed `min_keys=3` when performing `self.assign_data()` step so this just uses the number of keys passed.
 - Updated `optional_metadata` test so this uses one of the new required keys (was using `site` which is now optional)

  - [x] Closes #1103 and feeds into #1071 (which needs to assume expected required key logic is correct)

- Not needed for this PR
  - Documentation and tutorials NOT updated/added - No docs on the "column" data_type yet but should add in a separate PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased](https://github.com/openghg/openghg/compare/0.9.0...HEAD)
 
+### Updated
+
+- Required and optional keys for "column" data type were updated to reflect the two sources of this data (site, satellite) [PR #1104](https://github.com/openghg/openghg/pull/1104)
+
 ### Fixed
 
 - Bug where an input filepath list to standardise_surface was only storing the last file hash. This allowed for some files to bypass the check for the same files depending on where they were in the original filepath list. [PR #1100](https://github.com/openghg/openghg/pull/1100)
--- a/openghg/data/config/objectstore/defaults.json
+++ b/openghg/data/config/objectstore/defaults.json
@@ -41,6 +41,18 @@
     },
     "column": {
         "required": {
+            "species": {
+                "type": [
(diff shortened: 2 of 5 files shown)


SutarPrasad (approved):
Looks good, ready for merge after the changelog conflict is resolved

SutarPrasad (approved):

PR #1109, opened 2024-08-23
Tidy up: Split out sampling period check

rt17603:

Move check for sampling_period into the util._time.py module. This is to continue to simplify the checks within `ObsSurface.read_file()` and feeds into PR #1069 to make sure inputs to the read_file function call matches as closely as we can to what is accepted by the `parse` functions. Previously `sampling_period` input was converted to `sampling_period_seconds` which is what was passed to the `parse` functions.

Related to this, sampling_period of "1H" was changed to "1h" as this was causing a warning in the tests.

- [x] Related to #1071 but does not close this

- Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature - small internal change - needed?

Not needed

### DIFF
--- a/openghg/util/_time.py
+++ b/openghg/util/_time.py
@@ -945,10 +945,10 @@ def evaluate_sampling_period(sampling_period: Optional[Union[Timedelta, str]]) -
         # Check string passed can be evaluated as a Timedelta object and extract this in seconds.
         try:
             sampling_period_td = Timedelta(sampling_period)
-        except ValueError:
+        except ValueError as e:
             raise ValueError(
                 f"Could not evaluate sampling period: '{sampling_period}'. Must be specified as a string with valid unit (e.g. 1m for 1 minute)."
-            )
+            ) from e
 
         sampling_period = str(float(sampling_period_td.total_seconds()))
 
--- a/openghg/util/_time.py
+++ b/openghg/util/_time.py
@@ -1,9 +1,10 @@
 from datetime import date
 from typing import Dict, List, Optional, Tuple, Union
-from openghg.types import TimePeriod
-
 from pandas import DataFrame, DateOffset, DatetimeIndex, Timedelta, Timestamp
 from xarray import Dataset
+import re
+
+from openghg.types import TimePeriod
 
 __all__ = [
     "timestamp_tzaware",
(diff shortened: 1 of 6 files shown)


openghg/util/_time.py
```diff
@@ -911,3 +913,51 @@ def dates_in_range(
...
+                f"Invalid sampling period: '{sampling_period}'. Must be specified as a string with unit (e.g. 1m for 1 minute)."
+            )
+
+        # Check string passed can be evaluated as a Timedelta object and extract this in seconds.
+        try:
+            sampling_period_td = Timedelta(sampling_period)
+        except ValueError:
+            raise ValueError(
+                f"Could not evaluate sampling period: '{sampling_period}'. Must be specified as a string with valid unit (e.g. 1m for 1 minute)."
+            )
```
brendan-m-murphy, line 951:
```suggestion
            ) from e
```
This should make the error message more specific (it will say that the pandas `ValueError` directly caused the error that we raise).

openghg/util/_time.py
```diff
@@ -911,3 +913,51 @@ def dates_in_range(
...
+    Returns:
+        str : Sampling period as a string containing the number of seconds.
+
+    TODO: Integrate sampling_period handling into logic for time_period (if practical)
+    """
+    # If we have a sampling period passed we want the number of seconds
+    if sampling_period is not None:
+
+        # Check value passed is not just a number with no units
+        try:
```
brendan-m-murphy, line 935: It would probably be better to rewrite this with a regex that detects exactly what we're trying to avoid (integers as strings) e.g. `r"^\s*\d+\s*"` If it is a float (as a string), Timedelta will complain that there is no unit. If it is an int, pandas would treat it as a number of nanoseconds (I think), so that's the case to catch.

brendan-m-murphy (approved):
Looks good. I suggested some small changes. The inline changes around the exception should make the traceback says "this error was directly caused by..." instead of "during the handling of this exception another exception occurred" which I find slightly confusing. Anyway, feel free to leave as is, or re-request approval if you need it.

openghg/util/_time.py
rt17603: Thanks for the suggestion - this was mainly a PR to reorganise but I do agree it is worth trying to improve this functionality if/when we have the chance to.

I only spotted your suggested regex after I'd already written one - r"\d+[.]?\d*\s*[a-zA-Z]+" but overall they seem to achieve the same check.

brendan-m-murphy (approved):
Looks good!

PR #1112, opened 2024-09-02
Updating logic for collecting optional keys to use metadata not just optional_metadata

rt17603:

Splitting out https://github.com/openghg/openghg/pull/1069 to demonstrate the logical steps (Part 2): this is to update the creation of the lookup keys to use the overall metadata rather than just the optional_metadata input. This builds on Part 1 #1111 .

#### Updates to use of optional_metadata

The `optional_metadata` input parameter is now handled by:
 - checking this against the required keys from the config file separately to make sure there is no overlap (new method `BaseStore.check_info_keys()`)
 - adding this to the `additional_metadata` so this gets merged into the overall metadata
 
#### Updates to get_lookup_keys

The `get_lookup_keys` step is then updated to:
 - use the combined metadata (output from `BaseStore.update_metadata` which merges the output from the parse function, the remaining input parameters and the additional metadata) and match this against the optional keys from the config file.

#### Notes

 - The temporary code in `Flux` L215 has been removed now the logic is updated to grab optional metadata more centrally
 - The test files for `test_read_file_fails_ambiguous` had to be updated as the new structure was now using attributes from these files which didn't match to the test cases.

- [x] Part 2/3 in final stages towards finishing #1071 

Not needed


openghg/store/base/_base.py
```diff
@@ -204,13 +204,37 @@ def update_metadata(self, data: Dict, input_parameters: dict, additional_metadat
...
+    def check_info_keys(self, optional_metadata: Optional[Dict]) -> None:
+        """Check the informational metadata is not being used to set required keys.
+
+        Args:
+            optional_metadata: Additional informational metadata
+
+        Returns:
+            None
+            Raises ValueError if any keys within optional_metadata are
+                within the required set of keys.
```
brendan-m-murphy, line 216: I know we don't have an official docstring format, but I think there are some formats where "Raises" is a new section like "Args" and "Returns".

Also, is `optional_metadata` optional here so that you don't have to check when using the this function?

openghg/store/base/_base.py
```diff
@@ -220,19 +244,16 @@ def get_lookup_keys(self, optional_metadata: Optional[Dict]) -> List[str]:
...
         lookup_keys = list(required)
-        # Check if anything in optional_metadata tries to override our required keys
-        if optional_metadata is not None:
-            common_keys = set(required) & set(optional_metadata)
 
-            if common_keys:
-                raise ValueError(
-                    f"The following optional metadata keys are already present in required keys: {', '.join(common_keys)}"
-                )
+        # Note: Just grabbing the first entry in data at the moment
```
brendan-m-murphy, line 248: Now that this doesn't depend on passing `optional_metadata`, it could be incorporated into `datasource_lookup`, so it could be call on each entry separately. I think it could be called around line 380 of `_base.py`. (So `required_keys` wouldn't be passed to `assign_data`.)

openghg/store/_flux_timeseries.py
```diff
@@ -184,6 +184,8 @@ def read_file(
         if optional_metadata is None:
```
brendan-m-murphy, line 184: I'm not sure this is necessary, since `check_info_keys` allows `optional_metadata` to be `None`, and we check if `optional_metadata` is `None` on line 189. Probably a similar comment applies to the other data classes.

brendan-m-murphy (changes requested):
Looks good, removing the dependence of `get_lookup_keys` on `optional_metadata` is definitely an improvement. I have a few suggestions that take advantage of this change.

openghg/store/_flux_timeseries.py
rt17603: Makes sense to remove - I think this was left over from when additional optional keys were being added to `optional_metadata` so should only impact `Flux` and `FluxTimeseries`.

openghg/store/base/_base.py
rt17603: Yeah you're right - [Google style guide](https://google.github.io/styleguide/pyguide.html) gives:
```
Raises:
      ConnectionError: If no available port is found.
```

openghg/store/base/_base.py
rt17603: Good idea - I'd updated `assign_data` and the data type classes to accommodate this.

I did leave `required_keys` input but it is now optional and uses the output of `get_lookup_keys` by default.

brendan-m-murphy: I guess this would be a bigger change, but the datasource lookup should probably be done on each item in `data`. Maybe I'll open an issue for that. Then we know that the required keys are specific to the data for one ("proto") datasource.

brendan-m-murphy (approved):
Looks good!

openghg/store/base/_base.py
rt17603: Yes I know what you mean - before we were always assuming the same *keys* would be used for lookup with different values, and I believe in most cases that would still be true, but we could have the occasional situation where its not true. I don't believe we have any examples of that yet though. New issue for this makes sense to keep track but perhaps low priority for now.

Issue #1078, opened 2024-08-02
Umbrella: Adding in functionality to create an ObsPack


rt17603:
We want to be able to generate an ObsPack from an openghg object store. Some of the specifics we may want to split out into a separate or multiple separate repositories but for now we shall collect together the steps here. 

Steps:
- Define what goes into an obspack
   - Set up a very explicit config file which can be used to create search queries (could be a csv, json or toml)
   - Set up a special keyword "tag" which can be used to collect together data to include in an obspack
- Side check: ensure output is suitably standardised 
   - #1077 - links to making sure this is done for ICOS
- Search object store and retrieve the data (based on config initially and them generalising to be able to use tags)
- Collect together this data and output to the obspack framework
   - Need details to include in obspack framework and defined structure
   - Use metadata to create filenames for output nc files

rt17603:
We discussed recording actual UUIDs, versions for the data and search metadata to make sure creation of a obspack from a given object store is reproducible. This will be important once we have the fundamentals of this down and as we start to improve the process.

rt17603:
Suggestion from @brendan-m-murphy 

> You could parallelise this easily with dask "bags" if you like. Maybe after validating file names, make a dask "bag" containing tuples: (ds, output_name) and then apply a function like lambda x: x[0].to_netcdf(x[1]). If the output files are all unique, then you can just write everything in parallel.
>
> (If you do this, you would probably want to use a different loop to make the site_details_df.)
>
> (Also, it might be better to make the bag contain tuples (obs_data, output_name) and use .map with lambda x: x[0].data.to_netcdf(x[1]), since calling .data might trigger something that dask would like to know about.)

rt17603:
Excellent restructure suggestion based on current PR #1117 

> There's a lot of code here. I think it might be easier to understand if the code modelled the ObsPack more closely.
> 
> The code could broadly split into two phases:
> 
>  -  parse the search input to make the obspack structure (which would be represented in an ObsPack object)
>  - write out the data
> 
> Obviously this is what the code is already doing, but the point would be to focus on building the obs pack structure in the code using objects. Right now it feels like the structure is encoded in many functions that pass file paths around.
> 
> In addition to your StoredData object, you could have a Collection object (in the comments, I suggested ObsPack and Subfolder objects, although they both share some common features) that contains StoreData and other Collection objects. Basically, the Collection object is like a folder, and the StoredData object is like a file.
>
[...]

PR #1117, opened 2024-09-06
Adding initial obspack creation logic

### LINKED ISSUE #1080: Create an ObsPack from direct input keywords

Want to be able to put together an ObsPack output from the data stored in the object store.

Will need to add functionality which can take in search criteria - initially can be just a config file containing search details.

For future issues can move to using tags as needed), search the object store and then compile these details into an output ObsPack.

rt17603:

Creating a new `datapack` submodule and adding the initial framework for creation of an obspack. Mostly want to get in this initial version so we can keep building on it.

The process is:
 - Supply a input file with search terms or a DataFrame
 - These search terms are used to search an object store (*note object stores available currently depends on users own openghg.conf setup*) and extract the data
 - Obspack folder structure is created (subfolders for data files split by observation type). This includes:
    - Basic readme file is added with fair use policy details
    - Summary file is created containing site details
    - Data files are stored within the subfolders (surface-insitu, surface-flask) - current file naming based on species, site, inlet by default (this can be customised)

Features:
 - Obspack versioning - a major or minor version can be associated with an obspack using the `obspack_stub` input. This checks for other folders with the same `obspack_stub` and increments the version based on this.
 - File naming - if the filenames produced are not unique, this uses additional keys defined in the config in the object store to try and determine a unique name, if possible.

TODO (longer term):
 - For site_index file
    - Remove version for this
    - Include creation date of the file within the header
 - Bug in unique name checking where two names will be seen as unique if the versions are different even if the metadata keys are the same.

See: [Creating an obspack tutorial](https://github.com/openghg/openghg/blob/make_obspack/doc/source/tutorials/local/Outputting_data/Creating_obspack.rst) for some overall details.

See also: [ACRG-Bristol/gemma_obspack](https://github.com/ACRG-Bristol/gemma_obspack) for an example of how this can be used to create a GEMMA obspack.

- [x] Closes #1080 


openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,717 @@
...
+
+    site_details = df.groupby(site_column).apply(collate_strings)
+    site_details = site_details.dropna(axis=1, how="all")
+
+    # index_output_name = open(obspack_folder / f"site_index_details_{version}.txt", "w")
+
+    # Add header to file
+    output_file.write("++++++++++++\n")
+    output_file.write("Site details\n")
+    output_file.write(f"File created: {pd.to_datetime('today').normalize}\n")
```
brendan-m-murphy, line 622:
```suggestion
    output_file.write(f"File created: {pd.to_datetime('today').normalize()}\n")
```

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,717 @@
...
+
+        df_new[name] = [collated_value]
+
+    return df_new
+
+
+def define_site_details(ds: xr.Dataset, obs_type: str, strict: bool = False) -> dict:
+    """
+    Extract associated site details as a dictionary for a given Dataset. Expect these details to
+    be incuded within the dataset attributes.
```
brendan-m-murphy, line 537:
```suggestion
    be included within the dataset attributes.
```

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,717 @@
...
+
+    if site_column not in df.columns:
+        msg = "Unable to create site details file: 'Site code' column is not present in provided DataFrame."
+        logger.exception(msg)
+        raise ValueError(msg)
+
+    logger.info(f"Writing site details to: {output_filename}")
+    output_file = open(output_filename, "w")
+
+    site_details = df.groupby(site_column).apply(collate_strings)
```
brendan-m-murphy, line 614: This raises a warning (for me at least): DeprecationWarning: DataFrameGroupBy.apply operated on the grouping columns. This behavior is deprecated, and in a future version of pandas the grouping columns will be excluded from the operation. Either pass `include_groups=False` to exclude the groupings or explicitly select the grouping columns after groupby to silence this warning.

Seems like `include_groups=False` should fix it?

brendan-m-murphy, line 614: The site details don't seem to include the species of gas. For the search df in the tutorial, the co2 and ch4 results at TAC are combined into a single row. 

I guess there are two things I'm unclear on:
1. should the species be included?
2. why are the rows compressed like this?  It looks like this is meant to be human readable, and the comma separated lists inside the fields might make it hard to read if these lists are long.

Also, should there be a copy of the original search df/csv included for reproducibility?

brendan-m-murphy (changes requested):
Looks good overall. There are some typos that need to be fixed, and there are some bigger points I'm not clear on (but the answer might be "that is what obspacks look like").

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
```diff
@@ -0,0 +1,224 @@
...
+
+For this obspack the expected structure is:
+
+    test_obspack_v1/
+
+        obspack_README.md
+
+        site_index_details*.txt
+
+        site_insitu/
```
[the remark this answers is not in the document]
rt17603: I think this should actually be `surface-insitu` and `surface-flask` - I'll correct them both.

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
```diff
@@ -0,0 +1,224 @@
...
+
+::
+
+    %ls "{output_folder}/test_obspack_v1/surface-insitu"
+
+::
+
+    %ls "{output_folder}/test_obspack_v1/surface-flask"
+
+*Version included within the site filenames is currently just the version for the obspack but should be updated
```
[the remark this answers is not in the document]
rt17603: What metadata would you be thinking for this? Each nc file should have attributes which contains the key details at least?

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
```diff
@@ -0,0 +1,224 @@
...
+
+.. code:: ipython3
+
+    obspack_stub = "test_obspack"
+
+    create_obspack(search_df=search_df,
+                   output_folder=output_folder,
+                   obspack_name=obspack_name)
+
+This will create a new obspack with a version based on other folder which have the same
```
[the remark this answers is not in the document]
rt17603: > There's sort of weird behaviour if say, you make an obspack called "test_obspack", then make a second version using "test_obspack" as the stub. Then the second version has _v1 at the end, and the first obspack created has no version at the end.

Yeah you're right it does do this which may seem a bit strange - what would be an alternative though do you think? I don't think we'd want to rename the first obspack but would we just want to start the next one as v2 perhaps?

> Inside both of these obspacks, the filenames end with _v1.

Yeah at the moment it default to v1 if no version is specified and then uses v1 when it's version 1 - again this could perhaps be solved by making it `test_obspack` and then `test_obspack_v2` right off the bat but also see below.

Overall though I see this as 2 things:
1. How external obspack version number should behave
2. How internal versions are set
  - Really the *version* of the obspack and the *version* for the file shouldn't need to be connected at all e.g. as per the comment included above "Version included within the site filenames is currently just the version for the obspack but should be updated to be the object store internal version" so maybe if I just added that change now it would solve this second issue at least.

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
```diff
@@ -0,0 +1,224 @@
...
+
+To create a new minor version, this can be run with the ``minor_version_only`` flag:
+
+.. code:: ipython3
+
+    obspack_stub = "test_obspack"
+
+    create_obspack(search_df=search_df,
+                   output_folder=output_folder,
+                   obspack_name=obspack_name,
```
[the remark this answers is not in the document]
rt17603: Ah yep you're right (thought I'd caught those)

```suggestion
                   obspack_stub=obspack_stub,
```

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
```diff
@@ -0,0 +1,224 @@
...
+
+To create a new major version,  this can be run with the ``major_version_only`` flag:
+
+.. code:: ipython3
+
+    obspack_stub = "test_obspack"
+
+    create_obspack(search_df=search_df,
+                   output_folder=output_folder,
+                   obspack_name=obspack_name,
```
rt17603, line 201:
```suggestion
                   obspack_stub=obspack_stub,
```

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
```diff
@@ -0,0 +1,224 @@
...
+.. code:: ipython3
+
+    import pathlib
+    from openghg.datapack import create_obspack  
+
+.. code:: ipython3
+
+    obspack_name = "test_obspack_v1"
+
+    create_obspack(search_df=search_df,
```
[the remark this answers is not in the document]
rt17603: Yeah the `create_obspack` function should return the `obspack_path` and `output_folder` is the parent folder for this. Did you think anything else needed adding around this e.g. in the documentation?

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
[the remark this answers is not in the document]
rt17603: Ok I've addressed (2) by updating this to now use the internal data version rather than the obspack version.

For (1) I could take a look at updating this so that if "test_obspack" is found, it then created "test_obspack_v2" as the next obspack?

openghg/datapack/_obspack.py
rt17603: Yeah the idea with the site details file basically to include a quick summary of just the site information so the measurement owner could be contacted if people who had the obspack needed to do so. I think we can work on making it more human readable as needed but I think I prefer the tabular format for now as a summary? I can look into if splitting by header rows could make this easier / see the data team what they'd prefer.

Overall, we should work to ensure the attributes for each file contain the more exhaustive list of details needed for the data.

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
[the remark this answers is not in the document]
brendan-m-murphy: I guess adding numbers for files that would have the same name is kind of default on some operating systems (like if you download the same file twice), so `test_obspack`, `test_obspack_v1` isn't that weird really.

I think changing the versions for the file is good though, since it reduces redundant info and avoids that weird corner case.

Anyway, happy to leave (1) alone.

openghg/datapack/_obspack.py
```diff
@@ -68,7 +69,9 @@ def define_filename(
         metadata: Dictionary containing the metadata keys to use
         obs_type: Name of the observation type to be included in the obspack.
             See define_obs_types() for details of obs_type values.
-        version: Version number for obspack. Default = "v1"
+        include_version: Whether to include the data version in the filename. Default = True.
```
brendan-m-murphy, line 72: Just a thought: If the data is for external use, then this internal object store version might be confusing. (OTOH, at some point we discussed exporting UUIDs and versions so that we know exactly what data was used. Maybe that can go in some kind of "manifest" file describing what is in the obspack.)

openghg/datapack/_obspack.py
brendan-m-murphy: Okay, we can leave it as is. I guess a nested tabular structure like I suggested would be harder to work with in code, so maybe not worth trying unless necessary.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,771 @@
...
+
+    obspack_path = create_obspack_structure(
+        output_folder, obspack_name, obs_types=unique_obs_types, release_files=release_files
+    )
+
+    site_detail_rows = []
+    for data, obs_type in zip(retrieved_data, obs_types):
+        metadata = data.metadata
+
+        output_name = define_obspack_filename(
```
brendan-m-murphy, line 754: It might be worth checking that there are no duplicate file names before starting to write the data.

I guess the intention is to avoid this, but it seems possible to include data from the same species/site/inlet but different instruments or sources.

Checking for duplicate output filepaths could be a fairly easy way to check this. (And would preempt a later more confusing error when `.to_netcdf` complains.)

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,771 @@
...
+            metadata=metadata,
+            obs_type=obs_type,
+            obspack_path=obspack_path,
+            include_version=include_data_versions,
+        )
+        ds = data.data
```
brendan-m-murphy, line 760: You could parallelise this easily with dask "bags" if you like. Maybe after validating file names, make a dask "bag" containing tuples: `(ds, output_name)` and then apply a function like `lambda x: x[0].to_netcdf(x[1])`. If the output files are all unique, then you can just write everything in parallel.

(If you do this, you would probably want to use a different loop to make the `site_details_df`.)

(Also, it might be better to make the bag contain tuples `(obs_data, output_name)` and use `.map` with `lambda x: x[0].data.to_netcdf(x[1])`, since calling `.data` might trigger something that dask would like to know about.)

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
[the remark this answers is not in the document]
brendan-m-murphy: I just meant that doing, e.g.
```
result = create_obspack(search_filename=output_folder / "search_config.csv", output_folder=output_folder, obspack_name=obspack_name)
!ls -R {result}
```
 is handy.

But you don't need to include this if you don't want to. You give an alternate way to do it below.

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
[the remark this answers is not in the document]
brendan-m-murphy: At some point we discussed recording actual UUIDs and versions for the data. I guess this and the search metadata should be enough for reproducibility. This doesn't have to be included now.

brendan-m-murphy (approved):
Looks good! I included some suggested improvements, but they don't need to happen now.

openghg/datapack/_obspack.py
rt17603: When we were thinking from observation data perspective the idea around it was that the version in the object store would represent the reprocessed version of the data so e.g. v1 would be the initial data, then it would be reprocessed externally after some additional calibration (v2) etc. It would also help to distinguish between previous obspacks e.g. this new obspack contains an updated version of the data.

Its absolutely not a perfect assumption and requires good maintenance of an object store but that was the logic at least. We can see how that works in practice though.

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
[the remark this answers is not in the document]
rt17603: Yeah you're right - I'll make a note on the umbrella issue so we don't forget to do this as I think it would be valuable.

openghg/datapack/_obspack.py
rt17603: Thanks for catching that - should be added now

doc/source/tutorials/local/Outputting_data/Creating_obspack.rst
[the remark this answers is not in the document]
rt17603: I've updated this now so it does treat `test_obspack` as version 1 and so the next version will be `test_obspack_v2` - I think that does seem like a reasonable update to make and can always change it back if people using it decide the old way was better.

openghg/datapack/_obspack.py
rt17603: I've put in quite a few changes to try and address this. It basically does the following:
 - creates the output names (based on name_components - e.g. ["species", "site", "inlet"]
 - checks for duplicate names
 - if duplicates are present, creates new output names based on additional metakeys from data_type config e.g. ["species", "site", "inlet", "data_level"]

As part of this initial draft, to do this I started to construct an object to hold some of these details - `StoredData` as it got a bit complex trying to link the filename and metadata otherwise. This is still fairly minimal but can be expanded upon in future versions as needed.

Caveat: one known issue at the moment is that because the filename also includes the version it can see two filenames as different if the two data sources have different versions even if the other keys match. I can try and address that in this PR but can leave that for now as a corner case.

brendan-m-murphy (commented):
The new naming stuff looks really nice.

There's a lot of code here. I think it might be easier to understand if the code modelled the ObsPack more closely.

The code could broadly split into two phases:
1. parse the search input to make the obspack structure (which would be represented in an `ObsPack` object)
2. write out the data

Obviously this is what the code is already doing, but the point would be to focus on building the obs pack structure in the code using objects. Right now it feels like the structure is encoded in many functions that pass file paths around. 

In addition to your `StoredData` object, you could have a `Collection` object (in the comments, I suggested `ObsPack` and `Subfolder` objects, although they both share some common features) that contains `StoreData` and other `Collection` objects. Basically, the `Collection` object is like a folder, and the `StoredData` object is like a file.

Also, maybe some of the code can be split into separate modules? For instance, a submodule around the code that parses the input files and organises the search results.
This might also include the code that builds the `ObsPack`, `Subfolder`, `StoredData` objects.

An `ObsPack` object could also bring together some of the code into a logical collection, etc.

I'm happy to try to implement some of these suggestions if you don't have time.

Anyway, overall, there is lots of good functionality, but continuing what you started with the `StoredData` objects would bring some nice structure to the code.

rt17603:
> The new naming stuff looks really nice.
> 
> There's a lot of code here. I think it might be easier to understand if the code modelled the ObsPack more closely.
> 
> The code could broadly split into two phases:
> 
>     1. parse the search input to make the obspack structure (which would be represented in an `ObsPack` object)
> 
>     2. write out the data
> 
> 
> Obviously this is what the code is already doing, but the point would be to focus on building the obs pack structure in the code using objects. Right now it feels like the structure is encoded in many functions that pass file paths around.
> 
> In addition to your `StoredData` object, you could have a `Collection` object (in the comments, I suggested `ObsPack` and `Subfolder` objects, although they both share some common features) that contains `StoreData` and other `Collection` objects. Basically, the `Collection` object is like a folder, and the `StoredData` object is like a file.
> 
> Also, maybe some of the code can be split into separate modules? For instance, a submodule around the code that parses the input files and organises the search results. This might also include the code that builds the `ObsPack`, `Subfolder`, `StoredData` objects.
> 
> An `ObsPack` object could also bring together some of the code into a logical collection, etc.
> 
> I'm happy to try to implement some of these suggestions if you don't have time.
> 
> Anyway, overall, there is lots of good functionality, but continuing what you started with the `StoredData` objects would bring some nice structure to the code.

I've had similiar thoughts about restructuring to create an `ObsPack` object which came naturally from starting down the `StoredData` route - I think it does make sense.

However, I do think what is here now does work well for the moment and this does add an initial ability to create an obspack which we can then continue to build on. Restructuring as a next step to include these types of internal updates wouldn't need to change the front-facing code too much from what we have here. Overall, I'd like to get this in soon so we can move onto the next steps.

openghg/datapack/_obspack.py
```diff
@@ -7,14 +7,86 @@
...
+    """
+    This class contains details of the data extracted from the object store with details
+    of how this is related to the obspack this will be outputted to.
+    """
+
+    def __init__(
+        self,
+        data: ObsOutputType,
+        obs_type: str = "surface-insitu",
+        obspack_path: pathType | None = None,
```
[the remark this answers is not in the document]
rt17603: We do use `pathType` (or `optionalPathType`) in a few places in the code - perhaps we could do a find and replace separately for this if we did want to change this to `str | Path`? Actually though, I do think it's useful having an alias for this as it does get repeated - at least it means we're always allowing a `str` or a `Path` whereas otherwise we may forget and only include one or the other.

openghg/datapack/_obspack.py
```diff
@@ -7,14 +7,86 @@
...
+
+        self.obs_type = obs_type
+        self.obspack_path = obspack_path
+
+        if isinstance(obspack_filename, str):
+            obspack_filename = Path(obspack_filename)
+
+        self.obspack_filename = obspack_filename
+
+    def define_obspack_filename(
```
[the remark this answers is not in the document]
rt17603: It's more that sometimes you may want to call this without overwriting the currently stored value (whether that's via `__init__` or calling the function previously). That happens when you're cycling through creating new names and checking whether they're unique or now. That's why the `add_to_object` bool is included.

requirements.txt
```diff
@@ -14,6 +14,7 @@ numexpr
 numpy < 2.0
 openghg_defs
 openpyxl
+setuptools
```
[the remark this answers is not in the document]
rt17603: I seem to recall I had an issue where my environment claimed that `setuptools` wasn't available? I can remove it though if we don't think it's needed.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+            version=version,
+            major_version_only=major_version_only,
+            minor_version_only=minor_version_only,
+            current_obspacks=current_obspacks,
+            output_folder=output_folder,
+        )
+    elif version is None:
+        version = "v1"
+
+    obspack_name = cast(str, obspack_name)
```
brendan-m-murphy, line 1289: If `obspack_stub` is `None` and `obspak_name` is `None`, then `obspack_name` will still be `None` at this point, so it shouldn't be cast to `str`.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+    obspack_path: pathType | None = None,
+    subfolders: MultiSubFolder = None,
+    include_obs_type: bool = True,
+    include_version: bool = True,
+    data_version: str | None = None,
+    name_components: MultiNameComponents = None,
+    store: str | None = None,
+) -> list[StoredData]:
+    """
+    Based on the metadata associated with the retrieved data, create suitable obspack
```
brendan-m-murphy, line 724: "Add obspack filenames to StoredData objects in retrieved_data" might be a better first line. When I saw this function, my first question was "add the filenames to what?"

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+            filenames = [data.obspack_filename for data in data_group]
+            for additional_key in metakeys:
+                if check_unique(filenames):
+                    break
+                else:
+                    key_present = [additional_key in data.metadata for data in data_group]
+                    if np.all(key_present):
+                        name_components = name_components + [additional_key]
+
+                        # print(f"Checking new names with {additional_key}")
```
brendan-m-murphy, line 801: remove print statement

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+
+
+def default_release_files() -> list:
+    """
+    Release files which will be included in the created obspack by default.
+    This will return a list of filepaths to these default files.
+    """
+    from contextlib import ExitStack
+
+    file_manager = ExitStack()
```
brendan-m-murphy, line 957: You need to use `with file_manager: ...` or manually call `file_manager.close()` after you're done using it, otherwise the files aren't closed. 

Since there is only one context you need to enter into, you can just do
```
with importlib.resources.as_file(release_file_ref) as f:
    release_file_path = f
```

requirements.txt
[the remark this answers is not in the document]
brendan-m-murphy: You need to install setup tools in new virtual environments, so it could be missing for that reason. Does something fail CI if you don't include this requirement?

openghg/datapack/_obspack.py
```diff
@@ -277,30 +465,359 @@ def define_obspack_filename(
...
+        metadata: Metadata details associated with the stored data.
+        name_components: Current metadata keys used to define an obspack_filename.
+        store: Name of the object store to use when finding additional metakeys.
+    Returns:
+        list: Additional metakeys defined within the object store
+    """
+
+    from openghg.objectstore import get_metakeys
+
+    # TODO: How do we find out which bucket/store is being used? Is there a default if this is not specified?
```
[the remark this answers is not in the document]
brendan-m-murphy: I guess in this context it depends on what store the data has come from. I suppose you could make an option for passing the either a metakeys file or a list of potential keys to use to fix conflicts.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+        subfolders=subfolders,
+        include_obs_type=include_obs_type,
+        include_version=include_data_versions,
+        name_components=name_components,
+    )
+
+    site_detail_rows = []
+    for data in retrieved_data:
+        ds = data.data
+        ds.to_netcdf(data.obspack_filename)
```
brendan-m-murphy, line 1320: if you do `ds.to_netcdf(obspack_path / data.obspack_filename)`, then you won't need to pass `obspack_path` as an argument to many of the other functions, like `add_obspack_filenames`, `define_obspack_filenames`, `define_obspack_filename`

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+            search_df = read_input_file(search_filename)
+        else:
+            msg = "Either search_filename or search_df must be specified to create an obspack."
+            logger.exception(msg)
+            raise ValueError(msg)
+
+    retrieved_data = retrieve_data(search_df=search_df, store=store)
+
+    obs_types = search_df["obs_type"]
+    unique_obs_types = np.unique(obs_types)
```
brendan-m-murphy, line 1275: If you only need the unique obs types, you can use `search_df["obs_type"].unique()`

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+        pandas.DataFrame: Input file opened as a csv file using pd.read_csv(...)
+
+    TODO: Add checks for expected columns in input search file as required
+    """
+    search_df = pd.read_csv(filename)
+    return search_df
+
+
+def retrieve_data(
+    filename: pathType | None = None, search_df: pd.DataFrame | None = None, store: str | None = None
```
brendan-m-murphy, line 1031: I guess this should be `optionalPathType`? Although I prefer it the way it is.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+MultiSubFolder = dict[str, pathType] | pathType | None
+
+
+class StoredData:
+    """
+    This class contains details of the data extracted from the object store with details
+    of how this is related to the obspack this will be outputted to.
+    """
+
+    def __init__(
```
brendan-m-murphy, line 32: It seems like passing the subfolder here would simplify a lot of code (since you won't have to pass the subfolder info around). Also possibly "data version"?

openghg/datapack/_obspack.py
rt17603: Though I agree this could simplify things, one reason to include `obspack_path` as an input for these functions and not just include this in the `create_obspack` function would be to allow these functions to be called independently and still be able to define the complete filepath to where the file would be stored on disc.

I think this could actually shake out in a potential restructure as we could rely more on the `obspack_path` definition being readily available without having to pass around a variable (e.g. `obspack_path` is already a potential attribute on `StoredData` and could be used in more places). Would have to see.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+                store=store,
+            )
+
+            filenames = [data.obspack_filename for data in data_group]
+            for additional_key in metakeys:
+                if check_unique(filenames):
+                    break
+                else:
+                    key_present = [additional_key in data.metadata for data in data_group]
+                    if np.all(key_present):
```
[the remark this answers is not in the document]
rt17603: Good spot - I'll update that.

openghg/datapack/_obspack.py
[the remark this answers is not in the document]
brendan-m-murphy: Maybe use two different functions, like `make_obspack_filename` and `update_obspack_filename`, which would just use `make_obspack_filename` then set `obspack_filename` to the value.

openghg/datapack/_obspack.py
rt17603: Good spot - my original intention with this was to update this to a logger as I think storing the information that this process occurred could be useful to record. I'll update that.

openghg/datapack/_obspack.py
rt17603: Thanks for the feedback - I'll update this and the other part of the docstring you commented on.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+        data_version: Version of the data. If not specified and include_version is True this
+            will attempt to extract the latest version details from the metadata.
+        name_components: Keys to use when extracting names from the metadata and to use
+            within the filename. This can be specified per obs_type using a dictionary.
+            Default will depend on obs_type - see define_name_components().
+        store: Name of the object store to use if we need to find the config file for the data type.
+            This will be used if names are not unique to work out which keys to create a more
+            descriptive filename.
+    Returns:
+        list: Same list of StoredData objects passed to the function
```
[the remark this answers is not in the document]
rt17603: As commented on the other doc-string comment - I'll include a clarification for that.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+        name_components,
+        metadata=metadata,
+        name_suffixes=name_suffixes,
+        output_path=output_path,
+    )
+
+    return filename
+
+
+def find_data_version(metadata: dict) -> str | None:
```
brendan-m-murphy, line 418: Use this to add a method to `StoredData`, or use it in `StoredData.__init__` to add the data version there.

brendan-m-murphy (changes requested):
There are a few things that need to be fixed (the `obspack_name` issue in particular).

Again, I think it would benefit from some tidying, since the same arguments get passed around a lot, and there's a lot of disparate functionality in a single long file. I gave an example where you might avoid passing `obspack_path` to several functions, and a few other suggestions of functionality you might move into `StoredData`.

I think the code is more likely to be improved if it is done now, rather than adding an issue to do it later, unless there is a deadline that requires this functionality now.

rt17603:
> There are a few things that need to be fixed (the obspack_name issue in particular).
Was that the comment from the previous set of comments on `obspack_name` not being a string?

brendan-m-murphy:
> > There are a few things that need to be fixed (the obspack_name issue in particular).
> 
> Was that the comment from the previous set of comments on `obspack_name` not being a string? That should be fixed now.

Yes, sorry those all should have been on the review.

openghg/datapack/_obspack.py
rt17603: Yeah previously this did need `obs_types` as well but I can simplify that now since this has been updated.

openghg/datapack/_obspack.py
brendan-m-murphy: Yea one option would be to make that an attribute of `StoredData`, so the version of `define_obspack_filename` there doesn't need that argument. I mentioned some other changes like that as well (e.g. the subfolder of a `StoredData` object).

openghg/datapack/_obspack.py
rt17603: Yes, I know I made the point about using `pathType` before for consistency but I do agree this seems a bit nicer than `optionalPathType`.. I'm a bit torn! I'd be tempted to keep `pathType` but stop using `optionalPathType` and just use `pathType | None`?

openghg/datapack/_obspack.py
```diff
@@ -122,33 +126,55 @@ def define_get_functions() -> tuple[dict, dict]:
...
+            raise ValueError("When constructing names, separators must be >= depth of keys input.\n") from exc
+
+        # Generator expression to loop over elements
+        values = (_construct_name(key, dictionary, separators) for key in keys)
+        return separator.join(values)
+    else:
+        try:
+            value = str(dictionary[keys])
+        except KeyError:
+            raise ValueError("When constructing names, unable to find key within supplied dictionary")
```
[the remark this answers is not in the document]
rt17603: I'll add that

openghg/datapack/_obspack.py
brendan-m-murphy: Sure we can just replace `optionalPathType`.

openghg/datapack/_obspack.py
rt17603: Ok thanks for this.

So am I right that this is the suggested update for this?
```
    release_file_ref = importlib.resources.files("openghg") / "data/obspack/obspack_README.md"

    with importlib.resources.as_file(release_file_ref) as f:
        release_file_path = f

    release_files = [release_file_path]
    return release_files
```

Using what you included above it looks like the explicit creation of `file_manager` wasn't necessary and this still seems to run ok.

openghg/datapack/_obspack.py
rt17603: I've added `subfolder` and `data_version` as attributes for `StoredData` - at the moment they can still be supplied as an input for the other functions though. This is good groundwork for the discussed restructure though.

openghg/datapack/_obspack.py
rt17603: I've allowed `data_version` to be extracted from the metadata within `StoredData` now.

openghg/datapack/_obspack.py
rt17603: I guess this would be a tidy issue then rather than for this PR as it would impact other parts of the code?

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1330 @@
...
+logger = logging.getLogger("openghg.obspack")
+
+# TODO: Move to types submodule?
+ObsOutputType = ObsData | ObsColumnData
+NameComponents = list[str | list]
+MultiNameComponents = dict[str, NameComponents] | NameComponents | None
+MultiSubFolder = dict[str, pathType] | pathType | None
+
+
+class StoredData:
```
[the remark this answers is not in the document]
rt17603: I've added this as a comment on the parent issue so this can be tracked for the next version of this code - https://github.com/openghg/openghg/issues/1078

openghg/datapack/_obspack.py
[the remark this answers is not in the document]
rt17603: I'm going to leave this for now - I don't know if we'd want an explicit issue for this but we could look at making sure `get_metakeys` connects to the default internally if we have one.

openghg/datapack/_obspack.py
```diff
@@ -122,33 +126,55 @@ def define_get_functions() -> tuple[dict, dict]:
     return get_functions, get_fn_arguments
 
 
+def _construct_name(keys: list | str, dictionary: dict, separators: Sequence[str]) -> str:
```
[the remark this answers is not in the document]
rt17603: I'm going to leave this a direct constructor as I think it's a bit more clear about what it's doing but this could be a nice idea to move to - I've added the details to the parent issue #1078

openghg/datapack/_obspack.py
```diff
@@ -7,14 +7,86 @@
...
+        to include all output data types
+        """
+        self.stored_data = data
+        self.data = data.data
+        self.metadata = data.metadata
+
+        self.obs_type = obs_type
+        self.obspack_path = obspack_path
+
+        if isinstance(obspack_filename, str):
```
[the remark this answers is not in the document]
rt17603: For me, I'm not always sure compact syntax like this is the clearest to (mentally) parse.

In this case though, I think this seems clear enough so I've updated that and added some similar statements for other attributes.

openghg/datapack/_obspack.py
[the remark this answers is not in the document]
rt17603: I've updated this to use the two different functions as suggested.

openghg/datapack/_obspack.py
brendan-m-murphy: Sure, I don't think it matters much either way. I suspect there will be places where `pathType` isn't used, but maybe should be, etc.

openghg/datapack/_obspack.py
brendan-m-murphy: Okay seems like a good start. Something I've done in inversions is make the function with all the options start with "_", then make a method of the same name for a class that supplies a lot of the arguments.

openghg/datapack/_obspack.py
```diff
@@ -0,0 +1,1385 @@
...
+        # TODO: Decide how this could work with negative lookup?
+        kwargs = {key: value for key, value in kwargs.items() if not pd.isnull(value)}
+
+        # Pass any additional arguments needed for the get/search function
+        kwargs = kwargs | additional_fn_arguments
+
+        # TODO: Decide details around catching errors for no files/multi files found.
+        data_retrieved = get_fn(**kwargs)
+
+        DataRetrieved = StoredData(data_retrieved, obs_type=obs_type)
```
brendan-m-murphy, line 1140: This should adhere to "lower snake case" format, so maybe `stored_data` or `_data_retrieved`.

brendan-m-murphy, line 1140: If you could pass the `subfolders` info here, that would mean it doesn't need to be passed elsewhere. (Or maybe add the subfolder info in a separate step after this is called in `create_obspack`

brendan-m-murphy (approved):
Okay, looks good!

openghg/datapack/_obspack.py
rt17603: Sounds like a good convention - were you thinking that `define_obspack_filename` --> `_define_obspack_filename` in the code below then? This may mean a slight tweak to the current tests but that's probably ok for now.

openghg/datapack/_obspack.py
rt17603: Ah yes you're right - because this a list not a class itself:

```suggestion
        stored_data = StoredData(data_retrieved, obs_type=obs_type)
```

rt17603: I think for the moment I was trying to avoid having to pass `subfolders` to the `retrieve_data` function when this wasn't used in it.

openghg/datapack/_obspack.py
rt17603: I think for now I'll keep the naming as is (the methods are a different name now than the functions anyway) - but I think if we come up with conventions like that we should start recording them in the style-guide for developers so we can make sure to apply it consistently.

brendan-m-murphy (approved):
Looks good!

PR #1126, opened 2024-09-23
Iss1125 empty object store invalid

### LINKED ISSUE #1125: Object store becoming invalid when deleting the only Flux uuid present in this store
### What happened?

I have an object store which has only one source of flux and one source of boundary conditions. If I delete the uuid corresponding to flux data (which is the only key in data_manager(data_type='flux', store="user").metadata), with dm.delete_datasource(uuid=uuid), my object store becomes invalid and I can't add new flux data or access my boundary conditions anymore. 
This only happens when I have only one dataset of flux. When I delete data from another object store which has several types of fluxes, deleting one uuid is not an issue. I still have access to the remaining data.

The following error occurs when trying to search for BoundaryConditions
image

Note that my zarr store has been created in August, so should be in the new format..

For now, each time I want to delete something in an object store which has a single uuid per type of data (Flux, BoundaryConditions, ...), I have to create a new object store which is not ideal.

_No response_

### Relevant log output

_No response_

_No response_

### Environment

<details>

</details>

brendan-m-murphy:

Changed `LocalZarrStore.delete_all()` to delete the folder in the `data` dir associated with that zarr store. (That is, to delete `data/<datasource uuid>`, not just `data/<datasource uuid>/zarr`.)

It's possible that this should be handled by the `Datasource` instead. The way `Datasource` and `LocalZarrStore` interact at the moment, the path `data/<datasource uuid>` isn't explicitly used by `Datasource`. Instead, the datasource uuid is passed to `LocalZarrStore`, which decides where it should exist (i.e. in `data/<datasource uuid>/zarr`).

The datasource should probably know where its data is, and tell `LocalZarrStore` where to create the `zarr` directory, but that would be a bigger change. (Also, `LocalZarrStore` handles the locations of versions, which may or may not be what we want.)

Anyway, this fixes the problem of false negatives for the "zarr store check" in util/_user.py, from issue #1125, since that was caused by an empty folder like `data/<datasource uuid>`: the check says that `data/<datasource uuid>/zarr` doesn't exist, so it must not be a zarr store. The problem is that `data/<datasource uuid>` wasn't deleted when the datasource was deleted.

- [x] Closes #1125
- [x] Existing tests updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Bug where a datasource's folder in the `data` directory was not deleted by `Datasource.delete_all_data()`. This was causing `check_zarr_store` in `util/_user.py` to give a false negative [Issue #1125](https://github.com/openghg/openghg/issue/1125). [PR #1126](https://github.com/openghg/openghg/pull/1126) 
 - Bug where an input filepath list to standardise_surface was only storing the last file hash. This allowed for some files to bypass the check for the same files depending on where they were in the original filepath list. [PR #1100](https://github.com/openghg/openghg/pull/1100)
 - Bug where filepath needed to be a Path object when storing the file hash values. [PR #1108](https://github.com/openghg/openghg/pull/1108)
 - Catch an `AttributeError` when trying synchronise attributes and metadata and a user passes a `bool` - [PR #1029](https://github.com/openghg/openghg/pull/1029)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- Bug where a datasource's folder in the `data` directory was not deleted by `Datasource.delete_all_data()`. This was causing `check_zarr_store` in `util/_user.py` to give a false negative [Issue #1125](https://github.com/openghg/openghg/issue/1125). [PR #1126](https://github.com/openghg/openghg/pull/1126) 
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -19,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- Bug where a datasource's folder in the `data` directory was not deleted by `Datasource.delete_all_data()`. This was causing `check_zarr_store` in `util/_user.py` to give a false negative [Issue #1125](https://github.com/openghg/openghg/issue/1125). [PR #1126](https://github.com/openghg/openghg/pull/1126)
```
SutarPrasad, line 22: Bug where a datasource's folder in the `data` directory was not deleted by `Datasource.delete_all_data()`. This was causing `check_zarr_store` in `util/_user.py` to give a false negative. [PR #1126](https://github.com/openghg/openghg/pull/1126)

SutarPrasad (approved):
Just a small change in changelog, rest looks good

openghg/store/storage/_localzarrstore.py
```diff
@@ -259,7 +259,7 @@ def delete_all(self) -> None:
         """
         self._check_writable()
         self._stores.clear()
-        shutil.rmtree(self._stores_path, ignore_errors=True)
+        shutil.rmtree(self._stores_path.parent, ignore_errors=True)
```
rt17603, line 262: This may be overkill but do you think it may be worth adding a check that `self._stores_path.parent != self._bucket` or something? I can't see a situation where it would be needed, I'm just always a bit nervous about deleting directories especially when they're not direct but derived relative to another directory.

rt17603 (commented):
Looks good to me - I've added one small comment about an extra check before deleting the parent folder to discuss (though I think it may be overkill).

rt17603 (approved):
Looks good!

PR #1130, opened 2024-09-26
Adding option to get_obs_surface to skip the renaming of variables from species name to mf etc.

### LINKED ISSUE #1078: Umbrella: Adding in functionality to create an ObsPack

rt17603:

Within `get_obs_surface` function, this PR adds an additional `rename_vars` input to allow the renaming of the data variables to be bypassed if required.

This feeds into PR #1117 as this means we can use the nice feature of the `get_obs_surface` function to collate multiple inlets but still keep the species naming to help with creating an expected obspack output.

- [x] Related to #1078
- [x] **No tests added** but they do pass [Tests passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature
- [x] Documentation ~and tutorials~ updated/added

Not needed


openghg/retrieve/_access.py
```diff
@@ -410,24 +415,25 @@ def get_obs_surface_local(
...
-        if "status_flag" in var:
-            rename[var] = "status_flag"
-        if "integration_flag" in var:
-            rename[var] = "integration_flag"
-
-    data = data.rename_vars(rename)  # type: ignore
+    if rename_vars:
+        rename: dict[str, str] = {}
+
+        data_variables = [str(v) for v in data.variables]
```
SutarPrasad, line 421: We should employ pattern matching here, instead of conditional statements. 
for e.g variable_patterns = {"repeatability":"mf_repeatability", "variability":"mf_variability", ...} and loop and break for match. This will ensure we are increasing readability and scalability in future if required, whilst not repeating.

SutarPrasad (changes requested):
Apart from this suggestion everything else looks good

openghg/retrieve/_access.py
rt17603: I wouldn't want to make that change on this PR as this is just to allow these steps to be skipped but makes no other changes. Updating how the data variables names are updated should really be a separate issue.

SutarPrasad: I understand we add a check and ensure the steps are skipped, However, losing track of such code-refining changes is easier. Thus if changes like these are small and don't impact the functionality hugely, it is worth making the change.

rt17603: If you wanted to put in a separate PR for this you'd be welcome to but I think we'd be best to keep this PR simple and within the scope of the original idea.

SutarPrasad: Resolving and referenced in https://github.com/openghg/openghg/issues/1141

SutarPrasad (approved):

PR #1133, opened 2024-09-27
search_surface formats dictionary of data levels

### LINKED ISSUE #1132: ICOS `retrieve_atmospheric` raising an error due to `format_data_levels`
### What happened?

Using `openghg.retrieve.icos.retrieve_atmospheric` always raises a `MetadataFormatError` because it is passing a dictionary to `format_data_level`, but `format_data_level` doesn't accept dictionaries.

Probably to get an authentication error from ICOS ;)

_No response_

### Relevant log output

_No response_

_No response_

### Environment

<details>

</details>

brendan-m-murphy:

It is possible to pass a dictionary to any of the parameters in `search_surface` (or any `search_` functions), so the formatting functions we use must be able to format dictionaries. 

I fixed the specific case of `data_levels`. Formatting done in `_base_search` should work with dictionaries, since this formatting either uses `format_inlet`, which is overloaded to work with dictionaries and lists, or it applies `clean_string`, using cases for lists and dicts.

Note: I had to add a `type: ignore` for the dictionary check because passing a dictionary for `data_level` is not allowed by the signature of `search_surface`. However, this is a problem for all of the arguments in the signature of `search_surface`, so I think we need a more general fix. (Or just use dynamic typing here, since dicts are handled by `_base_search`.)

- [x] Closes #1132

### DIFF
--- a/tests/retrieve/test_search.py
+++ b/tests/retrieve/test_search.py
@@ -247,12 +247,13 @@ def test_many_term_search():
 
 
 def test_optional_term_search():
-    """Test search using dict inputs. This should create an OR search between the key, value pairs in the dictionaries"""
+    """Test search using dicst inputs. This should create an OR search between the key, value pairs in the dictionaries"""
     # Note: had to be careful to not create duplicates as this currently raises an error.
     res = search(
         site="bsd",
         inlet_option={"inlet": "42m", "height": "42m"},
         name_option={"station_long_name": "bilsdale, uk", "long_name": "bilsdale"},
+        data_level={"icos_data_level": 2, "data_level": "not_set"},
     )
 
     assert len(res.metadata) == 3
--- a/tests/retrieve/test_search.py
+++ b/tests/retrieve/test_search.py
@@ -247,7 +247,7 @@ def test_many_term_search():
 
 
 def test_optional_term_search():
-    """Test search using dicst inputs. This should create an OR search between the key, value pairs in the dictionaries"""
+    """Test search using dict inputs. This should create an OR search between the key, value pairs in the dictionaries"""
     # Note: had to be careful to not create duplicates as this currently raises an error.
     res = search(
         site="bsd",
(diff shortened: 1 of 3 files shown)


tests/retrieve/test_search.py
```diff
@@ -247,12 +247,13 @@ def test_many_term_search():
 
 
 def test_optional_term_search():
-    """Test search using dict inputs. This should create an OR search between the key, value pairs in the dictionaries"""
+    """Test search using dicst inputs. This should create an OR search between the key, value pairs in the dictionaries"""
```
rt17603, line 250:
```suggestion
    """Test search using dict inputs. This should create an OR search between the key, value pairs in the dictionaries"""
```

tests/retrieve/test_search.py
```diff
@@ -247,12 +247,13 @@ def test_many_term_search():
...
     # Note: had to be careful to not create duplicates as this currently raises an error.
     res = search(
         site="bsd",
         inlet_option={"inlet": "42m", "height": "42m"},
         name_option={"station_long_name": "bilsdale, uk", "long_name": "bilsdale"},
+        data_level={"icos_data_level": 2, "data_level": "not_set"},
```
rt17603, line 256: Just to check - is this included here to check this can be run with a dictionary input at all? Which is why we're not expecting this to change the assertions?

rt17603 (dismissed):
Looks good to me so I've added an Approve with one typo fix and a small question. Let me know if you need a re-approval for any reason.

tests/retrieve/test_search.py
brendan-m-murphy: I checked and the data we find has "not_set" for data level, so this search should work. None of the data we find has "icos_data_level". So some extra searches are done, but they don't find anything.

But yes, the main point was to check if passing a dictionary for `data_level` works at all.

rt17603 (dismissed):
Looks good

rt17603 (approved):

PR #1138, opened 2024-10-01
Iss986 sync surface metadata to store level

SutarPrasad:
Shifted a synch_surface_metadata from assign_attributes to store level in ObsSurface.read_file.
Also reduced the metadata from the `expected` sections of the parser test
Removed parse_surface_metachecker from parser
Added obssurface_metachecker and attributes_checker to the standardised data.

- [x] Partially completes #986 (Replace xxxx with the Github issue number)


tests/store/test_obssurface.py
```diff
@@ -1020,3 +1055,33 @@ def test_optional_metadata():
...
+                         ])
+def test_sync_surface_metadata_store_level(filepath, site, instrument, sampling_period, network, inlet, store,
+                               source_format, update_mismatch, caplog):
+    clear_test_stores()
+    bucket = get_writable_bucket(name="user")
+
+    standardised_data = standardise_surface(filepath=filepath, site=site, instrument=instrument, sampling_period=sampling_period, network=network,
+                        inlet=inlet, store=store, source_format=source_format, update_mismatch=update_mismatch)
+
+    standardised_data = standardised_data["processed"][str(filepath).split("/")[-1]]
```
rt17603, line 1080: I believe you should be able to achieve the same result using pathlib directly to get the `.name` attribute - https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.name

```suggestion
    standardised_data = standardised_data["processed"][filepath.name]
```

tests/store/test_obssurface.py
```diff
@@ -1020,3 +1055,33 @@ def test_optional_metadata():
     rgl_ch4_metadata = rgl_ch4.metadata
 
     assert "project" in rgl_ch4_metadata
+
+
+@pytest.mark.parametrize("filepath, site, instrument, sampling_period, network, inlet, store, source_format, update_mismatch",
```
rt17603, line 1060: Great to use parameterisation in these kinds of cases to make it easier to check out multiple scenarios but it can end up with a lot of inputs as you've found here.

One small suggestion to reduce the length of these inputs a little would be to just include a "filename" and create the filepath within the function using the source_format which is also passed. You could also just specify the store within the test as "user" is being used when getting the bucket e.g.
```
@pytest.mark.parametrize("filename, site, instrument, sampling_period, network, inlet, source_format, update_mismatch",
                         [
                             ("DECC-picarro_TAC_20130131_co2-185m-20220928.nc",
                              "tac", "picarro", "1h", "decc", "185m", "openghg", "from_definition"),

                             ("ch4_bao_tower-insitu_1_ccgg_all.nc",
                              "bao", None, None, "noaa", None, "noaa", "from_source"),

                             ("ICOS_ATC_L2_L2-2024.1_RGL_90.0_CTS.CH4",
                              "rgl", "g2301", None, "icos", None, "icos", "never")
                         ])
def test_sync_surface_metadata(filename, site, instrument, sampling_period, network, inlet,
                               source_format, update_mismatch):
    """
    Test metadata and attribues are correctly synced.
    """
    clear_test_stores()

    store = "user"
    bucket = get_writable_bucket(name=store)

    filepath = get_surface_datapath(filename, source_format)

...
```

rt17603 (commented):
Looks good - we may end up deciding to split out the code into more helper functions later on but for now this is a step towards reducing what needs to be included within the parser.

A few small comments included.

openghg/store/_obssurface.py
```diff
@@ -363,6 +364,20 @@ def read_file(
                 for key, value in data.items():
                     data[key]["data"] = value["data"].chunk(chunks)
 
+            for _, gas_data in data.items():
```
rt17603, line 367: I think it would be worth making this added section a method on the `ObsSurface` class e.g. `align_metadata_attributes(...)`

In future would be good to create a more generic function for this functionality on the `BaseStore` class which would then be applied to all of the data type classes. Splitting this out now should make this easier to use as a blueprint when we're able to do this.

rt17603 (commented):
One final suggested change and then this should be able to go in.

We're going to have to think about how best to test that the metadata and attributes are being created correctly. We will likely need to start improving the checks at the store level to do this.

openghg/store/_obssurface.py
```diff
@@ -683,3 +686,33 @@ def seen_hash(self, file_hash: str) -> bool:
 
     def set_hash(self, file_hash: str, filename: str) -> None:
         self._file_hashes[file_hash] = filename
+
+    @staticmethod
```
rt17603, line 690: Can I ask if there a reason why this needs to be a static method do you think? For `validate_data` and `scheme` these were static methods so they could be called on the class without needing to create a specific instance but I don't think that would the case for this method?

rt17603 (commented):
Have just included a question on whether we need the new method to be a static method or whether this would be better as a normal method (i.e. no decorator and called with `self.`)?

openghg/store/_obssurface.py
SutarPrasad: It functions as a utility for the class so created it as a staticmethod.

brendan-m-murphy (commented):
It doesn't seem like there is anything "surface specific" in this code. Perhaps `align_metadata_attributes` could be moved to `BaseStore` and `sync_surface_metadata` could just be `sync_metadata` (since the `metadata` and `attributes` could really be any dictionaries).

openghg/store/_obssurface.py
brendan-m-murphy: I guess `self` would be an unused argument in this function, so a staticmethod makes sense. Inside `ObsSurface`, you can use `self.my_static_method` instead of `ObsSurface.my_static_method`.

rt17603:
> It doesn't seem like there is anything "surface specific" in this code. Perhaps `align_metadata_attributes` could be moved to `BaseStore` and `sync_surface_metadata` could just be `sync_metadata` (since the `metadata` and `attributes` could really be any dictionaries).

I agree most is generic but the `metadata_default_keys()` function used within `sync_surface_metadata()` which is surface specific at the moment - 
https://github.com/openghg/openghg/blob/3f2a31789bb704dbb1bacf194f9519cbcd004291/openghg/standardise/meta/_metadata.py#L160 . So we'd need to make some attribute definitions for the other data type classes before this could be made fully generic.

openghg/store/_obssurface.py
rt17603: I don't think this particular function as written would be intended for use outside of a specific instance of the class in this way e.g. we wouldn't expect to call `ObsSurface.align_metadata_attributes(data)` (especially at the moment given that `data` is a more complicated object). We may later on want to pull some of the details out of this but I don't think we'd use the method in this form as a utility function.

I don't think that not using `self` at the moment within the function necessarily means this would be better as a static method?

rt17603 (changes requested):
For the method on the `ObsSurface` class I think we should make this a normal method (with `self` even though its not used) rather than a static method.

For reference the way static methods have been used in the past were largely as a way to expose a type specific method without needing to create an instance of that class. I don't think this function would fall into that category as a user would not be likely to use it especially with the internal `data` input we're currently expecting. I'm not sure if that's how they're always used but that's the way we've largely been using them within openghg. I also think it may be that this method would end up using `self` as some point as we continue to make changes.

The alternative would be creating this as a function within the util module but as we're still dealing with this more complex expected input for `data` (i.e. the nested dictionary form) I think that would be a bit of strange input to expect for a util function as its not very generic so for now at least I think it would still be better associated with the data type class.

Issue #1116, opened 2024-09-06
Standardise local ICOS files


joe-pitt:
### What is your issue?

Currently when you use standardise_surface with source_format="ICOS" it assumes a filename of the form mhd.ch4.hourly.g2401.15m.dat. This is deprecated - now all the files have names like "ICOS_ATC_L2_L2-2024.1_RGL_90.0_CTS.CH4". So a few tweaks are needed...

joe-pitt:
Closed by PR #1140

PR #1140, opened 2024-10-03
Iss1116/icos fnames

joe-pitt:
Update the parser for standardising data from local ICOS ASCII files to reflect the change in format since this was first written. We rarely need to standardise these files, because usually we use the retrieve function to get ICOS data straight from the ICOS CP, but for PARIS I had to load a set of local files from Stephan that are in the ICOS ASCII format.

Closes Issue #1116


joe-pitt:
Currently failing because:
```
openghg/standardise/surface/_icos.py:69: error: Argument "inlet" to "_read_data_large_header" has incompatible type "str | None"; expected "str"  [arg-type]
openghg/standardise/surface/_icos.py:79: error: Argument "inlet" to "_read_data_small_header" has incompatible type "str | None"; expected "str"  [arg-type]
openghg/standardise/surface/_icos.py:251: error: Incompatible types in assignment (expression has type "float", target has type "str | None")  [assignment]
```

SutarPrasad (dismissed):
Looks good, needs changelog. Approved

Issue #881, opened 2024-01-29
Adding data - implementupdate "replace"/"insert" case for if_exists functionality


rt17603:
### What is your issue?

See #591 for original details on this.

For the new `if_exists` input flag we want to up

rt17603:
PR #1166 is working to address this issue, including laying groundwork to support this.

For this the suggestion is to approach this with 3 goals in mind that build up this functionality:
1. Update version handling (neutral, restructure change)
2. Formalise and update index comparison (may be neutral or may include updated functionality)
3. Build on this to add the new mode for combining data

rt17603:
Part A completed: #1356

rt17603:
For Part B: #1391 this looks to add an "insert", "update" and "upsert" methods on a `Store` class. Details of how this is related to the expected front end methods for users should be included here.

brendan-m-murphy:
> Next steps in Part D: [#1532](https://github.com/openghg/openghg/pull/1532)
> 
> * adding `if_exists="update` mode (correct?)

Part D #1532 adds the ability to update data in a `ZarrStore`, but this hasn't been exposed via `Datasource` yet.

My plans for the next parts are:
- Part E: update `Datasource` to use new update methods (this could technically close the umbrella)
- Part F: refactor to remove old `LocalZarrStore` class and use new `ZarrStore` class directly in `Datasource`

We might want to change how the flags in `standardise_` work.

Also, we might want a new tutorial, or add more details to existing tutorials.

There are also lingering issues like controlling how attributes are updated (currently attributes are overwritten whenever we add new data, but we combine attributes during standardising, so I think we probably keep the attributes consistent that way).

PR #1150, opened 2024-10-11
Quick (and unpleasant) fix for force keyword to stop this always creating a new version

rt17603:

The `force` keyword is used to allow users to add data identical to what has been seen before. At the moment the `force` keyword always sets `if_exists="new"` to ensure that a new version is created and an error is raised because the data is already present. However, the problem is when there are multiple files to be added which have all been seen before this creates a new version every time and there's no way to stop this.

This is a quick fix to not set `if_exists="new"` but to include a warning that this may result in a DataOverlapError. This also includes the advice to use `if_exists="new"` for the first file and to not use this keyword (or use `if_exists="auto"`) afterwards. It's definitely not the nicest solution but would get past the issue of multiple files being added which have been seen before.

Note: we're still missing what is becoming a key mode which allows data to be combined with data even when there is an overlap (Issue #881 ) which would fill in some of the missing scenarios we have at the moment for adding new data.

### DIFF
--- a/openghg/store/_boundary_conditions.py
+++ b/openghg/store/_boundary_conditions.py
@@ -136,11 +136,14 @@ class BoundaryConditions(BaseStore):
                 "Overwrite flag is deprecated in preference to `if_exists` (and `save_current`) inputs."
                 "See documentation for details of these inputs and options."
             )
-            if_exists = "new"
+            if_exists = "new"  # TODO: Update to "combine" when available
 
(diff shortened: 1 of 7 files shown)


rt17603:
@rt17603 - update this to include on BaseStore.read_file instead.

Add details to Issue #1380 on need to update `if_exists="new"` --> `if_exists="combine"` once this option becomes available.

Issue #1121, opened 2024-09-13
get_obs_surface Returning Empty Data Due to Unused Variables with NaN Values


hdelongueville:
### What is your issue?

The function get_obs_surface is returning data with no values in it. This issue is caused by the presence of variables filled with NaN values, which are not used in the process.

For example, in the case where there is a variable sttb full of NaNs, all the values are dropped because of that.

A quick fix is to set keep_missing=True, to skip the step that drops the NaNs. What is the best long term solution though?

joe-pitt:
Here's a suggestion for dealing with this within inversions:
1. Add capability to use things other than mf_repeatability in the model-data mismatch uncertainty calculation (see Issue #1120 )
2. Drop any variables that aren't being used prior to dropping the nans (within inversion code)

rt17603:
> The function get_obs_surface is returning data with no values in it.

From this it also sounds like there's a second issue as well around `get_obs_surface` returning empty data which may not be helpful. As an alternative, this something that could be checked for and an error raised rather than returning the data? Would that be preferable?

joe-pitt:
> 
> ```
> check_nan_subset = ["mf", "mf_repeatability", "mf_variability", ...]
> ...
>         ds_resampled = ds_resampled.dropna(dim="time", subset=check_nan_subset )
> ```
> 
> And we could make it so that `subset` could be specified by the user with some useful default as shown above.
> 
> The question would be: what defaults would be sensible for this and cover enough cases?

This sounds like a good plan to me. The default should definitely include `mf` - maybe that is the only one that is absolutely essential?

rt17603:
Did anyone have a quick example of the data that caused this problem by the way?

rt17603:
Would it be possible to get an example of this data? Would be useful to allow checks to be added for these issues.

hdelongueville:
I currently only have access to the object store. @joe-pitt, think this might be something you could help with?

joe-pitt:
An example would be the files in: /group/chem/acrg/obs_raw/EYE-AVE-PAR/EYE-AVE-PAR_2.2. Many of these have nans for things like LTR, STTB and Unc_n2o. At the moment the icos standardise function only loads unc_n2o (at one point I experimented with adding the others, hence sttb is mentioned in original post). See also this issue: https://github.com/openghg/openghg_inversions/issues/212

rt17603:
[...]

PR #1153, opened 2024-10-14
Adding in final check to get_obs_surface_local to check size of time dimension

rt17603:

Within Issue #1121 , this showed a corner case where data can be returned which contains no entries on the time dimension if there is one variable in the data which contains NaN values. This is because after retrieving the data from the object store the NaN values will be dropped by default.

This is fix 1/2 for this Issue to make sure a `SearchError` is raised if we're trying to return data after the resampling and filtering which contains no data.

TODO:
 - [ ] Get example of data and set up which causes this error and add test

- [ ] Fixes 1/2 updates from  #1121 

Not needed:

### DIFF
--- a/openghg/retrieve/_access.py
+++ b/openghg/retrieve/_access.py
@@ -458,6 +458,11 @@ def get_obs_surface_local(
     #     )
     #     print("Suggestion: set calibration_scale to convert scales")
 
+    if obs_data.data["time"].size == 0:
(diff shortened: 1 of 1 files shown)


rt17603:
This may not be relevant anymore? Check and close if so - otherwise finish off.

rt17603:
Should find some data which previously triggered this problem and see if this fix is still needed.

PR #1164, opened 2024-10-31
1161 improve standardisation of multiple footprints at once

### LINKED ISSUE #1161: Improve standardisation of multiple footprints - monotonic error

Some footprints have some small error (order of 1e-8) in their stored longitude and latitude coordinates. When those errors are not the same across the footprints for a folder, this makes it impossible to standardise the folder all at once (e.g. `/group/chemistry/acrg/LPDM/fp_NAME/EUROPE/LIN-98magl/inert/`). The files have to be standardised one by one which seems less efficient (and less user friendly).

Standardisation attempt : 
```
from glob import glob
from openghg.standardise._standardise import standardise_footprint

standardise_footprint(filepath=glob(f'/group/chemistry/acrg/LPDM/fp_NAME/EUROPE/LIN-98magl/inert/*'),
                      site= 'LIN',
                      inlet= '98magl',
                      model= 'NAME',
                      met_model= 'UMG',
                      domain= 'EUROPE',
                      species='inert',
                      source_format= 'paris',
                      store= 'ad_test_zarr',
                      period= None,
                      continuous= True)
```

Error message : 
```
Traceback (most recent call last):
  File "/user/home/bq24992/workingDir/PARIS/Footprints/test.py", line 24, in <module>
    standardise_footprint(**kwargs)
  File "/user/home/bq24992/openghg/openghg/standardise/_standardise.py", line 621, in standardise_footprint
    return standardise(
           ^^^^^^^^^^^^
  File "/user/home/bq24992/openghg/openghg/standardise/_standardise.py", line 51, in standardise
    result = dc.read_file(filepath=filepath, **kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/user/home/bq24992/openghg/openghg/store/_footprints.py", line 371, in read_file
    footprint_data = parser_fn(**parser_input_parameters)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/user/home/bq24992/openghg/openghg/standardise/footprints/_paris.py", line 69, in parse_paris
    fp_data = xr_open_fn(filepath)
              ^^^^^^^^^^^^^^^^^^^^
  File "/user/home/bq24992/.local/lib/python3.12/site-packages/xarray/backends/api.py", line 1084, in open_mfdataset
    combined = combine_by_coords(
               ^^^^^^^^^^^^^^^^^^
  File "/user/home/bq24992/.local/lib/python3.12/site-packages/xarray/core/combine.py", line 958, in combine_by_coords
    concatenated_grouped_by_data_vars = tuple(
                                        ^^^^^^
  File "/user/home/bq24992/.local/lib/python3.12/site-packages/xarray/core/combine.py", line 959, in <genexpr>
    _combine_single_variable_hypercube(
  File "/user/home/bq24992/.local/lib/python3.12/site-packages/xarray/core/combine.py", line 645, in _combine_single_variable_hypercube
    raise ValueError(
ValueError: Resulting object does not have monotonic global indexes along dimension longitude
```

My suggestion is to make sure that the coordinates are perfectly aligned before trying to open them with `xarray.open_mfdataset`.

For example we could replace https://github.com/openghg/openghg/blob/3eca7392b0f170508509be2fbad46a6eaec2dd86/openghg/standardise/footprints/_paris.py#L70C1-L71C1 by something like : 
```

    try:
        fp_data = xr_open_fn(filepath)
    except ValueError as v :
        if not 'Resulting object does not have monotonic global indexes along dimension' in str(v):
            raise v

        import xarray as xr
        import numpy as np
        from openghg.util import align_lat_lon
        new_filepath = list()
        for file_name in filepath:
            with xr.open_dataset(file_name) as ds:
                ds_tmp = ds.rename({'longitude':'lon',
                                    'latitude':'lat'})
            lon,lat = ds_tmp.lon.values,ds_tmp.lat.values
            ds_realigned = align_lat_lon(ds_tmp,domain)
            lon_domain,lat_domain = ds_realigned.lon.values,ds_realigned.lat.values
            if np.sum(lon-lon_domain)!=0 or np.sum(lat-lat_domain)!=0 :
                split_file_name = file_name.split('/')
                split_file_name[-1] = 'tmp_'+split_file_name[-1]
                tmp_file_name = '/'.join(split_file_name)
                ds_realigned.rename({'lon':'longitude',
                                     'lat':'latitude'}
                                     ).to_netcdf(tmp_file_name)
                new_filepath.append(tmp_file_name)
            else :
                new_filepath.append(file_name)
        filepath = new_filepath 
        fp_data = xr_open_fn(filepath)
        pass
```

In the example, every file is realigned and stored because they does not perfectly aligned with the referenced coordinates from the domain. However, they align between them except for one, so one other option would be to realign along the coordinated of the first file of the list (thus only one file should be stored in a tmp file).

Those solutions may not reduce the time of processing compared to processing the file one by one (haven't checked though). But may be simpler from a user perspective.

Let me know your thoughts so that I may go on and make (or not) some changes to the code.

### Additional context

_No response_

alexdanjou:
Align the dataset(s) when opening the data while standardising footprints data to prevent error due to misalign coordinates.

- [X] Closes #1161 


alexdanjou:
Checking why the tests don't pass and coming back to you asap

qq23840 (approved):
All looks good to me

openghg/util/_file.py
```diff
@@ -247,27 +248,31 @@ def get_logfile_path() -> Path:
         return Path("/tmp/openghg.log")
 
 
-def check_function_open_nc(filepath: multiPathType) -> Tuple[Callable, multiPathType]:
+def open_and_align_dataset(filepath: multiPathType, domain: str) -> Tuple[xr.Dataset, multiPathType]:
```
rt17603, line 251: Rather than replacing this function (which is used more widely than just for footprints), it would be good to create this as a separate function as an alternative option for footprints to use.

Updating this to restore the `check_function_open_nc` may also go some way towards getting past the `mypy` issues which are coming up as well.

rt17603 (changes requested):
Overall looks like a useful change for this - my main comment would be in creating `open_and_align_dataset` as a new function rather than replacing `check_function_open_nc`. 

That may also go some way to helping with the mypy checks which are failing (I can assist with working those out if they don't all go away though).

openghg/util/_domain.py
```diff
@@ -337,14 +337,22 @@ def align_lat_lon(data: XrDataLikeMatch, domain: str) -> XrDataLikeMatch:
...
+        data: spatial data to be checked. Must have 'lat'/'latitude' and 'lon'/'longitude' dimensions
         domain: domain in question. Must be a valid domain in openghg_defs
 
     Returns:
         xarray.DataArray / xarray.Dataset: data with lat, lon ranges aligned to openghg domain.
     """
-
-    data = check_coord_alignment(data=data, domain=domain, coord="lat")
-    data = check_coord_alignment(data=data, domain=domain, coord="lon")
+    dims = list(data.sizes.keys())
```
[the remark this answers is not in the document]
brendan-m-murphy: You can probably just use `dims = data.dims`. Right now `data.dims` is a dictionary, but I think it will change to tuple in the future. But e.g. `time in {"time": 1234}` will return true.

openghg/util/_file.py
brendan-m-murphy: To keep the same behaviour, you could use e.g. `xr_open_fn = partial(xr.open_mfdataset, preprocess=lambda x: align_lat_lon(x, domain))`.

On the other hand, is there a reason for returning the function to call and the filepath(s) separately? It might make sense to just do the final `xr_open_fn(filepath)` in this function.

alexdanjou: Are you sure that `check_function_open_nc` is used elsewhere @rt17603 ? I cannot find it anywhere else when I search for it, that's why I suppressed it.

rt17603: Yeah you're right - it doesn't look like it is being used elsewhere at the moment - likely because it was put in when the footprint functions were being created.

However, with a nice generic function like this it would be useful to keep this broad as we will likely move to using it in other places (e.g. flux and boundary conditions would benefit from this) so would be good to keep as 2 separate functions.

alexdanjou: So for me the function `check_function_open_nc` is only used her https://github.com/openghg/openghg/blob/5eea4a4eb88a8581e44699cc3dd9ff31a9874951/openghg/standardise/footprints/_paris.py#L68 and here https://github.com/openghg/openghg/blob/5eea4a4eb88a8581e44699cc3dd9ff31a9874951/openghg/standardise/footprints/_acrg_org.py#L68 
(see https://github.com/search?q=repo%3Aopenghg%2Fopenghg%20check_function_open_nc&type=code)
But if you confirm @rt17603 that it need to be kept, I'll put it back.

The only purpose for me to use `check_function_open_nc` is to avoid `mypy` complaining at this stage. 
As we need to align the data in the case when calling `open_mfdataset`, I think it is better to align also the data while opening them when using `open_dataset` (for consistency).

Thus, another possibility would be to change `check_function_open_nc` to :
```
def check_function_open_nc(filepath: multiPathType, domain: str) -> Tuple[Callable, multiPathType]:
    """
    Check the filepath input to choose which xarray open function to use:
     - Path or single item List - use open_dataset
     - multiple item List - use open_mfdataset

    Args:
        filepath: Path or list of filepaths
        domain: domain of dataset
    Returns:
        Callable, Union[Path, List[Path]]: function and suitable filepath
            to use with the function.
    """
    import xarray as xr
    from functools import partial
    from openghg.util import align_lat_lon

    if isinstance(filepath, list):
        if len(filepath) > 1:
            xr_open_fn : Callable = partial(xr.open_mfdataset, preprocess=lambda x: align_lat_lon(x, domain))
        else:
            xr_open_fn = lambda x : align_lat_lon(xr.open_dataset(x), domain)
            filepath = filepath[0]
    else:
        xr_open_fn = lambda x : align_lat_lon(xr.open_dataset(x), domain)

    return xr_open_fn, filepath
    ```
    This seems to be ok from `mypy` point of view.

Comment (alexdanjou) [reply openghg/util/_file.py:251]: Ah just saw your answer @rt17603 sorry. I'll keep them separated then

Comment (brendan-m-murphy) [reply openghg/util/_file.py]: The alternative in your comment looks good. Maybe we can have `open_and_align_data` as a function that calls `check_function_open_nc` and then returns `xr_open_fn(filepath)`? Then we could use `open_and_align_data` in `read_file` and so on?

Also, does `align_lat_lon` work if there are no lat/lon coords? Otherwise I think we need to distinguish what happens by data type.

**Review by brendan-m-murphy — COMMENTED** [review]
Your changes are good, the only issue is that we planned to all lists of files to be passed for any data type, but haven't yet implemented that. So there is sort of a hidden criteria here that the solution works for all data types.

I think either:
1) we use your PR as is, then make modifications later when we extend this feature to all data types
2) the ability to standardise multiple files at once is added to all data types first, then your PR is incorporated after that.

I think this is partially up to @rt17603. My vote is for option 1, since we don't know exactly what we need to extend this to all other types. (Maybe just add a comment to the function that calls `align_lat_lon` saying that this won't work for data that doesn't have lat and lon coords.)

**Comment by rt17603** [comment]
> Your changes are good, the only issue is that we planned to allow lists of files to be passed for any data type, but haven't yet implemented that. So there is sort of a hidden criteria here that the solution works for all data types.
> 
> I think either:
> 
>     1. we use your PR as is, then make modifications later when we extend this feature to all data types
> 
>     2. the ability to standardise multiple files at once is added to all data types first, then your PR is incorporated after that.
> 
> 
> I think this is partially up to @rt17603. My vote is for option 1, since we don't know exactly what we need to extend this to all other types. (Maybe just add a comment to the function that calls `align_lat_lon` saying that this won't work for data that doesn't have lat and lon coords.)
> 
> Option 2 might not be too hard to carry out, but I think the `Flux.transform` case will be tricky (we could defer it though).

Yes I agree with this - I've been playing around with an alternative that matches to what you suggested in your comment:
> ..."fp_data = open_and_align_dataset(filepath, domain) and open_and_align_dataset calls check_function_open_nc and then returns xr_open_fn(filepath)"...

But it perhaps doesn't seem reasonable to impose that on this PR as it's fixing the current issue for this already. Shall we go with Option 1 and I'll put in another issue (which can hopefully be resolved quickly)?

File: openghg/util/_file.py [hunk 3]
```diff
@@ -247,27 +248,59 @@ def get_logfile_path() -> Path:
...
+#             filepath = filepath[0]
+#             ds = xr.open_dataset(filepath) # type: ignore
+#             ds = align_lat_lon(ds, domain)
+#     else:
+#         ds = xr.open_dataset(filepath) # type: ignore 
+#         ds = align_lat_lon(ds, domain)
+
+#     return ds, filepath
+
+def check_function_open_nc(filepath: multiPathType, domain: str) -> Tuple[Callable, multiPathType]:
```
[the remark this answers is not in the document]
rt17603: Yes I think either of those suggestions would be good - I've opened an Issue #1169 so we can try and implement this.

brendan-m-murphy (dismissed):
I think this is good to go in. @alexdanjou can you fix the merged conflicts and see if the tests pass before merging? Let me know if you need me to approve again.

openghg/util/__init__.py
```diff
@@ -31,6 +31,7 @@
     load_standardise_parser,
     load_transform_parser,
     read_header,
+    # open_and_align_dataset,
```
brendan-m-murphy, line 34:
```suggestion
```

brendan-m-murphy (approved):
Looks good! We should be able to use this with the other data types too. I suggested one inline fix. You should be able to commit it without needing another review, but let me know.

brendan-m-murphy (approved):
Looks good!

PR #1168, opened 2024-11-05
Added check for file lock permissions

### LINKED ISSUE #1167: Metastore lock files have wrong permissions
## What happened?

        I added fluxes to the shared zarr store, and later Alex was unable to add fluxes to the same store due to a permission error.
        The Flux metastore itself had rw permissions for the group, but 
                the lock for the metastore only had read permissions for
                the group.

## What did you expect to happen?
        If someone has write permissions for a metastore, then they should
                be able to acquire the lock for that metastore.

brendan-m-murphy:
* **Summary of changes** 

If a user does not have the right permissions, an error is raised with a suggested solution.

When the lock is first created, it has rw permissions for user and group.

- [x] Closes #1167 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


rt17603 (approved):
Looks good to me! Thanks for putting that in.

Issue #1075, opened 2024-08-02
Replace nested metadata/data dictionaries with list of MetadataAndData objects


brendan-m-murphy:
### What is your issue?

Many functions related to standardising data pass around a dictionaries whose keys are not used and whose values are dictionaries that always contain metadata and data keys.

Replacing this nested dictionary with a list of objects that satisfy the `HasMetadataAndData` protocol would more clearly express what is in these objects.

SutarPrasad:
https://github.com/openghg/openghg/issues/1105 can be the covered in this issue

rt17603:
We discussed a few weeks ago now but a first step initial step towards (and to make the update to `HasMetadataAndData` objects a bit smoother) - we could add a update to update the "standardised" format returned from parse functions from a dict to a list e.g. move from

The current format returned from `parse_*` functions takes the form:

{KEY1: {"data": ..., "metadata": ..., "attributes": ...}, KEY2: {"data": ..., "metadata": ..., "attributes": ...}, ...}

to

[{"data": ..., "metadata": ..., "attributes": ...}, {"data": ..., "metadata": ..., "attributes": ...}, ...]

Ultimately this could then be replaced with:

[Object1, Object2, ...]

where Object1, Object2 were `HasMetadataAndData` objects (which could act like mappables?).

PR #1180, opened 2024-11-15
Add parser to boundary conditions

### LINKED ISSUE #980: Add the parser calling framework to the other storage classes
We should be able to call/pass different parsers for the different data types as we do for ObsSurface etc.

SutarPrasad:

   Added parser framework for Boundary class
   Updated BoundaryCondition class framework
   Added test for Boundary conditions parser

With the recent comments, I have made openghg as a source_format and updated relevant parsr frameworks. Source_format is passed from the standardise_bc function now. 

- [x] Closes #980 (Replace xxxx with the Github issue number)


openghg/store/_boundary_conditions.py
```diff
@@ -160,66 +184,39 @@ def read_file(
...
-            key = "_".join((species, bc_input, domain))
+            additional_metadata["time_period"] = period_str
 
-            boundary_conditions_data: dict[str, dict] = {}
-            boundary_conditions_data[key] = {}
-            boundary_conditions_data[key]["data"] = bc_data
-            boundary_conditions_data[key]["metadata"] = metadata
-
-            matched_keys = set(metadata) & set(fn_input_parameters)
+            matched_keys = set(bc_data) & set(fn_input_parameters)
```
brendan-m-murphy, line 219: `bc_data` is a dataset, so it acts like a mapping from data variable names to data arrays, so `set(bc_data)` will be the data variable names from `bc_data`. In the old code, it looks like it should be the set of keys in the metadata though.

openghg/store/_boundary_conditions.py
```diff
@@ -232,7 +229,7 @@ def read_file(
 
             # Mop up and add additional keys to metadata which weren't passed to the parser
             boundary_conditions_data = self.update_metadata(
-                boundary_conditions_data, additional_input_parameters, additional_metadata
+                boundary_condition_data, additional_input_parameters, additional_metadata
```
brendan-m-murphy, line 232: Maybe `boundary_condition_data` should be `boundary_conditions_data` here and above. Or the result of `self.update_metadata` should be `boundary_condition_data`

openghg/standardise/boundary_conditions/_boundary_conditions.py
```diff
@@ -0,0 +1,75 @@
...
+import xarray as xr
+
+from openghg.util import clean_string, timestamp_now, synonyms
+from openghg.store import update_zero_dim
+
+logger = logging.getLogger("openghg.standardise.boundary_conditions")
+logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
+
+
+def parse_boundary_conditions(
```
rt17603, line 13: The way we've done this with other data types is to call the `parse_*` function which is basically just our internal format `parse_openghg`. So to be consistent with the other data types, this should be `parse_openghg` and the source_format should be "openghg".

openghg/store/_boundary_conditions.py
```diff
@@ -258,11 +255,12 @@ def read_file(
             #     metastore=metastore,
             #     update_keys=update_keys,
             # )
-
+            results["processed"][filepath.name] = datasource_uuids
```
brendan-m-murphy, line 258: Is there another key besides "processed"? (Either for boundary conditions, or for other data classes?)

openghg/store/_boundary_conditions.py
```diff
@@ -56,6 +60,7 @@ def read_file(
         species: str,
         bc_input: str,
         domain: str,
+        source_format: str = "boundary_conditions",
```
rt17603, line 63: As per comment within `_boundary_conditions.py` this should be "openghg"
```suggestion
        source_format: str = "openghg",
```

openghg/store/_boundary_conditions.py
```diff
@@ -66,7 +71,7 @@ def read_file(
         filters: Optional[Any] = None,
         chunks: Optional[Dict] = None,
         optional_metadata: Optional[Dict] = None,
-    ) -> dict:
+    ) -> Dict:
```
rt17603, line 74: Question for @brendan-m-murphy on this - were we aiming to move to `dict` rather than using `Dict`? In which case we should keep this as `dict`?
```suggestion
    ) -> dict:
```

openghg/store/_boundary_conditions.py
```diff
@@ -76,6 +81,7 @@ def read_file(
               - a model name such as "MOZART" or "CAMS"
               - a description such as "UniformAGAGE" (uniform values based on AGAGE average)
             domain: Region for boundary conditions
+            source_format: "boundary_conditions is default"
```
rt17603, line 84: As per other comments, this should be "openghg". Would be good to include some more description here of what this is (even if copied / derived from other standardise functions).

brendan-m-murphy (changes requested):
I think maybe a problem with the definition of `matched_keys` in `BoundaryConditions.read_file`. Other than that, I just have some minor suggestions.

openghg/store/_boundary_conditions.py
```diff
@@ -105,31 +111,49 @@ def read_file(
...
 
+        data_type = self._data_type
+        standardise_parsers = define_standardise_parsers()[data_type]
+
+        try:
+            source_format = standardise_parsers[source_format.upper()].value
+        except KeyError:
+            raise ValueError(f"Unknown data type {source_format} selected.")
+
+            # Get current parameter values and filter to only include function inputs
```
rt17603, line 141:
```suggestion
        # Get current parameter values and filter to only include function inputs
```

openghg/store/_boundary_conditions.py
```diff
@@ -160,66 +184,39 @@ def read_file(
...
-            metadata["end_date"] = str(end_date)
-            metadata["data_type"] = data_type
-
-            metadata["max_longitude"] = round(float(bc_data["lon"].max()), 5)
-            metadata["min_longitude"] = round(float(bc_data["lon"].min()), 5)
-            metadata["max_latitude"] = round(float(bc_data["lat"].max()), 5)
-            metadata["min_latitude"] = round(float(bc_data["lat"].min()), 5)
-            metadata["min_height"] = round(float(bc_data["height"].min()), 5)
-            metadata["max_height"] = round(float(bc_data["height"].max()), 5)
+            additional_metadata["start_date"] = str(start_date)
```
rt17603, line 204: I think we would probably want these lines [204-217] which are adding the additional metadata details within the parse function. This may end up being something we put into a new utility function at some point so multiple parse functions can use it but I think in general we were trying to remove these kinds of direct metadata updates from with the data type classes.

openghg/types/_enum.py
```diff
@@ -73,3 +73,7 @@ class DataTypes(Enum):
     NPL = "NPL"
     OPENGHG = "OPENGHG"
     INTEM = "INTEM"
+
+
+class BoundaryConditions(Enum):
+    BOUNDARY_CONDITIONS = "Boundary_Conditions"
```
rt17603, line 79: As per other suggestions for source_format:
```suggestion
    OPENGHG = "OPENGHG"
```

openghg/standardise/boundary_conditions/_boundary_conditions.py
rt17603, line 1: We should make sure the `source_format` option is passed through to the `standardise_bc` function (with the default value):
https://github.com/openghg/openghg/blob/add-parser-to-boundary-conditions/openghg/standardise/_standardise.py#L386

rt17603 (changes requested):
Thanks for putting this in and this is mostly there I think but I'll take another look once the requested changes are in.

Main requested updates:
 - source_format should be "openghg" to keep consistent with other data types
 - source_format should be an input for the standardise_bc functions as well
 - some more details from the data type class around metadata could be included in the parse function.

openghg/store/_boundary_conditions.py
brendan-m-murphy: `dict` is now preferred, but we'll probably use a linter to update the type hints now that we have 3.10 (e.g. before the next release, just update everything)

openghg/store/_boundary_conditions.py
rt17603: I actually don't think we need this or the separate `additional_input_parameters` dictionary defined below anymore if we've moved to using the parse framework. The new L151 in the openghg/store/_boundary_conditions.py file creates the `additional_input_parameters` dictionary against the parse function itself.

I think the previous set up was a temporary stopgap before we moved to the parse framework for this since we didn't have the parse function call to compare against.

rt17603 (changes requested):
Some small additional changes to bring this in line with the other data type classes (some around the metadata itself and some minor). Have included all as suggested changes so hopefully should be simple to check and update.

SutarPrasad:
results is used to store the datasource uuids. Complying to format with other classes

rt17603:
> results is used to store the datasource uuids. Complying to format with other classes

We actually still have a slightly different format between ObsSurface and the other data types (e.g. Flux, Footprint) - this previously matched to the other data types. As this will be updated anyway when #1075 is implemented this, this is less important.

rt17603 (approved):

openghg/standardise/boundary_conditions/_openghg.py
```diff
@@ -0,0 +1,96 @@
...
+        metadata["min_longitude"] = round(float(bc_data["lon"].min()), 5)
+        metadata["max_latitude"] = round(float(bc_data["lat"].max()), 5)
+        metadata["min_latitude"] = round(float(bc_data["lat"].min()), 5)
+        metadata["min_height"] = round(float(bc_data["height"].min()), 5)
+        metadata["max_height"] = round(float(bc_data["height"].max()), 5)
+        metadata["input_filename"] = filepath.name
+
+        metadata["time_period"] = period_str
+
+        key = "_".join((species, bc_input, domain))
```
brendan-m-murphy, line 90: This line is the only reason species, bc_input, and domain are needed by this parser. Maybe they should be removed from `metadata` so that when the outputs of `standardise_` are reworked, they will show up as unused arguments.

openghg/store/_boundary_conditions.py
```diff
@@ -160,109 +187,57 @@ def read_file(
...
-            # existing Datasources
-            datasource_uuids = self.assign_data(
-                data=boundary_conditions_data,
-                if_exists=if_exists,
-                new_version=new_version,
-                data_type=data_type,
-                compressor=compressor,
-                filters=filters,
-            )
+        # TODO: MAY NEED TO ADD BACK IN OR CAN DELETE
```
brendan-m-murphy, line 222: I know this was just flagged because it was reformatted, but can this commented out code be removed?

brendan-m-murphy (approved):
Just two small suggestions, but they're not essential.

openghg/standardise/boundary_conditions/_openghg.py
rt17603: It's a good idea but doesn't quite work at the moment because the `additional_input_parameters` are filtered based on what the parse function accepts. Because it accepts `domain`, `bc_input`, it then doesn't appear in the `additional_input_parameters` and so isn't added afterwards. Once `domain`, `bc_input` etc. are removed from the `parse_openghg` signature this would work though.

openghg/store/_boundary_conditions.py
rt17603: Yes, this commented out code is still hanging around in a few places - this functionality was to make sure certain keys were updated in the internal metadata for each Datasource (start_date, end_date and latest_version) but this moved somewhere else when it got generalised to updating _all_ internal metadata from the metastore. I don't think we'd go back to this old way of doing it so we can start removing this when we see it.

openghg/standardise/boundary_conditions/_openghg.py
brendan-m-murphy: Ah right, okay

PR #1198, opened 2025-01-07
Metakeys config not found but config folder exists error handling bug

### LINKED ISSUE #1157: If metakeys config not found but config folder exists, standardising always raises an error
### What happened?

The tests for inversions were trying to standardise some data and `get_metakeys` found that `<bucket>/config/metakeys.json` did not exist, so it called `create_default_config`. However, the folder `<bucket>/config` did exist, so this causes `create_default_config` to raise an error.

`get_metakeys` should probably check if the config folder exists, or create it with `exists_ok=True`.

_No response_

### Relevant log output

_No response_

_No response_

### Environment

<details>

</details>

SutarPrasad:
 Added exist_ok-True where it creates folderpath. Removed error inside create_default_config

- [ ] Closes #1157 (Replace xxxx with the Github issue number)

### DIFF
--- a/openghg/objectstore/_config.py
+++ b/openghg/objectstore/_config.py
@@ -5,7 +5,7 @@ from pprint import pformat
 from pathlib import Path
 import pkgutil
 
-from openghg.types import ConfigFileError, ObjectStoreError
+from openghg.types import ConfigFileError
 from openghg.util import timestamp_now
 
 logger = logging.getLogger("openghg.objectstore")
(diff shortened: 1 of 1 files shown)


openghg/objectstore/_config.py
```diff
@@ -127,10 +127,8 @@ def create_default_config(bucket: str) -> None:
...
     """
     config_folderpath = _get_config_folderpath(bucket=bucket)
-    if config_folderpath.exists():
-        raise ObjectStoreError(f"config folder already exists at {config_folderpath}")
 
-    config_folderpath.mkdir(parents=True)
+    config_folderpath.mkdir(parents=True, exist_ok=True)
 
     # Make the expected folder structure
     db_config_folderpath = config_folderpath.joinpath("config")
```
brendan-m-murphy, line 134: the `mkdir` below would also cause an error without `exist_ok=True` for the example from the issue (I think).

This function is kind of confusing because it looks like `_write_metakey_config` depends implicitly on the preceding code (for instance, where is `db_config_folderpath` used?). Not sure if it's worth doing something about that though.

brendan-m-murphy (commented):
It looks like this could still raise the same kind of error later on. It would be useful to have a test to know it works.

PR #1197, opened 2025-01-03
Fx1194 sync metadata to retrieve icos

SutarPrasad:
* **Summary of changes** (Bug fix, feature, docs update, ...)
 Shifted align_metadata_attributes function declaration from obsurface to metadata.py and called it in retrieve_remote so that update mismatch works for icos remote retrieval.

* **Please check if the PR fulfills these requirements**

- [x] Closes #1194 (Replace xxxx with the Github issue number)
- [x] [Tests added and passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature
- [x] All code checks passing - `black --line-length 110` run over code, `mypy` and `flake8` not showing any errors
- [x] Documentation and tutorials updated/added
- [x] Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature
- [x] Added any new requirements to `requirements.txt` and `recipes/meta.yaml`

PR #1196, opened 2024-12-20
Iss1190/icos flags

joe-pitt:
Resolves Issue #1190 . Now the retrieve function only accepts data with a "valid" flag ("O","U" or "R").

In order to get this working I merged in the branch from PR #1195 . So I'll make this a draft PR for now until #1195 gets merged into devel.

Not sure if extra tests are needed here?

* **Please check if the PR fulfills these requirements**

- [x] Closes #1190 
- [ ] [Tests added and passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature
- [ ] All code checks passing - `black --line-length 110` run over code, `mypy` and `flake8` not showing any errors
- [ ] Documentation and tutorials updated/added
- [ ] Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature
- [ ] Added any new requirements to `requirements.txt` and `recipes/meta.yaml`

Issue #1190, opened 2024-12-07
Flag out icos data


jpitt-test:
### What is your issue?

ICOS data that has been flagged invalid is usually associated with a mole fraction reported as -999.99. At the moment I think the retrieve function uses this -999.99 value to exclude invalid data. However, we've recently noticed a few datasets (e.g. https://meta.icos-cp.eu/objects/dvsu8Y6AX9-YmpPwzp-hXVKC) from icos that have data flagged as bad without the -999.99. So we should probably add another criteria that excludes data whenever the data flag is "N" or "K".

PR #1199, opened 2025-01-07
Make parser output list of `MetadataAndData` objects

### LINKED ISSUE #1075: Replace nested metadata/data dictionaries with list of MetadataAndData objects

brendan-m-murphy:

The output of parser functions has been changed to a list of `MetadataAndData` objects.
Previously, the output was a nested dictionary.

This required changing what is returned from the `standardise_` functions, since the
old return values depended on the keys in the nested dictionary returned by the parsers.
The new return value is a list of dictionaries that contain the uuid of the standardised data,
a flag to indicate if the datasource is new, and any required metadata used for datasource lookup.

Note: to fix the tests, I added a helper function to create keys based on the metadata returned by `standardise_`.
The assert statements where this is used could be changed to dictionary comparisons instead.

- [x] Closes #1075


rt17603:
Note that we're waiting for PR #1197 and a PR based on #1196 (Issue #1190 ) to go in before merging this PR.

SutarPrasad:
Once PR #1200 gets merged the blocker will be clear for this PR.

tests/store/test_boundary_conditions.py
```diff
@@ -37,8 +56,11 @@ def test_read_data_monthly(mocker):
         source_format='openghg'
     )
 
-    # assert proc_results == {"ch4_mozart_europe": {"uuid": "test-uuid-1", "new": True}}
-    assert proc_results["ch4_mozart_europe"]["new"] is True
+    assert proc_results is not None
+    print(proc_results)
+    print(fake_uuid.hex_num, fake_uuid.uuid_num)
+    assert {"uuid": "test-uuid-1", "new": True}.items() <= select(proc_results, "new", "uuid")[0].items()
+    assert make_keys(proc_results, "species", "bc_input", "domain")[0].lower() == "ch4_mozart_europe"
```
rt17603, line 63: Rather than having a `make_keys` function which recreates the original key, I'm wondering if we'd be better off moving to checking those keys instead (essentially that's what we're doing anyway):
```
result = proc_results[0]
assert result["species"] == "ch4"
assert result["bc_input"] == "mozart"
assert result["domain"] = "europe"
```
or using `proc_results[0].get("species")` instead but is perhaps nicer to split this up. This type of check also aligns with the types of checks in test_standardise as well.

This would also apply to all other times the `make_keys` function is used.

tests/standardise/test_standardise.py
```diff
@@ -52,11 +53,12 @@ def test_standardise_obs_two_writable_stores():
         store="user",
     )
 
-    results = results["processed"]["hfd.picarro.1minute.100m.min.dat"]
+    results = filt(results, file="hfd.picarro.1minute.100m.min.dat")
```
rt17603, line 56: For this PR I think this looks like a reasonable solution but I think the need for this `filt()` function says to me that our `results` output should perhaps be more file-focussed (at least for our current set up). This could be another dict form with keys or, as you've suggested previously, files relates to `SearchResults` objects.

e.g. something like:
```
{FILENAME1: [{"uuid": ..., ...}, {"uuid":, ..., ...}], FILENAME2: [{"uuid": ..., ...}, {"uuid":, ..., ...}]}
```
or
```
{FILENAME1: SearchResults, FILENAME2: SearchResults}
```
At least for the cases where we will have accepted or will accept multiple files. It's another complex object but at least it allows the details associated with each file to be easily linked.

We can perhaps open another issue to look into this a bit more.

openghg/store/base/_base.py
```diff
@@ -387,18 +387,29 @@ def assign_data(
...
                     self._metastore.insert(datasource_metadata)
+                    logger.info("Created new datasource with UUID %s", datasource.uuid())
                 else:
                     self._metastore.update(where={"uuid": datasource.uuid()}, to_update=datasource_metadata)
+                    logger.info("Added data to existing datasource with UUID %s", datasource.uuid())
+
+                required_info = {
+                    k: v
+                    for k, v in metadata.items()
+                    if k in required_keys and v is not None and v != "not_set"
```
rt17603, line 398: Rather than referencing the term "not_set" directly we can use the definition for these values from the [not_set_metadata_values() function](https://github.com/openghg/openghg/blob/devel/openghg/util/_metadata_util.py#L37)
e.g. something like:
```
if k in required_keys and v is not None and v not in not_set_metadata_values()
```

tests/store/test_boundary_conditions.py
```diff
@@ -9,8 +10,26 @@
 
 
 def test_read_data_monthly(mocker):
-    fake_uuids = ["test-uuid-1", "test-uuid-2", "test-uuid-3"]
-    # mocker.patch("uuid.uuid4", side_effect=fake_uuids)
+    class FakeUUID:
```
rt17603, line 13: Would this be useful as a more general class rather than defined within a function?

rt17603 (commented):
Looks like this a right pain to do so well done in getting something put together! This is a great stepping stone - I made a comment about another issue but I don't think this needs to be addressed here.

My main comment would be in maybe retiring the idea of the short keys all together since we've moved away from dict --> list now (e.g. don't use the `make_keys` function - just check the key values). Other than that looks good.

tests/standardise/test_standardise.py
brendan-m-murphy: The dictionaries returned by `ObsSurface.read_file` seemed more complicated than the other classes (it had another layer of "processed" vs. some alternative that never seemed to be use), and might have been the only case where the filename was a key in some part of the nested dictionary. I'm not sure the other classes were "file focused"; they seemed to just use the "keys" like "gas-site-inlet" as dictionary keys.

I'd prefer to leave the results in the current somewhat generic/flatter form. As it is, you could convert the list of dicts returned to a DataFrame and query it like `SearchResults.results`; I just made the `filt` helper function because it seemed simpler. (I guess the DataFrame approach could make `filename` an index, and then you'd get a view like the one you're suggesting.)

We can look at it again later; it might be worth waiting until we combine the remote, read file, transform, etc pathways, or at least come up with a plan for that.

tests/store/test_boundary_conditions.py
brendan-m-murphy: Okay yea, that seems like the best way to do these tests. I'll implement that.

tests/standardise/test_standardise.py
brendan-m-murphy: Also, the `filt` function is used elsewhere, like in `tests/store/test_flux.py` on line 50, and there it is used without the filename. This might not be the best way to do this test though.

tests/store/test_boundary_conditions.py
brendan-m-murphy: Possibly, although ideally we wouldn't be testing UUID related functionality at this level. I think this was necessary because some other function kept using up the mocked UUIDs, or was trying to call a method on them. Anyway, I did this to maintain the content of the test, but definitely would prefer to avoid elaborate mocking if possible.

rt17603: Ok. Would you mind adding a few comments to that effect around the test and class - just useful to understand the context of why something is included in case we decide to update it later on.

openghg/store/base/_base.py
brendan-m-murphy: I'll make this change.

tests/store/test_boundary_conditions.py
brendan-m-murphy: I added a bit of explanation to the `FakeUUID` class.

rt17603 (approved):
All looks good to me - would it be worth asking Joe just to check the ICOS retrieve before merging though?

brendan-m-murphy:
> All looks good to me - would it be worth asking Joe just to check the ICOS retrieve functionality before merging though?

Joe ran a test and it seems like NRT is overwriting level 2 data, so something is being ignored by datasource lookup. I'll add a test based on his example and then try to sort it out.

PR #1211, opened 2025-01-15
Fix to store ICOS data by data level

brendan-m-murphy:
* **Summary of changes** (Bug fix, feature, docs update, ...)

The method `ObsSurface.store_data` was still using the old keyword
`icos_data_level` instead of the new keyword `data_level` to distinguish
data sources.

This was causing ICOS data with different levels to be written to the same
datasource.

A test was added to check this; it failed before the change to `store_data`,
and passes afterwards.

* **Please check if the PR fulfills these requirements**

- [x] Closes #1210 
- [x] [Tests added and passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature
- [x] All code checks passing - `black --line-length 110` run over code, `mypy` and `flake8` not showing any errors
- [x] Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature

PR #1212, opened 2025-01-15
Iss1106/read icos nc

### LINKED ISSUE #1106: Read LTR from ICOS files

Our default setup for RHIME inversions currently uses mf_variability and mf_repeatability in calculating the measurement-model mismatch error for AGAGE/DECC sites. At the moment for ICOS sites only the variability is read into the object store by openghg. The rough equivalent to mf_repeatability in the ICOS files is the variable LTR. Should we be reading this in too and using it in our inversions?

We could read in LTR (and possibly other variables such as CMR and STTB) from the ICOS files

_No response_

### Additional context

LTR, CMR are explained in [Yver Kwok et al., 2015], doi:10.5194/amt-8-3867-2015

joe-pitt:
Addresses Issue #1106 - adds ability to retrieve combined Obspack .nc data from the carbon portal

- [x] Closes #1106


joe-pitt:
I've had a go at this - I'm sure someone more competent could do this in a neater way, but I wonder whether a more general tidy of this bit of code might be a task for the future... In any case feel free to dive in and edit stuff, or suggest changes.

Does this need a test adding? I've never written one so I might need a bit of help.

joe-pitt:
I'm also not wedded to the default icos variable names (`icos_ltr`, `icos_smr`, `icos_sttb`), although it might be good to keep the end bits so that they correspond to the terms used in Camille's paper. @rt17603 - any thoughts on this?

brendan-m-murphy:
An option for tidying this would be to make separate functions for each dataset source, and then call the right function in `_retrieve_remote`. This might result in a few functions that are a bit easier to understand, and one if-else chain at the beginning to select the right option.

For a test, you might do something like:
```
@pytest.mark.icos
def test_icos_obspack():
    # some code to retrieve obspack data

    # some checks that the data has the expected properties
    assert "icos_ltr" in result.data  # e.g. check we have a dataset with the icos_ltr variable
```

joe-pitt:
Thanks for looking at this @brendan-m-murphy. I like the idea of different functions - do we want to do this now or as a follow up? I'd like to retrieve this data later this week, but I can keep this on a branch for now. I'll try to add the test too.

In the meantime I've changed my mind about the naming - let's stick with `spec repeatability` for now so that the PARIS files remain consistent

rt17603:
> I'm also not wedded to the default icos variable names (`icos_ltr`, `icos_smr`, `icos_sttb`), although it might be good to keep the end bits so that they correspond to the terms used in Camille's paper. @rt17603 - any thoughts on this?

I think direct traceability is good for this - perhaps we stick with them for now if they don't correspond to other attributes we may want to include for other data?

joe-pitt:
> > I'm also not wedded to the default icos variable names (`icos_ltr`, `icos_smr`, `icos_sttb`), although it might be good to keep the end bits so that they correspond to the terms used in Camille's paper. @rt17603 - any thoughts on this?
> 
> I think direct traceability is good for this - perhaps we stick with them for now if they don't correspond to other attributes we may want to include for other data?

Would another option be to call this e.g. `ch4_repeatability` then have an attribute that says it is the ICOS LTR? Maybe with a link to the relevant paper? Is this easily achievable within the retrieve function? It looks like the other variable attributes are assigned elsewhere

brendan-m-murphy:
> > > I'm also not wedded to the default icos variable names (`icos_ltr`, `icos_smr`, `icos_sttb`), although it might be good to keep the end bits so that they correspond to the terms used in Camille's paper. @rt17603 - any thoughts on this?
> > 
> > 
> > I think direct traceability is good for this - perhaps we stick with them for now if they don't correspond to other attributes we may want to include for other data?
> 
> Would another option be to call this e.g. `ch4_repeatability` then have an attribute that says it is the ICOS LTR? Maybe with a link to the relevant paper? Is this easily achievable within the retrieve function? It looks like the other variable attributes are assigned elsewhere

Either a standard variable name, or an attribute could work. To decide how to do the resampling, we would then need to hard code a mapping of variable name/attribute to resampling method.

Another option that would work for resampling would be to use an attribute to indicate how the data should be resampled, although this would introduce internal OpenGHG stuff into the data attributes. (I suppose that info could be added to metadata for an obs. datasource, or could be in some kind of config file.)

(For context, I'm referring to the part of `get_obs_surface` that does things like "if "variability" in data_var: then ...".

rt17603:
Yes I agree that we should aim for consistent internal variable names where we're including details for a given property - especially around error types. I agree that a variable name and an associated attribute could work well e.g. perhaps something like:
 - Data variable: ch4_repeatability
 - Attribute: ch4_repeatability_comment: "Repeatability derived from ICOS LTR"

joe-pitt:
> Yes I agree that we should aim for consistent internal variable names where we're including details for a given property - especially around error types. I agree that a variable name and an associated attribute could work well e.g. perhaps something like:
> 
> * Data variable: ch4_repeatability
> * Attribute: ch4_repeatability_comment: "Repeatability derived from ICOS LTR"

Sounds good to me! Should we assign this attribute within retrieve_remote? I guess that makes the most sense because then we can include it only for the icos data that comes with LTR, but I don't think any other variable attributes are assigned in retrieve_remote at present so I wondered if there was a reason for this?

rt17603:
Just to check this PR is still current? Have the changes from PR #1211 been merged in and this all works ok?

joe-pitt:
> Just to check this PR is still current? Have the changes from PR #1211 been merged in and this all works ok?

Sorry yeah it's still current but I'm struggling to find time to tidy it up (add tests etc). I've been using it and it seems to be working. I'm pretty sure I merged devel in here after that PR went it

openghg/retrieve/icos/_retrieve.py
```diff
@@ -284,48 +294,70 @@ def _retrieve_remote(
...
-
-        attributes["species"] = the_species
+        spec = measurement_type.split()[0]
+        attributes["species"] = spec
         acq_data = specific_info["acquisition"]
         station_data = acq_data["station"]
 
+        # Hack to convert units for Obspack nc files
+        if dataset_source == "ICOS Combined":
+            if spec == "co2":
```
rt17603, line 311: I'm wondering if we could include something which uses the internal definitions for unit conversions:
```
    openghg_attributes = load_internal_json("attributes.json")
    unit_interpret = openghg_attributes["unit_interpret"]

[...]
    if dataset_source == "ICOS Combined":

            if spec == "co2":
                units = "ppm"
            else:
                units = "ppb"

            unit_value = float(unit_interpret.get(units, "1"))
            dataframe["value"] = dataframe["value"] * unit_value 
            dataframe["value_std_dev"] = dataframe["value_std_dev"] * unit_value 
            dataframe["icos_LTR"] = dataframe["icos_LTR"] * unit_value 
            dataframe["icos_SMR"] = dataframe["icos_SMR"] * unit_value 
            dataframe["icos_STTB"] = dataframe["icos_STTB"] * unit_value

[...]
```
This uses the details within the ["attributes.json" file (e.g. "ppb": "1e-9". "ppm": "1e-6")](https://github.com/openghg/openghg/blob/devel/openghg/data/attributes.json#L2) to define the conversion factors. This is also currently used in [the plotting steps](https://github.com/openghg/openghg/blob/devel/openghg/plotting/_timeseries.py#L203) to apply unit conversion so is perhaps a good way to start to centralise this.

rt17603 (commented):
I've made one small suggestion but I'm also happy for this to be merged if devel has been merged in already and this is all working ok.

openghg/retrieve/icos/_retrieve.py
joe-pitt: This seems good! I think it needs to be `/ unit value` in this case?

joe-pitt:
@rt17603 @brendan-m-murphy - I think this is ready to merge now. Sorry it's taken me so long!

rt17603 (approved):

PR #1218, opened 2025-02-04
Standardise footprint and pipeline to incorporate appropriate keys

### LINKED ISSUE #1216: Update standardise_footprint (or new specific method) and pipeline to incorporate appropriate keys
Because we need to link (satellite) column data to associated footprint data when we build our inputs for the inversion methods (via ModelScenario), we need to ensure that the appropriate keys from the satellite data can also be associated with the right footprints.

Issue #1215 proposes to add `region` and update the `selection` keywords associated with the satellite column data. For this we also need to allow footprints to be added which can be associated with these keywords within their metadata.

The suggestion is to update `standardise_footprint` (or new specific method) and pipeline to incorporate the keys:
  - `region`
  - `selection`

as a minimum.

SutarPrasad:
Similar to obs column added `fp_region` and `selection` args to:
standardise_column
Footprints.read_file()
get_footprint()
Further checks for domain and region

- [x] Closes #1216 (Replace xxxx with the Github issue number)


openghg/store/_footprints.py
```diff
@@ -283,7 +288,23 @@ def read_file(
...
-        site = clean_string(site)
+        if satellite is None and site is None:
+            raise ValueError("Please pass either site or satellite value")
+
+        if satellite is not None:
+            satellite = clean_string(satellite)
+        else:
+            site = clean_string(site)
+
+        if domain is not None and fp_region is not None:
```
rt17603, line 299: For footprint files we wouldn't want to apply the same logic for the keywords that we needed to for the obscolumn data. For footprint `domain` is an essential keyword as this describes the area for the 2D map. For the obscolumn data we were essentially allowing `domain` to be another way to specify `obs_region` which is why we applied this logic.

Instead though we may want to include checks to ensure either `site` or `satellite` AND `obs_region` are specified so:
 - If `site` is specified, use this keyword
 - If `site` is not specified, check if `satellite` and `obs_region` are specified, otherwise raise an Error

openghg/standardise/footprints/_acrg_org.py
```diff
@@ -228,7 +235,16 @@ def parse_acrg_org(
     # This might seem longwinded now but will help when we want to read
     # more than one footprints at a time
     # TODO - remove this once assign_attributes has been refactored
-    key = "_".join((site, domain, model, inlet))
+    key_parts = [
```
rt17603, line 238: In the short term, for this we'd likely want either:
 - (site, domain, model, inlet)
 - (satellite, obs_region, domain, model, inlet)

rt17603 (changes requested):
I've included some details on how we will likely want to interpret the new keywords which means there's a few places to update. We'll likely also want to apply the same changes to `_paris.py` as `_acrg_obs.py` so we're forward planning here - we could also consider if there's code (such as the metadata definition) which could move to the `Footprint` data type class but perhaps this is not in scope for this PR.

openghg/retrieve/_access.py
```diff
@@ -544,8 +544,10 @@ def get_bc(
 
 
 def get_footprint(
-    site: str,
     domain: str,
+    obs_region: str | None = None,
```
rt17603, line 548: Could we move `obs_region` to be below `site` and `satellite` (as we have for the `standardise_footprint` function)? I think keeping this consistent would emphasise the relative importance of the keywords as we're still going be including `site` footprints most often.

tests/standardise/test_standardise.py
```diff
@@ -636,3 +636,67 @@ def test_standardise_sorting_false(caplog):
...
+    """
+    Tests standardise footprint for satellite data and associated metadata keys."""
+    clear_test_stores()
+
+    datapath = get_footprint_datapath("GOSAT-BRAZIL-column_TEST_SA_201004_compressed.nc")
+
+    satellite = "GOSAT"
+    network = "GOSAT"
+    domain = "TEST_SA"
+    obs_region = "TEST_SA"
```
rt17603, line 677: I think we should keep the `domain` and `obs_region` values different for the tests to at least showcase these values don't have to be the same. The `obs_region` of "BRAZIL" was fine previously:
```suggestion
    obs_region = "BRAZIL"
```

tests/standardise/test_standardise.py
```diff
@@ -636,3 +636,67 @@ def test_standardise_sorting_false(caplog):
...
+            filepath=datapath,
+            satellite=satellite,
+            network=network,
+            model="CAMS",
+            height=6500,
+            period='1S',
+            domain=domain,
+            obs_region=obs_region,
+            selection="LAND",
+            force=True,
```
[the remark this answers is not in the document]
rt17603: It still looks like the `force` keyword is being used here? Can this be removed?

tests/standardise/test_standardise.py
```diff
@@ -636,3 +636,67 @@ def test_standardise_sorting_false(caplog):
...
+
+def test_standardise_footprints_satellite_raises_error():
+    """
+    Tests standardise footprint raises value error when site and obs_region values are not supplied.
+    """
+    datapath = get_footprint_datapath("GOSAT-BRAZIL-column_SOUTHAMERICA_201004_compressed.nc")
+
+    satellite = "GOSAT"
+    network = "GOSAT"
+    domain = "BRAZIL"
```
rt17603, line 649: Was this using the same data `test_standardise_footprint_satellite()`? If so the domain (and file name) should be updated to be consistent here as well
```suggestion
    domain = "TEST_SA"
```

tests/standardise/test_standardise.py
```diff
@@ -636,3 +636,67 @@ def test_standardise_sorting_false(caplog):
...
+def test_standardise_footprint_satellite():
+    """
+    Tests standardise footprint for satellite data and associated metadata keys."""
+    clear_test_stores()
+
+    datapath = get_footprint_datapath("GOSAT-BRAZIL-column_TEST_SA_201004_compressed.nc")
+
+    satellite = "GOSAT"
+    network = "GOSAT"
+    domain = "TEST_SA"
```
rt17603, line 676: Can I just check, was this a cutdown lat-lon domain or did this still match the original "SOUTHAMERICA" domain?

rt17603 (commented):
Nearly there - just a few suggestions based on previous comments.

tests/standardise/test_standardise.py
SutarPrasad: cut down

SutarPrasad: I just looked at domain_info.json and we dont have SouthAmerica domain listed. So it might be possible it is working on passing values as SouthAmerica as well as TEST_SA. In this case what would we want to do? Keep it SouthAmerica or TEST_SA?

openghg/standardise/footprints/_acrg_org.py
```diff
@@ -228,7 +237,25 @@ def parse_acrg_org(
...
+        key_parts = [
+            cast(str, site),
+            cast(str, domain),
+            model,
+            inlet,
+        ]
+
+    key_parts = [str(part) for part in key_parts if part is not None]
+
+    key = "_".join(key_parts)
```
brendan-m-murphy, line 258: The keys aren't actually used anymore, so any placeholder value could work.

tests/standardise/test_standardise.py
rt17603: > I have reduced the time points for this. It seems to work with domain as well as test_sa

So the domain itself hasn't been cut down? If that's the case this should still match to "SOUTHAMERICA" and so we can use this. You're right there's no "SOUTHAMERICA" domain in domain_info.json at the moment but there is a set of lat, lon values associated which were used for publications (in which case I think I may try it to openghg_defs as this is not currently exhaustive).

SutarPrasad: I have not specifically cut down on the domain. So I'll make it SOUTHAMERICA.

rt17603: Ok, I'll wait for this to go in before adding another review.

SutarPrasad: its done

tests/standardise/test_standardise.py
```diff
@@ -656,3 +656,66 @@ def test_standardise_surface_niwa(caplog):
...
+    network = "GOSAT"
+    domain = "SOUTHAMERICA"
+
+    with pytest.raises(ValueError):
+        standardise_footprint(
+            filepath=datapath,
+            satellite=satellite,
+            network=network,
+            model="CAMS",
+            height=6500,
```
rt17603, line 677: Even for the fail test, for consistency this should use `inlet` key word and be set to "column"
```suggestion
            inlet="column",
```

SutarPrasad: oh nice catch

rt17603 (approved):
This looks all good now - happy for this to be merged.

PR #1221, opened 2025-02-21
Fix filelock permissions

### LINKED ISSUE #1219: Error message for file lock permissions doesn't work correctly.
### What is your issue?

In `_classic_metastore.py` the following code is suppose to give a useful error message if you don't have the right permissions:

```
      if "w" not in perms[0]:
             raise PermissionError(
                 "You do not have the correct permissions to add data to this object store."
                 f"Ask {lock_path.owner()} to set group permissions for {lock_path} to 'rw'."
             )
```

But `perms[0]` gives the permissions of the file owner, which might not be the person who called this code. For this to work correctly, this part needs to figure out if the current user is either "owner", in the "group", or "other", then check for "w" in the right part of the permissions.

brendan-m-murphy:

Changed to make sure `FileLock` object is created with rw-rw-r
permissions.

Changed file lock naming scheme so we'll start fresh with this PR
and won't need to fix the broken permissions on all the old files.

- [x] Closes #1219

### DIFF
--- a/openghg/objectstore/_local_store.py
+++ b/openghg/objectstore/_local_store.py
@@ -303,7 +303,7 @@ def get_object_lock_path(bucket: str, key: str) -> Path:
     Returns:
         Path to object lock file
     """
-    lock_path = Path(f"{bucket}/{key}._data.lock")
+    lock_path = Path(f"{bucket}/{key}._data.lock_v2")
     if not lock_path.parent.exists():
         lock_path.parent.mkdir(parents=True)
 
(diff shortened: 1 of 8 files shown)


openghg/objectstore/_local_store.py
```diff
@@ -303,7 +303,7 @@ def get_object_lock_path(bucket: str, key: str) -> Path:
     Returns:
         Path to object lock file
     """
-    lock_path = Path(f"{bucket}/{key}._data.lock")
+    lock_path = Path(f"{bucket}/{key}._data.lock_v2")
```
rt17603, line 306: Just to check did you have to do this to not interfere with any current lock files?

rt17603 (commented):
One small clarification question but apart from that looks good to me.

openghg/objectstore/_local_store.py
brendan-m-murphy: This is so that we don't have to fix all the permissions on the existing lock files.

rt17603 (approved):

PR #1235, opened 2025-03-07
Updates to make openghg compatible with numpy 2.0

### LINKED ISSUE #1042: Make OpenGHG compatible with numpy 2.0
### What is your issue?

Numpy had a major release in June 2024, which includes backwards-incompatible changes.

Their migration guide is [here](https://numpy.org/devdocs/numpy_2_0_migration_guide.html). There is a plugin for the `ruff` linter which should make our code compatible with numpy 1.26 and 2.0.

Pytensor (dependency of PyMC) is not yet compatible with numpy 2.0, so we need to be compatible with numpy 1.26 until they upgrade.

brendan-m-murphy:

- Unpinned numpy and netcdf4 (which was previously pinned because
  they hadn't yet updated to numpy 2.0)
- Fixed test fixture where values selected from an `xr.DataArray` were passed to `np.linspace`.
- Replaced deprecated `numpy.NaN` with `numpy.nan`
- Fixed two tests that assume that the folder containing the `tests`
  directory is always called "openghg"

- [x] Closes #1042


rt17603 (approved):
Looks good to me.

rt17603 (approved):
Looks good - a few very small suggestions.

One comment is that the additional changes to `calc_dim_resolution` will change the functionality a little (this would not produce an error and would return a np.nan but will now raise an error for non-numerical values and if unable to find the resolution and dim != 1). However, that should be fine for most circumstances.

rt17603 (approved):

PR #1251, opened 2025-03-17
Refactor obspack

### LINKED ISSUE #1231: Restructure obspack creation considering how to move to a datapack
The initial version of the obspack creation is a prototype and from creating this we've learned about what kind of structures can work for this. Thinking about moving to a datapack (i.e. more general than just an obspack) we can consider:

 - `StoredData` object - this already exists but has limited functionality
 - `ObsPack` object - to contain `StoredData` and details about the `ObsPack` itself (location, version etc.)

Could also consider:
 - `ObsPack` as a subclass of `DataPack` which adds specific functionality? What other `Packs` would we want to make? Could we just have `DataPack` and no `ObsPack` and just build optionality into this?

See also #1078 for original discussion points on moving towards using an `ObsPack` object structure.

rt17603:

Refactoring _obspack.py to be more object-oriented. This should make it easier to move to various datapacks we may want as we move forward.

- [x] Closes #1231  (Replace xxxx with the Github issue number)


openghg/datapack/_obspack.py
```diff
@@ -143,6 +156,569 @@ def update_obspack_filename(
...
+        Currently this will write:
+            - retrieved_data
+            - release files
+            - site index file
+
+        """
+        retrieved_data = self.check_retrieved_data()
+
+        for data in retrieved_data:
+            data.write()
```
brendan-m-murphy, line 703: At this point, you could pass `{obspack_folder} / {obspack_name}` to the write method, so the stored data objects wouldn't need to know this info.

openghg/datapack/_obspack.py
```diff
@@ -143,6 +156,569 @@ def update_obspack_filename(
 
         return obspack_filename
 
+    def add_subfolder(self, subfolder: MultiSubFolder | None = None) -> None:
```
brendan-m-murphy, line 159: I still think it would make sense to have a `Folder` (or `SubFolder`) object that:
1. has a folder name/path
2. contains `StoredData` objects that will be stored in that folder
3. maybe has the code to check for duplicate filenames (since this only matters within a folder?)
4. can write out the `StoredData` objects it contains, combining the obs pack path with the folder path, and then passing this to `StoredData.write`.

openghg/datapack/_obspack.py
```diff
@@ -143,6 +156,569 @@ def update_obspack_filename(
...
+
+        self.obspack_name = obspack_name
+        self.version = version
+
+        return obspack_name, version
+
+    def define_obspack_path(self) -> Path:
+        """
+        Define the full output path for the obspack folder.
+        Args:
```
brendan-m-murphy, line 348: Don't need args in docstring. You could maybe make this a property called `path`, so you'd use `ObsPack.path` to get the path. (If you make it a property, then the docstring should read as those it were an attribute.)

openghg/datapack/_obspack.py
```diff
@@ -143,6 +156,569 @@ def update_obspack_filename(
...
+        """
+        release_file_ref = importlib.resources.files("openghg") / "data/obspack/obspack_README.md"
+
+        with importlib.resources.as_file(release_file_ref) as f:
+            release_file_path = f
+
+        release_files = [release_file_path]
+        return release_files
+
+    def check_retrieved_data(self) -> list[StoredData]:
```
brendan-m-murphy, line 422: You could make this a property called `retrieved_data` (and store the attribute as `self._retrieved_data`).

openghg/datapack/_obspack.py
```diff
@@ -120,18 +122,29 @@ def make_obspack_filename(
 
     def update_obspack_filename(
```
brendan-m-murphy, line 123: Maybe this should just be called `update_filename`? The main point is to define the filename that will be used to store the data in the `StoredData` object.

openghg/datapack/_obspack.py
```diff
@@ -143,6 +156,569 @@ def update_obspack_filename(
...
+    def check_retrieved_data(self) -> list[StoredData]:
+        """ """
+        if self.retrieved_data is None:
+            msg = "ObsPack does not contain details to write to disc. Please populate ObsPack.retrieved_data"
+            logger.exception(msg)
+            raise ValueError(msg)
+
+        return self.retrieved_data
+
+    def define_obspack_filenames(
```
brendan-m-murphy, line 431: Maybe this should be called `define_stored_data_filenames`?

openghg/datapack/_obspack.py
```diff
@@ -485,15 +1081,57 @@ def find_data_version(metadata: dict) -> str | None:
...
+        input = Path("")
+    else:
+        input = Path(input)
+
+    return input
+
+
+# CALLED BY StoredData
+def define_full_obspack_filename(
+    obspack_filename: pathType,
```
brendan-m-murphy, line 1096: maybe replace "obspack_filename" (in function name and parameter name) with just "filename"? The obs pack is an object that has a folder associated to it, which is what I think of when I read "obspack_filename".

brendan-m-murphy (commented):
Lot's of nice improvements. A few comments:
1. many methods/functions have the word "obspack" in them, but now that `ObsPack` is an object, it's a bit confusing. In some places, maybe "obspack" could be replaced by "stored_data". Also, a method doesn't need to contain the name of the class in it. (Alternatively, if "obspack filename" is a concept, maybe it should have a class associated to it.)
2. `StoredData` objects shouldn't need to know about the ObsPack's path until it is time to write.
3. A `Folder` (or `SubFolder`) object seems like a natural addition. The responsibility for organising folders seems to be split between ObsPack and StoredData.

Once `StoredData` doesn't need to know about obs pack or folder paths, then the filename associated with `StoredData` will be based purely on the metadata it contains.

openghg/datapack/_obspack.py
```diff
@@ -143,6 +156,569 @@ def update_obspack_filename(
...
+            df : DataFrame containing the collated details for the site from the associated metadata.
+            output_filename : Filename to use for writing site details.
+        Returns:
+            None
+        """
+
+        # Site index should include key metadata for each file (enough to be distinguishable but not too much?)
+        # Want to create a DataFrame and pass a file object to this - then can add comments before and after the table as needed
+
+        if not hasattr(self, "site_details"):
```
brendan-m-murphy, line 649: You can use a property here:

```
@property
def site_details(self) pd.DataFrame:
    if self._site_details is None:
        <code to create site details and have it to self._site_details>
    return self._site_details
```

rt17603:
> Lot's of nice improvements. A few comments:
> 
>     1. many methods/functions have the word "obspack" in them, but now that `ObsPack` is an object, it's a bit confusing. In some places, maybe "obspack" could be replaced by "stored_data". Also, a method doesn't need to contain the name of the class in it. (Alternatively, if "obspack filename" is a concept, maybe it should have a class associated to it.)
> 
>     2. `StoredData` objects shouldn't need to know about the ObsPack's path until it is time to write.
> 
>     3. A `Folder` (or `SubFolder`) object seems like a natural addition. The responsibility for organising folders seems to be split between ObsPack and StoredData.
> 
> 
> Once `StoredData` doesn't need to know about obs pack or folder paths, then the filename associated with `StoredData` will be based purely on the metadata it contains.

Thanks for the feedback Brendan, that's really helpful. I haven't looked into the potential benefits of adding the property feature yet but I've updated some of the details around points (1) and (2) to tidy this up.

For point (3), reflecting on this I think it would be preferable not to impose additional (potentially more rigid) structure within the obspack itself as I think we should try and keep this flexible based on the needs of different outputs. In this specific case at the moment, we've set this up to create folders based on obs type names but for other obspacks  / datapacks we may not want this or want a different structure entirely. I think I've been seeing `StoredData` as in the world of the obspack and so I think it's ok for it to have some knowledge of it's position within obspack.

An alternative I was thinking could be instead of "subfolder" and "filename", as it is now, this could instead be treated as "uuid"-style identifier within a bucket where the overall identifier could include folder structure. So this could be something like "site-insitu/ch4_mhd_10m_site-insitu_v1.nc" and that could be used to create any folder structure specified by this. However, I haven't quite worked out the full potential flow for that yet.

openghg/datapack/_obspack.py
rt17603: That's been updated now.

openghg/datapack/_obspack.py
rt17603: I've updated a few of the names on the method now to remove "obspack" - hopefully that makes this a bit more clear.

openghg/datapack/_obspack.py
rt17603: This should now have been updated.

openghg/datapack/_obspack.py
rt17603: I kept this as `define_full_obspack_filename` for the moment as this is the function which creates the filename with the full path including the obspack_name and output_folder. I think that keeps the terminology consistent for now but you're right this for a `StoredData` object.

openghg/datapack/_obspack.py
rt17603: Missed this comment before - this should now have been updated.

brendan-m-murphy:
I don't think the subfolders would be required in the set up I recommend. The top level obs pack can contain "stored data" and "subfolder" objects. The subfolder objects can contain more stored data and/or subfolder objects. Everything gets written recursively, passing its path to the next level. None of the objects need to know about the path of the objects that contain them until the "write" method is called. (The write method doesn't need to create nested directories necessarily, you could create longer filenames recursively instead.)

rt17603:
I applied what I feel are the necessary changes for this - there's a few things I haven't done (like add a property or create the SubFolder object) as I'm not sure they're necessary for this at the moment. Overall, this refactor should improve the structure and hopefully make it easier to expand when we need to.

openghg/datapack/_obspack.py
```diff
@@ -73,985 +83,717 @@ def __init__(
...
+        if self.filename is None:
+            logger.info("Creating filename before writing to file.")
+            self.update_filename(
+                include_obs_type=include_obs_type,
+                include_version=include_version,
+                data_version=data_version,
+                name_components=name_components,
+                name_suffixes=name_suffixes,
             )
+            self.filename = cast(Path, self.filename)
```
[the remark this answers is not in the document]
rt17603: Yeah, if I remove the cast `mypy` raises an error:
```
openghg/datapack/_obspack.py:219: error: Argument 1 to "define_full_obspack_filename" has incompatible type "Path | None"; expected "str | Path"  [arg-type]
```

openghg/datapack/_obspack.py
```diff
@@ -57,9 +68,8 @@ def __init__(
         self,
         data: ObsOutputType,
         obs_type: str = "surface-insitu",
-        obspack_path: pathType | None = None,
-        obspack_filename: pathType | None = None,
-        subfolder: MultiSubFolder = None,
+        subfolder: MultiSubFolder | None = None,
```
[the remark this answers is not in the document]
rt17603: Should now be done

openghg/datapack/_obspack.py
```diff
@@ -1320,66 +950,40 @@ def create_obspack(
             logger.exception(msg)
             raise ValueError(msg)
 
-    retrieved_data = retrieve_data(search_df=search_df, store=store)
-
-    unique_obs_types = search_df["obs_type"].unique()
+    obspack = ObsPack(output_folder, obspack_name)
```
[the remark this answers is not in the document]
rt17603: I've added in the options for creating the obspack_name using the obspack_stub directly to the __init__ when creating the object. In this case I don't think we need to add obspack_stub as an attribute - perhaps better not to do this to avoid duplication with obspack_name (and so potentially include mismatched data)?

openghg/datapack/_specification.py
```diff
@@ -0,0 +1,48 @@
...
+        "surface-flask": get_obs_surface,
+        "column": get_obs_column,
+    }
+
+    get_fn_arguments = {
+        get_obs_surface: {"rename_vars": False},
+        get_obs_column: {"return_mf": False},
+    }
+
+    return get_functions, get_fn_arguments
```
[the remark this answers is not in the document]
rt17603: Ok yeah we could approach it like this - I think this style (though a bit clunkier) makes it a little more obvious what this is doing especially if you're not as familiar with partials. I can update if we think this is essential though.

openghg/datapack/_obspack.py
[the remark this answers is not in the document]
brendan-m-murphy: Yea that makes sense.

PR #1259, opened 2025-03-18
1151 data with period varies cannot be standardised

### LINKED ISSUE #1151: Data with `period = "varies"` cannot be standardised
### What happened?

For some data types and source formats (including footprints in `acrg_org` format), if the period/frequency of the data cannot be inferred, it is set to "varies" by the parser.

This data cannot be added to the object store, because `Datasource` will try to create a "representative date range string". This calls the utility function `relative_time_offset`, which will always throw an error if it is given `period = "varies"`.

See Issue #1137 for traceback.

If `continuous = False` is used and a value for `period` is provided, then the data should be standardised with that period. (This could mean that the value for `period` is not actually correct, thus a strong warning should be issued.)

_No response_

### Relevant log output

_No response_

There are actually two issues here:
1. `period = "varies"` causes an error when `period` is extracted from the metadata from the standardisation process
2. the argument `period` in the `standardise_` and `read_file` functions is only used in limited cases.

### Environment

<details>

</details>

### LINKED ISSUE #1254: Allow `period` to be varies for satellite footprint data
For satellite footprint data, we don't expect this to have a regular frequency and so we shouldn't expect this to adhere to `continuous=True` condition.

Suggestion to add check when standardising the footprint data:
 - if `satellite` key is specified, set `continuous` to False and set `period` as varies
 - add details to the logger that this has been done

**Requires: #1151 to be completed before updating this**

SutarPrasad:
  Added relevant warnings, If time_period is varies, the default is set to "1 second".
  If continuous=false inferred frequency is set to input_frequency

- [x] Closes #1254 
- [x] Closes #1151  (Replace xxxx with the Github issue number)

### DIFF
--- a/openghg/util/_time.py
+++ b/openghg/util/_time.py
@@ -714,8 +714,10 @@ def parse_period(period: str | tuple) -> TimePeriod:
             unit = period.lstrip(value_str).strip()  # Strip found value and any whitespace.
         else:
             value = 1
-            unit = "seconds"
-            logging.warning("For time period 'varies' value is set `1` and unit is set to `seconds`")
+            unit = period
+            if period == "varies":
+                unit = "seconds"
+                logging.warning("For time period 'varies' value is set `1` and unit is set to `seconds`")
 
     offset_naming = time_offset_definition()
 
@@ -832,7 +834,7 @@ def relative_time_offset(
     elif value is None or unit is None:
         raise ValueError("If period is not included, both value and unit must be specified.")
 
-    relative_units = ("weeks", "months", "years")
+    relative_units = ("seconds", "weeks", "months", "years")
 
     if unit in relative_units:
         time_delta = DateOffset(**{unit: value})
--- a/openghg/util/_time.py
+++ b/openghg/util/_time.py
@@ -716,7 +716,7 @@ def parse_period(period: str | tuple) -> TimePeriod:
             value = 1
             unit = period
             if period == "varies":
-                unit = "seconds"
+                unit = "s"
                 logging.warning("For time period 'varies' value is set `1` and unit is set to `seconds`")
 
     offset_naming = time_offset_definition()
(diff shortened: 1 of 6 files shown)


openghg/util/_time.py
```diff
@@ -830,7 +834,7 @@ def relative_time_offset(
     elif value is None or unit is None:
         raise ValueError("If period is not included, both value and unit must be specified.")
 
-    relative_units = ("weeks", "months", "years")
+    relative_units = ("seconds", "weeks", "months", "years")
```
brendan-m-murphy, line 837: Relative units are ones that can't be used to create `pd.Timedelta`, so seconds shouldn't be here.  (If it needs to be here, then maybe some different logic is needed.)

openghg/util/_time.py
```diff
@@ -714,6 +715,9 @@ def parse_period(period: str | tuple) -> TimePeriod:
         else:
             value = 1
             unit = period
+            if period == "varies":
+                unit = "seconds"
```
brendan-m-murphy, line 719: Use `unit = "s"` so this can be used in `pd.Timedelta`?

brendan-m-murphy (changes requested):
I don't think seconds should be part of relative units, but besides that looks good. Maybe use "s" instead of "seconds"?

brendan-m-murphy (approved):
Looks good!

rt17603 (commented):
I know this has been merged now but I do think we should be cautious about adding false information to the metadata, so it would be preferable to set "time_period" to "varies" rather than "1 second" if we don't know the time period. Really, I think the only time the period should be set to 1 second is when a value is needed for the code to run e.g. when creating the representative date time. If we need to read this value from the metadata we could have a similiar function which checks if this is "varies" and substitutes this for a value (e.g. 1 second).

A few other related examples of how we handle similiar examples in the code at the moment:
 - For sampling_period if this is not clear but for a particular source_format we have an idea of what this could be "sampling_period" is included as "not_set" and a new "sampling_period_estimate" value is included in the attributes - https://github.com/openghg/openghg/blob/devel/openghg/standardise/surface/_noaa.py#L343
 -  For inlet where we expect a particular format e.g. "10m" we allow `special_keywords` (e.g. "multiple") which are looked for and then bypass the main logic of the code - https://github.com/openghg/openghg/blob/devel/openghg/util/_inlet.py#L47

I'll open a tidy issue (or add a note to a related issue) so we can track this.

PR #1260, opened 2025-03-18
Update reading of config file to use internal defaults when file is not present in object store

### LINKED ISSUE #1255: Previously made object stores rely on old config details
### What is your issue?

The required and optional keys for all data types have previously been moved into a config file to allow for better transparency on how these are being used when storing data (e.g. see Issue #978).
 - [Default config file](https://github.com/openghg/openghg/blob/devel/openghg/data/config/objectstore/defaults.json)

The problem with this setup is that object stores which have already been created contain a static config file and so don't see any updates which have made to the default setup.

As we're finding we're often needing to make changes to this setup it may be better to rethink this a little to make sure previously made object stores use the latest changes by default.

Suggested approach:
 - [x] Update metakeys lookup  
   - Look for config file in object store first (new name)
      - Look for old name and raise a deprecation warning (don't worry unless you want to customise)
   - Load defaults.json
 - [x] Add easy method for creating a custom config file
   - Copies default.json to object store (store input)
   - Make sure this is named differently than the previous config file (new name)
 - [x] Update object store creation to not add the config file to object store going forward
   - ~Could removing the config file from object stores and using the defaults.json file~ Better to name new file to a different name for backwards compatibility, ignore current files going forward

rt17603:

The keys used for each data_type to distinguish data are defined within a config file. At the moment this config file is copied into an object store when this is first created but this then doesn't get updated if further changes are made to the default.

This PR is to update this workflow:
 - Don't create the config file in an object store by default
 - Still look for a config file in the object store BUT the filename this looks for has now been changed to "_v2" so the file in old object store will no longer be found and accessed.
    - If previous file is detected a warning is included
 - If no config file is found this now uses the internal defaults in openghg
 - It is still possible to create a config file which will be put in the object store but this is now defined as a "custom" file and should be created deliberately using `create_custom_config` (formerly `create_default_config`)

- [x] Closes #1255 

Not included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Dropped `exposure_id` variable for GOSAT data to avoid change in dimension size error raised from `to_zarr`. [PR #1243](https://github.com/openghg/openghg/pull/1243)
 - Added ability to process ModelScenario for Observation and Footprint satellite data. Added `platform` keyword to split the process and added ability to pass `satellite` as argument.[#PR 1244](https://github.com/openghg/openghg/pull/1244)
 
+### Updated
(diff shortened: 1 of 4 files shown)


brendan-m-murphy (approved):
Looks good!

PR #1264, opened 2025-03-18
Setting align_to_obs for footprint satellite data with tolerance

### LINKED ISSUE #1262: Ensure ModelScenario uses the correct logic for footprint_data_merge.
For satellite data NAME footprints are created to exactly match the observation time points and so the time matched footprints should be aligned (e.g. 1 footprint / obs time point) (see e.g. [acrg code - resample is avoided](https://github.com/ACRG-Bristol/acrg/blob/develop/acrg/name/name.py#L780))

SutarPrasad:

- [x] Used reindex_like with tolerance of 1ms and method='nearest' for align_to_obs.
- [x] Also updated relevant sections of code missing obs_column data to be saved as obs during align to obs
- [x] Added tests to call fp_data_merge on the modelscenario with relevant data.

- [x] Closes #1262 (Replace xxxx with the Github issue number)


openghg/analyse/_scenario.py
```diff
@@ -748,7 +748,10 @@ def _align_obs_footprint(
 
         if resample_to is None:
             if align_to_obs:
-                footprint_data = footprint_data.reindex_like(obs_data, method="ffill")
+                if platform == "satellite":
+                    footprint_data = footprint_data.reindex_like(obs_data, tolerance=pd.Timedelta("1ms"))
```
rt17603, line 752: Can we check if we need a `method` input for this as well? And if so can be make this `method="nearest"`?

tests/analyse/conftest.py
```diff
@@ -276,3 +276,30 @@ def data_read():
...
+
+    col_filepath = get_column_datapath("gosat-fts_gosat_20160101_ch4-column.nc")
+    col_fp_filepath = get_footprint_datapath("GOSAT-BRAZIL-column_SOUTHAMERICA_201601.nc")
+    flux_filepath = get_flux_datapath("ch4-all_SOUTHAMERICA_2016_SWAMPS-v32-5_Saunois-Annual-Mean_20160101.nc")
+
+    standardise_column(filepath=col_filepath,
+                       species="ch4",
+                       platform="satellite",
+                       satellite="gosat",
+                       domain="southamerica",
```
rt17603, line 290: The `domain` input here is only meant to be used a shorthand if this is the same as `obs_region`. In this case it isn't the same because `domain` is "southamerica" but `obs_region` is "brazil" which represents a different area (as included in `standardise_footprint`). This means that this should be:
```suggestion
                       obs_region="brazil",
```

tests/analyse/test_scenario.py
```diff
@@ -1580,11 +1580,11 @@ def test_scenario_with_satellite_data():
     species = "ch4"
 
     obs_column = get_obs_column(
-        species=species, max_level = 3, satellite=satellite,
+        species=species, max_level = 3, satellite=satellite, store="user"
```
rt17603, line 1583: Can I check why this is using `store="user"` but in the `conftest.py` some of the data was added to `store="group"`? And for these tests would we be better off keeping the files all in the same object store to minimise aspects that are being tested?

tests/analyse/test_scenario.py
```diff
@@ -1602,6 +1602,33 @@ def test_scenario_with_satellite_data():
     xr.testing.assert_equal(model_scenario.footprint.data, footprint.data)
 
 
+def test_model_scenario_col_fp_data_merge():
+    """This is to test satellite data aligns fp to obs and satifies the fp_data_merge functionality"""
+
+    obs_column_data = get_obs_column(species="ch4", max_level=3,
+                     satellite="gosat", domain='southamerica', store="group")
```
rt17603, line 1609: As above this should be updated to:
```suggestion
                     satellite="gosat", obs_region='brazil', store="group")
```

tests/analyse/test_scenario.py
```diff
@@ -1602,6 +1602,33 @@ def test_scenario_with_satellite_data():
...
+    fp_column_data = get_footprint(domain="southamerica",
+                satellite= 'gosat',
+                store="group")
+    flux_data = get_flux(species="ch4", source="all", domain="southamerica",store="group")
+
+    satellite_scenario = ModelScenario(obs_column=obs_column_data,
```
rt17603, line 1615: We may want to just check that the input for the `platform` value is actually being used.

Within the code this will also look at the [site_info.json on openghg_defs](https://github.com/openghg/openghg_defs/blob/main/openghg_defs/data/site_info.json#L1153) for the `platform` keyword so I *think* it would  be using that extracted value in preference but we should just make sure we understand the logic here.

One thing to try would be not passing the `platform` keyword and seeing if this still produces the expected output.

tests/analyse/test_scenario.py
```diff
@@ -1602,6 +1602,33 @@ def test_scenario_with_satellite_data():
...
+                                             sources="all",
+                                             cache=False)
+
+    assert "mf_mod" in fp_data_merge
+    attributes = fp_data_merge.attrs
+
+    attributes["model"] == "name"
+    attributes["data_type"] == "column"
+    len(attributes["heights"]) == 20
+
```
rt17603, line 1630: If this is a test to check alignment has worked ok, we should probably examine the details of the output timeseries (e.g. `time` coordinate specifically) to make sure this aligns with what we expect.
So if we can extract the input `obs.time` values, we can check the `time` values in the output match to this.

rt17603 (changes requested):
Just a few suggested updates in the tests to make sure we know what we're checking and we know the aligning is working ok.

tests/analyse/test_scenario.py
SutarPrasad: I was in the process of testing ModelScenaio works as expected even when data is in different places. But I can move to store="user"

tests/analyse/test_scenario.py
SutarPrasad: it works only if the platform is specified here.

tests/analyse/test_scenario.py
SutarPrasad: Sure. will add the check

SutarPrasad:
This needs to go in before the current PR. https://github.com/openghg/openghg/pull/1273

tests/analyse/test_scenario.py
rt17603: Ok yes that does sound like a useful test but perhaps we could find a way to test that separately? Just thinking its useful to be specific in tests and its makes it easier to identify what the problem is when it fails.

tests/analyse/test_scenario.py
rt17603: Ok, interesting. Useful to know - thanks!

openghg/analyse/_scenario.py
```diff
@@ -182,6 +182,8 @@ def __init__(
         """
         self.obs: ObsData | ObsColumnData | None = None
         self.obs_column: ObsColumnData | None = None
+        if self.obs_column is not None:
+            platform = platform
```
rt17603, line 186: Can I just check - was this meant to be `self.platform = platform`?

tests/analyse/test_scenario.py
```diff
@@ -1580,11 +1580,11 @@ def test_scenario_with_satellite_data():
     species = "ch4"
```
rt17603, line 1580: Looking at this, I think this test should be split into two - one for testing the failing check and the other checking the passing check

tests/analyse/test_scenario.py
```diff
@@ -1580,11 +1580,11 @@ def test_scenario_with_satellite_data():
...
 
     obs_column = get_obs_column(
-        species=species, max_level = 3, satellite=satellite,
+        species=species, max_level = 3, satellite=satellite, selection="land", store="user"
```
rt17603, line 1583: Can I just ask - is this expecting to not find any data? If I'm mapping from the `standardise_*` inputs in the `conftest.py` we're not setting `selection="land"` for instance? If so, could we add a comment just to make that clear.

tests/analyse/test_scenario.py
```diff
@@ -1580,11 +1580,11 @@ def test_scenario_with_satellite_data():
...
     )
 
     footprint = get_footprint(
-        satellite=satellite, domain=domain, obs_region=obs_region
+        satellite=satellite, domain=domain, obs_region=obs_region, model="cams", store="user"
```
rt17603, line 1587: Similiar question to above - is `model="cams"` included here because we don't expect this to find any data?

rt17603 (commented):
A few quick questions around the tests (and a suggestion to split out one test into two) but otherwise looks good.

tests/analyse/test_scenario.py
SutarPrasad: oh okay, these tests were written for the previous PR. I will separate these.

tests/analyse/test_scenario.py
SutarPrasad: These values are set in conftest.py line 243. Because we are separately testing satellite data for new tests we need to add these additional keywords here to fetch the correct expected data for the particular ModelScenario. As get functions will complain with multiple data for specified initial 3 keywords.

tests/analyse/test_scenario.py
SutarPrasad: These values are set in conftest.py line 270.

tests/analyse/test_scenario.py
rt17603: Ok I see - so the data being used in this test is different than the data that has been added for this PR then?

tests/analyse/test_scenario.py
rt17603: Ok I see that - thanks for pointing it out. I think I was a little confused as "cams" is the model we use for boundary conditions rather than footprints. I think the `model` in the conftest.py should be set to "name" for this but I can see this wasn't set in this PR so perhaps something to look at separately.

Did you know if there was a reason why there's two sets of satellite data added in the `conftest.py` - were these serving different purposes? I'm just wondering if we could get away with using the same data for both tests.

rt17603: (Note this is a suggestion - if it's a pain to do or there's a reason not to, I think we can go ahead and get this merged)

SutarPrasad: Yes, the second scenario data can be the only test as it tests some parts of the previously added PR test(Creation of model scenario to be not None). I added these new tests as they are from the same domain and the same timeline for all of the data types. The conftest with selection=land are from the timeline 2017 and all the fp data and flux data I had were from 2016.

openghg/analyse/_scenario.py
SutarPrasad: this is for platform in that scope only.

rt17603: Ok - I think I'm just not quite understanding what this is doing then? Won't the `platform` keyword have been set by the input to the `__init__` function?

SutarPrasad: oh yes right.

rt17603 (approved):

PR #1267, opened 2025-03-20
Attributes period str as varies

### LINKED ISSUE #1151: Data with `period = "varies"` cannot be standardised

SutarPrasad:
  As discussed create a separate util function checking if its satellite and `period_str` is 1 second then update it to `varies` in the metadata.

- [x] Closes #1151 (Replace xxxx with the Github issue number)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,13 +8,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased](https://github.com/openghg/openghg/compare/0.13.0...HEAD)
 
 ### Added
+
 - Added ability to retrieve ICOS combined Obspack .nc data. [PR #1212](https://github.com/openghg/openghg/pull/1212)
-- Dropped `exposure_id` variable for GOSAT data to avoid change in dimension size error raised from `to_zarr`. [PR #1243](https://github.com/openghg/openghg/pull/1243) [PR #1257](https://github.com/openghg/openghg/pull/1257)
-- Drop `id` coordinate for GOSAT data to avoid merging errors [PR #1257](https://github.com/openghg/openghg/pull/1257)
 - Added ability to process ModelScenario for Observation and Footprint satellite data. Added `platform` keyword to split the process and added ability to pass `satellite` as argument.[#PR 1244](https://github.com/openghg/openghg/pull/1244)
+
+### Updated
+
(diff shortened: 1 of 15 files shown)


CHANGELOG.md
```diff
@@ -13,6 +13,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Drop `id` coordinate for GOSAT data to avoid merging errors [PR #1257](https://github.com/openghg/openghg/pull/1257)
 - Added ability to process ModelScenario for Observation and Footprint satellite data. Added `platform` keyword to split the process and added ability to pass `satellite` as argument.[#PR 1244](https://github.com/openghg/openghg/pull/1244)
 
+### Updated
+- Updated `time_period` to be "varies" instead of "1 second" for non-continuous and period = varies.[#PR 1267](https://github.com/openghg/openghg/pull/1267)
+
 ### Fixed
 - Fixed bugs in ModelScenario for satellite data e.g. requiring max_level as argument [#PR 1261](https://github.com/openghg/openghg/pull/1261)
```
rt17603, line 21: Rather than adding this as "Updated" I'd suggest adding this within the "Fixed" section (as a bug) and would also tie this back to the original PR as well e.g.

    - Fixed bug where `period="varies"` could not be used or set when determining the time period associated with the input data. [#PR 1259](https://github.com/openghg/openghg/pull/1259) and [#PR 1267](https://github.com/openghg/openghg/pull/1267)

rt17603 (commented):
Thanks for putting in this update - one small comment about the CHANGELOG.md but other than that looks good.

rt17603 (approved):
Looks good.

Issue #879, opened 2024-01-25
Improve alignment/resampling for obs with long sampling periods.


brendan-m-murphy:
### What is your issue?

NOTE: this is a quick draft, I will add more detail later. (Or hopefully, @joe-pitt will be able to elaborate)

Currently, GC Werks data is stored as measurements at times corresponding to the start of the sampling period. If this data is resampled, and a sampling period overlaps with more than one resampling "bin", then the data is only counted for the first bin. It would be better if the measurement from each sampling period was split proportionally between any resampling "bins" that overlap with the sampling period.

Another consideration is with less frequent flask measurements: we align footprints to obs by taking a mean over the resampling period. But if a measurement was taken at a specific time point within a 7 day period (for instance), we would only want to select the footprint times that overlap with the sampling period for that measurement, rather than averaging over all 7 days of footprints.

A further issue is that the number of observations produced is not necessarily the same as the number of sampling periods. For some Medusa data, the sampling period is 20 minutes, but only one observation is produced per hour. (The issue here is that we are conflating `sampling_period` with the frequency of the observations.)

joe-pitt:
I think this captures the issue nicely! Just to add an example, the CGO Medusa has the following measurements in the original file (output by GCWerks):
```
time                    mf              mf_repeatability inlet 	sampling_period
2022-12-31 15:56:00     121.106880 	0.220027 	 70 	 	1200
2022-12-31 17:01:00 	120.770119 	0.221671 	 70 	 	1200
```
This original file timestamp represents the middle of the sampling, but when the file is read in (using standardise), half the sampling_period is subtracted so that the timeseries in the object store looks like:
```
time                    mf              mf_repeatability inlet 	sampling_period
2022-12-31 15:46:00     121.106880 	0.220027 	 70 	 	1200
2022-12-31 16:51:00 	120.770119 	0.221671 	 70 	 	1200
```
Really these are integrated measurements which start at the above time, and continue for 20 mins. So for the footprint representing 16:00:00-16:59:59, there are a total of 15 minutes of measurement within this hour, and we would ideally like to take a weighted average as follows:
```
meas_avg = 6/15*meas_point_1 + 9/15*meas_point_2
```
Or alternatively we could take a weighted avg of the footprints corresponding each medusa measurement, but I feel like averaging the measurements is going to be easier to implement in a universal way.

rt17603:
See also Issue #253

Issue #253, opened 2022-02-21
Resampling - check for irregular data (e.g. flask) and resample correctly


rt17603:
For e.g. flask data this is not expected to be on a regular frequency and so resampling the whole time period at once is not always appropriate.

In this case we need to ensure any reindexing and resampling e.g. footprints happens on a point by point basis so we align the appropriate footprint to the data.

rt17603:
This could be approached using [`IntervalIndex`](https://pandas.pydata.org/docs/reference/api/pandas.IntervalIndex.html) (or [`IntervalArray`](https://pandas.pydata.org/docs/reference/api/pandas.arrays.IntervalArray.html)) objects from pandas. These create values with represent a start and end time associated with a data point and can be used as index values.

rt17603:
> This could be approached using [`IntervalIndex`](https://pandas.pydata.org/docs/reference/api/pandas.IntervalIndex.html) (or [`IntervalArray`](https://pandas.pydata.org/docs/reference/api/pandas.arrays.IntervalArray.html)) objects from pandas. These create values with represent a start and end time associated with a data point and can be used as index values.

Was previously able to create an `IntervalIndex` and use [`groupby_bins`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.groupby_bins.html) but there seems to be a bug in pandas which xarray is triggering when trying to call `pd.cut(..., retbins=True)` e.g.

```
import xarray as xr
import numpy as np

ds_irregular_1 = xr.Dataset({"ch4": ("time", np.array([1,2,3,4]))}, 
                  coords={"time": np.array(["2012-01-01T12:00:00", "2012-02-01T12:00:00", "2013-01-01T01:00:00", "2013-01-01T02:00:00"], dtype=np.datetime64)})

sampling_period_1 = 60*60  # seconds

import pandas as pd
from pandas import IntervalIndex

small_offset = pd.Timedelta(milliseconds=1)

start_dates = ds_irregular_1["time"].values - small_offset
end_dates = ds_irregular_1["time"].values + pd.Timedelta(seconds=3600) - small_offset

# Interval index creates a start and end windows equal to the array length. By default this is closed on the right 
interval_index = IntervalIndex.from_arrays(left=start_dates, right=end_dates)

# Sanity check - if we use pd.cut directly on our time axis, do we get back the same time values
pd.cut(ds_irregular_1["time"], bins=interval_index, ret_bins=True)
```

rt17603:
[...]

PR #1285, opened 2025-04-03
Iss1246 create a parser to process flask data for corso

### LINKED ISSUE #1246: Create a parser to process CORSO data
### What is your issue?

Requirements:
ICOS_ATC_L1_FAST_TRACK
These are all flask measurements.
 
For the 14C data please process the following variables for each measurement site:
·         Sampling start
·         Sampling end (or if easier make a new variable called sampling time that is the difference of sampling end and sampling start)
·         14C (this will be “14CO2” in OpenGHG)
·         MeasUnc (this is the measurement uncertainty)
·         EstRep (this is the reproducibility)
Please filter out any data points that have an “N” or “K” in the flags column and make sure the icos_data_level is set to 1.
 
For the co2 data process the following variables for each measurement site:
·         Sampling start
·         Sampling end
·         Co2
·         Stdev (this is the measurement uncertainty)
·         EstRep (this is the reproducibility)
·         ScalLink (this is the calibration uncertainty)
·         CombUnc (this is combined uncertainty for each flask sample)
Filter out any data points that have an “N” or “K” in the flags column and make sure the icos_data_level is set to 1.
 
For the DeltaO2N2 data process the following variables for each measurement site:
·         Sample start
·         Sample end
·         O2 (in units of “per meg”)
·         Stdev
·         EstRep
·         ScalLink
·         CombUnc
Please filter out any data points that have an “N” or “K” in the flags column and make sure the icos_data_level is set to 1.
 
ICOS_ATC_L2
For the 14C flask files, this should be exactly the same as for the 14C data in ICOS_ATC_L1_FAST_TRACK except the icos_data_level can be set to 2.
 
For the 14C integrated sampling files. Include the following variables for each site:
·         Sample start date (It’ll have to be created from the different date columns)
·         Integration time
·         14C
·         WeightedStdErr
·         AnalyticalStdev
·         SamplingPattern
Please filter out any data points that have an “N” or “K” in the flags column and make sure the icos_data_level is set to 2.
 
 
clean_14C_data_v2025
Add data for only these sites:
·         uheicrl_l2_2025_1_jfj_5m_int_14day_clean.c14
·         uheiiup_l2_2025_1_alt2023_20m_int_14day_clean.c14
·         uheiiup_l2_2025_1_alt_20m_int_14day_clean.c14
·         uheiiup_l2_2025_1_mhd_24m_int_14day_clean.c14
 
NB. Please do not share these data with anyone as they are unpublished!
 
For each of these include the following variables for each site
·         Time
·         Integration time
·         14C
·         WeightedStdErr
·         AnalyticalStdErr
·         SystematicalUncert
·         SamplingPattern
Please filter out any data points that have “N” or “K” in the flags column.

SutarPrasad:
Added icos_corso parser.
added util function at the bottom of the file.
added standardisation tests for each of the data variant of the corso data

- [x] Closes #1246 (Replace xxxx with the Github issue number)


openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,414 @@
...
+        metadata["data_owner_email"] = data_owner_email
+    else:
+        f_header = [s for s in header if "CONTACT POINT EMAIL" in s]
+        if len(f_header) == 1:
+            data_owner_email = f_header[0].split(":")[1].strip()
+            metadata["data_owner_email"] = data_owner_email
+        else:
+            raise ValueError("Couldn't identify data owner email")
+
+    if sampling_period is None:
```
rt17603, line 274: For sampling period we should be able to extract more detail from the files - we actually won't have one sampling_period for all data measurements though so would probably be best to include the sampling period details as a `{species}_sampling_period` variable. We can still include a sampling_period attribute for consistency but we can call this "multiple" as a special keyword.

Having a separate sampling_period variable actually isn't a fully supported workflow yet but looking at this data and the long integration time (14 days) I think we're going to need to update some of the steps when we're aligning obs and footprints for this data anyway - links to #879 and #253 . If we can get these details included within the data stored in the object store then we can work on the next steps.

rt17603 (commented):
Looks good overall!

Looking over the data I realised this data is an example with multiple sampling periods and so it would be good to make sure we're incorporating these details when adding the data. I added a comment to the issue #1246 on how we could approach this looking at the details included in the data.

openghg/standardise/surface/_icos_corso.py
rt17603: Added issue #1288 to help track this.

SutarPrasad: Added a data var for {species}_sampling_period wherever applicable. For data which has integration time and days mentioned, I have updated the sampling_period to be multiple instead. Ready for review

rt17603: As commented in #1246 , in addition to the (end - start) calculation for sampling period, for the other files we can use the "integrationtime" column to create the sampling period variable.

For this, the best approach would probably be to convert this into seconds and set the "units" attributes for the variable to be seconds.

If there are any issues with this number being too big we can keep this as days as long as also add units="d" as the units attribute (but we would then have to make sure this was correctly interpreted down the line).

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,443 @@
...
+        f_header = [s for s in header if "TIME INTERVAL" in s]
+        interval_str = f_header[0].split(":")[1].strip()
+        if interval_str == "hourly":
+            metadata["sampling_period"] = "3600.0"
+        else:
```
rt17603, line 291: I liked the way you were doing this previously where you were checked for the words "integrated sampling" before setting the "sampling_period" value. We could have this as an if-elif-else block where for the else we raise an error and ask the user to specify a sampling_period directly.

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,443 @@
...
+            "time",
+            "sampling_start_date",
+            "integration_time",
+            "weighted_std_err",
+            species + "_variability",
+            "sampling_pattern",
+            "flag",
+        ]
+
+        df = df = df[columns_to_keep]
```
rt17603, line 240: ?
```suggestion
        df = df[columns_to_keep]
```

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,443 @@
...
+        species_fname.lower(): species,
+        "samplingstart": "sampling_start",
+        "samplingend": "sampling_end",
+        "measunc": species + "_variability",
+        "estrep": species + "_repeatability",
+        "nbpoints": species + "_number_of_observations",
+        "stdev": species + "_variability",
+        "scallink": species + "_calibration_uncertainty",
+        "combunc": species + "_combined_uncertainty",
+        "integrationtime": "integration_time",
```
rt17603, line 84: If the "integrationtime" column is there we can use this as the species + "_sampling_period" (or "sampling_period") data variable.

As discussed earlier a step further on, we can convert this into seconds and set the "units" attributes for the variable to be "s" as well. (Though can keep the units as days if the seconds value seems unreasonably big)

openghg/standardise/surface/_icos_corso.py
rt17603: To clarify on this, I was suggesting checking "integrated sampling" was written in the header rather than including an `else:`
```suggestion
        elif "integrated sampling" in interval_str:
```

rt17603 (commented):
I've added a few related comments at a few points in the code but for checking and including the sampling period the overall the steps could be:
 - check header for sampling period value - set sampling_period attr if value is found
 - extract appropriate columns and calculate sampling period for each value (end-start or integration time column) - add as {species}_sampling_period (or just sampling_period) data variable
 - use this calculated sampling_period column to check unique sampling period values
   - if unique - set sampling_period to value in seconds
   - if not unique set attr to "multiple"

openghg/standardise/surface/_icos_corso.py
SutarPrasad: did not convert the values to seconds as they will be big in this case of integrated sampling, the minimum of 14 days is the interval time.

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,451 @@
...
+            data_owner_email = f_header[0].split(":")[1].strip()
+            metadata["data_owner_email"] = data_owner_email
+        else:
+            raise ValueError("Couldn't identify data owner email")
+
+    if sampling_period is None:
+
+        f_header = [s for s in header if "TIME INTERVAL" in s]
+        interval_str = f_header[0].split(":")[1].strip()
+        if interval_str == "hourly":
```
rt17603, line 280: There's essentially two things we can do to check the sampling period (1) check the header and (2) check the data variables. However, these two things aren't completely disjointed. This is because for some of the files when we have "integrated sampling" as a part of the header this seems to also be tied to being able to determine the sampling period details from the data columns themselves.

Given that, the elif associated with "integrated sampling" in interval_str.lower() will likely always trigger before looking at the data variables - I think we can actually skip this because it's more important if the sampling_period column is included in the data at all than "integrated sampling" being in the header.

With the current data we've seen, I think this could be structured as something like:
```
if interval_str == "hourly":
    metadata["sampling_period"] = "3600"
elif f"{species}_sampling_period" in data.data_vars:
	rounded_values = np.round(data[f"{species}_sampling_period"].values, decimals=2)
	unique_values = np.unique(rounded_values)

	if len(unique_values) == 1:
		metadata["sampling_period"] = unique_values.astype("str")
	else:
		metadata["sampling_period"] = "multiple"
else:
	missing_err_msg = "Unable to determine sampling_period from input header/data. Please specify value in seconds."
	logger.exception(missing_err_msg)
	raise Exception(missing_err_msg)
	# Would actually be nice to add a MetadataMissingError() for this as another error type
```

Note I omitted the check for 3600, 3601 from here at the moment but wanted to check why this was needed? Or if there was a way to include this as a more generic check? This could mean checking the values and units for the "{species}_sampling_period" data variable and finding a way to allow for a tolerance of e.g. 1 second
```
	if [3600, 3601] in unique_values and len([3600, 3601]) == len(unique_values):
		metadata["sampling_period"] = "3600"
```

rt17603 (changes requested):
I think the sampling_period logic just needs a tweak to make sure this will trigger at the right times with the data examples we've seen - details included below.

openghg/standardise/surface/_icos_corso.py
SutarPrasad: I added MetadataMismatchError and also np.all with tolerance.

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,453 @@
...
+        ]
+
+        df = df[columns_to_keep]
+
+    df = clean_dataframe(df=df, species_name=species)
+    df = calculate_sampling_period(dataframe=df, species=species)
+
+    data = df.to_xarray()
+    # setting units to sampling period data var
+    data[f"{species}_sampling_period"].attrs["units"] = "s"
```
rt17603, line 240: As this is written now, I don't the "{species}_sampling_period" units would always be in seconds. If the "{species}_sampling_period" column was taken directly from the integration time this would be in days.

One suggested approach could be:
 - keep "integration_time" naming for column initially
 - use "integration_time" to get "{species}_sampling_period" within your `calculate_sampling_period()` function by taking this column and converting from days --> seconds

I think having values on the order of ~1e9 values for the sampling period is probably fine and at least this would then be consistent for when you wanted to add the units as "s".

SutarPrasad: I have gone with the second approach, partially where {species}_sampling_period will be present converted from integration_time if it is not a flask file. And converted it to seconds.

rt17603: The suggested approach was actually to do both steps - since "integration_time" is a specific name that's coming from the file we can probably trust this will be in units of days (for now). I think we probably want to be a bit careful about assuming "{species}_sampling_period" is generally in days as we could decide later to derive this from another variable etc.

SutarPrasad: Based on the data study conducted, it is evident that files containing the integration_period column are renamed to "species_sampling_period" in the early stages of processing. For files that do not include this column, a calculation is required before generating the "species_sampling_period".

One potential approach could be to check for the presence of the word "days" in the file’s long header. This may serve as a helpful indicator, as files with "flask" in their names are unlikely to include this term. While this method may not be entirely precise, it provides a reasonable means of confirming that the "species_sampling_period" — generated after renaming integration_time — contains values in days, justifying the conversion from days to seconds.

However, I do have some concerns that this approach could introduce additional processing overhead, potentially impacting performance. Nevertheless, it appears to be the most reliable method to ensure that the "species_sampling_period" is in days before proceeding with the conversion.

rt17603: Ideally, we should always associate numerical data with a unit. Because we're initially processing this from a DataFrame --> Dataset we don't have that option as easily when the data is originally read in. My suggestion is just that keeping "species_sampling_period" to be consistently in seconds helps with this aim.

Secondarily to that, we can (but don't have to assume) that "integration time" in is in days - a reasonable assumption for this flask data for the moment and given the data. We could include something which checks that "integration time" represents days - we can do this because we know the header includes those details (details were added as a comment on #1246 ) but regardless, for clarity, in the code I'm just suggesting making sure we're not mixing and matching values in different units when we have no way to store that initially.

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,465 @@
...
+        species_fname.lower(): species,
+        "samplingstart": "sampling_start",
+        "samplingend": "sampling_end",
+        "measunc": species + "_variability",
+        "estrep": species + "_repeatability",
+        "nbpoints": species + "_number_of_observations",
+        "stdev": species + "_variability",
+        "scallink": species + "_calibration_uncertainty",
+        "combunc": species + "_combined_uncertainty",
+        "integrationtime": species + "_sampling_period",
```
rt17603, line 84: Suggestion to change this back to what you had previously:
```suggestion
        "integrationtime": "integration_time",
```

openghg/standardise/surface/_icos_corso.py
```diff
@@ -0,0 +1,465 @@
...
+        if "sampling_end" in columns:
+            dataframe[f"{species}_sampling_period"] = (
+                dataframe["sampling_end"] - dataframe["sampling_start"]
+            ).dt.total_seconds()
+        else:
+            ValueError(
+                f"Unable to find sampling_start, sampling_end, or  {species}_sampling_period in the data to convert the time values into seconds"
+            )
+    else:
+        dataframe[f"{species}_sampling_period"] = pd.to_timedelta(
```
rt17603, line 461: If the column is named "integrationtime" or "integration_time" these lines can be updated to use this column to create the sampling period in seconds after the for this being in days:
```
    else:
        dataframe[f"{species}_sampling_period"] = pd.to_timedelta(
            dataframe["integration_time"], unit="D"
        ).dt.total_seconds()
```
The additional

rt17603 (commented):
From reviewing the ability to track the sampling period units, I would still suggest reverting back to keeping the integration time name (rather than renaming to sampling_period initially). Then within `calculate_sampling_period` the "integration_time" column  header details can be checked and this then converted to seconds for the sampling period - explicit details included in the code.

Without the ability to easily track the units, converting "integration_time" --> "sampling_period" at the start (which I know was one of my initial suggestions) means we're having to make assumptions later on that the "sampling_period" column was originally the "integration_time" column and I just think it's more consistent and clear to do this in one place rather than two.

openghg/standardise/surface/_icos_corso.py
rt17603: Superseded by other, more direct suggestions.

openghg/standardise/surface/_icos_corso.py
rt17603: Also, superseded by later suggestions

PR #1289, opened 2025-04-04
Formalise and format platform values

rt17603:

### DIFF
--- a/openghg/store/_obscolumn.py
+++ b/openghg/store/_obscolumn.py
@@ -97,6 +97,7 @@ class ObsColumn(BaseStore):
         from openghg.store.spec import define_standardise_parsers
         from openghg.util import (
             clean_string,
+            format_platform,
             load_standardise_parser,
             split_function_inputs,
             check_if_need_new_version,
@@ -106,6 +107,8 @@ class ObsColumn(BaseStore):
         # TODO: Evaluate which inputs need cleaning (if any)
         species = clean_string(species)
         species = synonyms(species)
+
+        platform = format_platform(platform)
         platform = clean_string(platform)
 
         if site is None and satellite is None:
--- a/openghg/util/__init__.py
+++ b/openghg/util/__init__.py
@@ -47,7 +47,7 @@ from ._metadata_util import (
     get_overlap_keys,
     merge_dict,
 )
-from ._platform import get_platform_from_info
+from ._platform import define_platform, format_platform, get_platform_from_info
 from ._registry import AutoRegistry, Registry
 from ._site import get_site_info, sites_in_network
 from ._species import (
--- a/openghg/util/_platform.py
+++ b/openghg/util/_platform.py
@@ -1,5 +1,105 @@
+from typing import overload
+import logging
 from openghg.util._site import get_site_info
-from openghg.types import optionalPathType
+from openghg.types import optionalPathType, MetadataFormatError
+
+logger = logging.getLogger("openghg.util")
+logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
+
+
+def define_platform(data_type: str | list | None = None) -> list:
+    """
+    Define platform values which can be used. These describe the type of data and
+    platform used to measure the data.
+
+    Args:
+        data_type: Specify the data_type to select the associated platforms.
+            For "surface" this is currently:
+                - "surface-insitu"
+                - "surface-flask"
+            For "column" this is currently:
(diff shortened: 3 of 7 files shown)


openghg/util/_platform.py
```diff
@@ -1,5 +1,105 @@
...
+            This is generally associated with observation measurements.
+            These are defined within openghg.store.spec.define_data_types function.
+    Returns:
+        str | None: checked platform value
+    """
+
+    platform_values = define_platform(data_type=data_type)
+    not_set_values = not_set_metadata_values()
+
+    if platform is not None:
```
brendan-m-murphy, line 92: Maybe invert some of these conditionals and return as soon as we know what to do. For example: 

```
if platform is None or platform in not_set_values:
    return platform
if platform.lower() in platform_values:
    return platform.lower()
<raise error>
```

openghg/util/__init__.py
```diff
@@ -47,7 +47,7 @@
     get_overlap_keys,
     merge_dict,
 )
-from ._platform import get_platform_from_info
+from ._platform import define_platform, format_platform, get_platform_from_info
```
brendan-m-murphy, line 50: Maybe we need a `metadata` submodule... we already have `_metadata_util` in `util`, and these new functions are also metadata related. This would probably be a separate tidying PR though.

openghg/store/_obscolumn.py
```diff
@@ -106,6 +107,8 @@ def read_file(
         # TODO: Evaluate which inputs need cleaning (if any)
         species = clean_string(species)
         species = synonyms(species)
+
+        platform = format_platform(platform)
         platform = clean_string(platform)
```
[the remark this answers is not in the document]
rt17603: At the moment this does keep it consistent with what we do in `ObsSurface.read_file` for different input metadata strings - we are generally including formatting + clean_string at the moment.

e.g. L257-L270
```
...
data_level = format_data_level(data_level)
        if data_sublevel is not None:
data_sublevel = str(data_sublevel)

platform = check_and_set_null_variable(platform)
data_level = check_and_set_null_variable(data_level)
data_sublevel = check_and_set_null_variable(data_sublevel)
dataset_source = check_and_set_null_variable(dataset_source)
platform = clean_string(platform)
data_level = clean_string(data_level)
data_sublevel = clean_string(data_sublevel)
dataset_source = clean_string(dataset_source)
...
```

openghg/util/_platform.py
```diff
@@ -1,5 +1,105 @@
...
+def format_platform(platform: str, data_type: str | None = None) -> str: ...
+
+
+@overload
+def format_platform(platform: None, data_type: str | None = None) -> None: ...
+
+
+def format_platform(platform: str | None, data_type: str | None = None) -> str | None:
+    """
+    Check platform is a suitable value based on the values within define_platform.
```
rt17603, line 78:
```suggestion
    Check platform is a suitable value based on the values within define_platform.
    This could change the case of the platform value to match to the define_platform values if the original case does not match.
```

openghg/util/_platform.py
rt17603: So I separated out
- if platform not in platform_values
from 
- if platform.lower() in platform_values
because I wanted to allow this *in principle* to have a different case than all lower. It's not a big deal but I was trying to ensure this didn't have to be lowercase.

openghg/util/__init__.py
rt17603: I think we could get the best of both really - I like this being split out in such an obvious way but as we've discussed before around the idea of centralised formatting we could have a more central place which collects together these format steps.

brendan-m-murphy: Yea we could have `metadata/_platform.py`, `metadata/_common.py` etc. I just think maybe the metadata functions are out-growing `util` a bit.

brendan-m-murphy (dismissed):
Looks good!

I'd prefer the nested if-statements to be flattened out a bit, but it's okay to leave it if you prefer.

openghg/util/_platform.py
rt17603: Yeah that's reasonable - I think I was just trying to avoid including lots of if... `return` statements. I'll have a look at implementing a more flattened version.

rt17603: I've updated this now to be a bit more flat - hopefully makes it more readable.

brendan-m-murphy (approved):
Okay, looks good!

Issue #1226, opened 2025-02-27
Umbrella: Ensure satellite data can be used in inversions


rt17603:
### Is your feature request related to a problem?

We need to be able to add and run inversions with column data and so need to be able to add these to the object store.

#### Satellite

We already have satellite data and should ensure we can able to run with the same inversion set up as previous.

***Satellite (and some joint with site-column)***

- [x] Add appropriate metadata keywords for column and associated column objects - #1215 , #1216 
- [x] Make sure data can be standardised and added to the object store
    - [x] Review internal `ObsColumn` format and update if needed (currently based on gosat)
    - [x] Ensure parse framework is present for the `ObsColumn` data type
    - [x] Create/update `parse_` function(s) to be appropriate for this
        - [x] Update GOSAT parse function to work with new zarr framework - resolve issue on `exposure_ids` - (https://github.com/openghg/openghg/issues/1240)
- [x] Ensure logic within [`get_gosat` code](https://github.com/ACRG-Bristol/acrg/blob/develop/acrg/obs/read.py#L636) from acrg code is implemented in openghg (logic could simply be transferred to equivalent function e.g. within [`get_obs_column` function](https://github.com/openghg/openghg/blob/devel/openghg/retrieve/_access.py#L469))
- [x] Update ModelScenario to be able to access appropriate column data and associated products
    - [x] Split the process by `platform` ("satellite") and create new `get_obs_column` and `get_satellite_footprint` methods for this. - link to PR #1244 
- [x] #1262
    - [x] For satellite data NAME footprints are created to exactly match the observation time points and so the time matched footprints should be aligned (e.g. 1 footprint / obs time point) (see e.g. [acrg code - resample is avoided](https://github.com/ACRG-Bristol/acrg/blob/develop/acrg/name/name.py#L780))
- [x] Ensure openghg_inversions can run with satellite data
    - [x]  Create ini example file to run with original GOSAT data using the same setup as previously (e.g. ch4, gosat-brazil-land data and footprints, emissions and bc for SOUTHAMERICA domain, 1 month in 2014 - all data should be within shared area)
    - [x] Raise and fix issues as needed
        - [x] https://github.com/openghg/openghg/issues/1258
        - [x] https://github.com/openghg/openghg/issues/1151
        - [x] https://github.com/openghg/openghg/issues/1254
- [x] Add tutorial for creating Model Scenario with satellite data.(https://github.com/openghg/openghg/pull/1304)
[...]

PR #1291, opened 2025-04-04
Simplify ModelScenario alignment around platform

rt17603:

This *should* be a neutral change which just refactors some steps to avoid duplication.

### What has been updated

This PR is to tidy up the alignment steps in ModelScenario based on platform as I realised we were accidentally duplicating a few steps. It does this by:
 - moving the platform logic into the overall `self.combine_obs_footprint` function
 - renaming the `align..` function (and variables) to `resample..` to make this distinction between steps more clear
 - moving the generic `_indexes_match` and `combine_datasets()` function to the `_alignment.py` submodule since these functions are also very much about alignment.
 - updating use of `combine_datasets()` to set tolerance and method values differently depending on platform.
    - platform="satellite" --> tolerance="1ms", method="nearest"
    - other platforms --> tolerance=None (i.e. don't set a tolerance), method="ffill"

### Reasoning behind this

We have made changes recently to the way data is resampled and aligned based on platform including original PR #1236 & #1264 which built on this. This considered the fact that for some platforms we only want to align the data and not resample and changes were made within the [`self._align_obs_footprint(...)`](https://github.com/openghg/openghg/blob/devel/openghg/analyse/_scenario.py#L833) function to account for this.

However, within [`self.combine_obs_footprint`](https://github.com/openghg/openghg/blob/devel/openghg/analyse/_scenario.py#L789) the resampling and overall alignment was **already happening in two different steps** which wasn't necessarily clear. This function was:
1. Calling [`self._align_obs_footprint(...)`](https://github.com/openghg/openghg/blob/devel/openghg/analyse/_scenario.py#L833) was the main resampling step
2. Calling the generic [`combine_datasets()` function](https://github.com/openghg/openghg/blob/devel/openghg/analyse/_scenario.py#L834C28-L834C44) which would align the two datasets

This means that the changes made were now sometimes duplicating these steps as we were completing these updates within `self._align_obs_footprint(...)` AND `combine_datasets()` for some platforms.

- [x] [Tests passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature
- Related to #1226 

Not updated
- ~Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature~


openghg/analyse/_alignment.py
```diff
@@ -243,3 +244,74 @@ def align_obs_and_other(
...
+        dataset_B_temp = dataset_B
+    else:
+        # load dataset_B to avoid performance issue (see xarray issue #8945)
+        dataset_B_temp = dataset_B.load().reindex_like(dataset_A, method=method, tolerance=tolerance)
+
+    merged_ds = dataset_A.merge(dataset_B_temp)
+
+    if "fp" in merged_ds:
+        if all(k in merged_ds.fp.dims for k in ("lat", "lon")):
+            flag = np.where(np.isfinite(merged_ds.fp.mean(dim=["lat", "lon"]).values))
```
brendan-m-murphy, line 314: maybe use `xarray.Dataset.where`:
```
flag = np.isfinite(merged_ds.fp.mean(["lat", "lon"])
merged_ds = merged_ds.where(flag, drop=True)
```

openghg/types/__init__.py
```diff
@@ -41,6 +41,7 @@
     ArrayLikeMatch,
     XrDataLike,
     XrDataLikeMatch,
+    methodType,
```
brendan-m-murphy, line 44: Maybe call this `ReindexMethod`? This is what pandas calls it: https://github.com/pandas-dev/pandas/blob/0691c5cf90477d3503834d983f69350f250a6ff7/pandas/core/indexes/base.py#L3820-L3953 (the xarray options come from pandas)

openghg/analyse/_scenario.py
```diff
@@ -755,36 +754,23 @@ def _align_obs_footprint(
...
         footprint_data = footprint.data
 
-        # Keyword only used if not resampling (at the moment)
-        align_to_obs = True
-
         if platform is not None:
             platform = platform.lower()
             # Do not apply resampling for "satellite" or "flask"
             if platform == "satellite":
                 resample_to = None
```
brendan-m-murphy, line 761: Possibly ignore this in favour of the comment to move logic to `combine_obs_footprint`: Rather than using `resample_to = None` as a flag, you could just return the obs and footprint data here. (Although this is probably easier to see what is going on outside of GH's diff mode, I found this slightly confusing for a moment.)

Also, do we need to do slicing when platform is not None? That is part of `resample_obs_footprint`, although the `get_` functions slice data already, so maybe it isn't needed. (Maybe we need a separate issue to review slicing in `ModelScenario`, I think this has been a point of confusion in the past, mostly around how fluxes are filled/sliced.)

openghg/analyse/_scenario.py
```diff
@@ -755,36 +754,23 @@ def _align_obs_footprint(
         obs_data = obs.data
         footprint_data = footprint.data
 
-        # Keyword only used if not resampling (at the moment)
-        align_to_obs = True
-
         if platform is not None:
```
brendan-m-murphy, line 757: I think maybe this logic should be in `combine_obs_footprint`, since there is already logic to decide how to align there based on platform, and this function doesn't do anything other than get the data from `obs` and `footprint` if platform is not `None`.

Maybe just use a `resample` True/False flag in `combine_obs_footprint`, and then `_resample_obs_footprint` would just be a thin wrapper that substitutes "other" for "footprint"

brendan-m-murphy (changes requested):
Looks good. I think using `ReindexMethod` instead of `methodType` would be better because that is the name pandas uses for the same type alias (and `methodType` is a bit vague without the context of aligning data).

Also, maybe the logic of deciding what to do based on platform should all be in `combine_obs_footprint`, even if this makes `_resample_obs_footprint` just a thin wrapper.

openghg/analyse/_alignment.py
rt17603: Unfortunately doesn't seem to be a simple switch to use these steps in xarray as this seems like this would need a few more steps to deal with dask.

For instance, when I run this the test return:
> KeyError: 'Indexing with a boolean dask array is not allowed. This will result in a dask array of unknown shape. Such arrays are unsupported by Xarray.Please compute the indexer first using .compute()'

openghg/analyse/_scenario.py
rt17603: Yeah ok - you're right we could take this a little further. I've reorganised this to include all `platform` checks within `combine_obs_footprint` now and `_resample_obs_footprint` doesn't accept `platform`.

For now I've left this as `resample_to=None` in `_resample_obs_footprint` just returns `obs, footprint` as a generic use case which this makes use of.

openghg/analyse/_scenario.py
rt17603: I think the first point is addressed in the restructure mentioned in the previous comment.

For the second point, which part of the code are you referring to when you say "do we need to do slicing when platform is not None"? The code deliberately doesn't apply hard slicing in some cases such as co2 data as we need flux data from before start_time and this was a way to make sure we had the right data available when were aligning this.

openghg/analyse/_alignment.py
brendan-m-murphy: Oh I forgot about that (it was a problem in inversions at some point). You would need to do `flag.compute()` before using it in `where`, e.g. `ds.where(flag.compute(), drop=True)`.

openghg/analyse/_scenario.py
brendan-m-murphy: > For the second point, which part of the code are you referring to when you say "do we need to do slicing when platform is not None"? The later reindex step should cover making sure footprint data only covers the same times as obs for instance?

Ah okay I didn't realise the reindex would cover it. I should just check an example... but looking at the xarray source code it looks like you might need to drop NaNs to actually remove the obs. values that didn't align with the footprints when you use `reindex_like`. Anyway, it's kind of a moot point for obs, since `get_obs_surface` will already slice. (Unless the `add_` function for model scenario differs here...)

openghg/analyse/_alignment.py
brendan-m-murphy: Another issue here is that `fp.mean(...)` will ignore NaNs, so the `isfinite` will only detect `np.inf`. Is this expected?

rt17603: > Another issue here is that fp.mean(...) will ignore NaNs, so the isfinite will only detect np.inf. Is this expected?
This could have been to remove infinite values which could have been introduced by 0 division but this code is quite old so we may want to evaluate what it's original purpose was.

As this is quite a key bit of code (and I think this may be starting to move a bit of out of the scope for this PR) I've put this in as a separate PR #1292

brendan-m-murphy (approved):
Looks good!

PR #1295, opened 2025-04-09
change monthly_s variable in openghg/standardise/surface/_noaa

Brelage:
changed the monthly_s variable to be calculated by multiplying the daily_s variable by 28. 
Before, this variable was calculated by multiplying the weekly_s variable by 28, meaning that one month was calculated as 28 weeks instead of 28 days.


brendan-m-murphy:
@Brelage thanks for finding this! Can you merge devel into your branch and push again? One of our dependencies had an update that breaks the tests. This is fixed on devel now though.

brendan-m-murphy (approved):
Looks good, thanks!

PR #1298, opened 2025-04-10
Fix issue 1297

### LINKED ISSUE #1297: Resampling drops all the data when `"{species}_variability"` is almost all nan even though  `"{species}_repeatability"` has none
### What is your issue?

I have a case (`site= 'MHD', species= 'n2o', inlet= slice(0, 20, None), start_date= '2018-04-01', end_date= '2018-05-01', store= 'eye_ave_par_2p2_store'`) where there are the variables `"n2o_variability"`, `"n2o_repeatability"` and `"n2o_number_of_observations"`. But... `"n2o_variability"`is almost always `nan` so every data are dropped at https://github.com/openghg/openghg/blob/a75e2b52957f4d585854844cb0d9508e638dc0be/openghg/data_processing/_resampling.py#L408 when `drop_na = True`.

I would like openghg to see that it should focus on `"n2o_repeatability"` rather than `"n2o_variability"` in that case.
Another solution would be to replace the `nan` with zeros for those 2 variables as a zero uncertainty is not physical and should raise an error down the road.

alexdanjou:
Replace nans with zeros for variables "{species}_variability" and "{species}_repeatability" during resampling to prevent dropping of all data when one of this variables is almost filled with nans.

- [X] Closes #1297

### DIFF
--- a/openghg/data_processing/_resampling.py
+++ b/openghg/data_processing/_resampling.py
@@ -405,9 +405,9 @@ def resampler(
     result = apply_funcs(ds, funcs, func_vars, **apply_func_kwargs)
 
     for dv in result.data_vars:
-        if 'variability' in str(dv) or 'repeatability' in str(dv):
+        if "variability" in str(dv) or "repeatability" in str(dv):
             result[dv] = result[dv].fillna(0.0)
-            
+
     if drop_na:
         result = result.dropna("time")
 
--- a/openghg/data_processing/_resampling.py
+++ b/openghg/data_processing/_resampling.py
@@ -404,6 +404,10 @@ def resampler(
 
     result = apply_funcs(ds, funcs, func_vars, **apply_func_kwargs)
 
+    for dv in result.data_vars:
+        if 'variability' in str(dv) or 'repeatability' in str(dv):
+            result[dv] = result[dv].fillna(0.0)
+            
     if drop_na:
         result = result.dropna("time")
 
--- a/openghg/data_processing/_resampling.py
+++ b/openghg/data_processing/_resampling.py
@@ -405,13 +405,15 @@ def resampler(
     result = apply_funcs(ds, funcs, func_vars, **apply_func_kwargs)
 
     if drop_na:
-        if 'species' in kwargs.keys():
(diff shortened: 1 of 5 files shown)


openghg/data_processing/_resampling.py
```diff
@@ -405,7 +405,13 @@ def resampler(
     result = apply_funcs(ds, funcs, func_vars, **apply_func_kwargs)
 
     if drop_na:
-        result = result.dropna("time")
+        if 'species' in kwargs.keys():
```
brendan-m-murphy: Maybe this can go in `surface_obs_resampler`. Then you don't have to check for `species`. (Plus this is pretty surface obs specific? Or does satellite data have these same variable names?)

alexdanjou: Very good point!

openghg/data_processing/_resampling.py
```diff
@@ -405,7 +405,15 @@ def resampler(
     result = apply_funcs(ds, funcs, func_vars, **apply_func_kwargs)
 
     if drop_na:
-        result = result.dropna("time")
+        if "species" in kwargs.keys():
```
brendan-m-murphy: Maybe this can go in `surface_obs_resampler`? Then you don't need to check for `species`. (Assuming these variable names are specific to surface data... I'm not sure if they're used elsewhere.)

e.g. like
```
result = resampler(...)
if drop_na:
    ...
return result
```

brendan-m-murphy (commented):
Maybe the new code is better placed in `surface_obs_resampler` (unless these variable names work for, say, satellite data).

I'll see if I can write a test where half the repeatabilities and half the variabilities are NaN, and see if it keeps all the data.

openghg/data_processing/_resampling.py
alexdanjou: New push made. `resampler` is now the same as before. all the changes have been moved to `surface_obs_resampler`

openghg/data_processing/_resampling.py
alexdanjou: Just made a new push with this ! (Didn't add a test though)

brendan-m-murphy (approved):
Looks good! Thanks for the improvements.

PR #1314, opened 2025-04-15
Suggestion to improve resampler in openghg/data_processing/_resampling.py

alexdanjou:
Change `drop_na` in _resample to be able to accept kwargs for xarray Dataset.dropna


alexdanjou:
The idea is to give more flexibility to the dropna part of the resampling. Right now we want to do it for the surface obs, but we have to do it outside of this function due to its rigidity. This would allow to do it inside while keeping the flexibility for the other kind of data

brendan-m-murphy (commented):
This looks good, I think you just need to run the formatting.

Also, I guess the behaviour has changed slightly since we are not dropping times when both repeatability and variability are NaNs. (This probably doesn't matter for inversions because there is separate code related to that in `get_mf_error` (or whatever that function is called).)

PR #1324, opened 2025-04-23
Updates to time-resolved footprint calculations

brendan-m-murphy:

Changes have been made to make the modelled obs. calculation with "time resolved" footprints work with the older footprints (with a `fp_HiTRes` variable) and the newer PARIS style footprints (with `fp_time_resolved` and `fp_residual` data variables).

A number of simplifications have been made to the "fp x flux" calculation. The biggest change are the new helper functions `_make_low_freq_flux` and `_make_high_freq_flux`, which do as they say.

The function `fp_x_flux_time_resolved` in `analyse._modelled_obs` accepts either a DataArray, which is assumed to be the `fp_HiTRes` data variable from the old footprint style, or a Dataset, which is assumed to be the new PARIS style. The old style is converted into the new style, then the calculation proceeds along one path.

Since there is one less `H_back` value in the new footprint style, the calculation of "max H back" is done by a helper function, which adds 1 to the new max value. This is consistent with the old code; maybe some other method is better, in which case we can change the helper function.

Tests related to the time-resolved modelled obs calculation have been moved to a new tests file (`test_modelled_obs.py`).


openghg/analyse/_alignment.py
EricSaboya, line 1: Looks good. Only possible concern is using 'datetime' as a variable in the offset function since  `import datetime` is sometimes used

EricSaboya (commented):
Looks good to me. Happy to approve, unless you want to wait for Rachel to check

EricSaboya (approved):
Nice one

PR #1328, opened 2025-04-25
Iss1308 option to get fpxflux in footprints data merge

### LINKED ISSUE #1308: Add option to return `fpXflux` from `footprints_data_merge`
### What is your issue?

Current, `footprints_data_merge` only returns the modelled obs, but to calculate and apply basis functions in inversions, we need the "fp X flux" dataset that is used to derive the modelled obs.

Allowing the `fpXflux` matrix to be returned by `footprints_data_merge` would avoid duplicating this calculation in inversions (or needed to call `_calc_modelled_obs_integrated` or `_calc_modelled_obs_HiTRes` separately).

To avoid making the `ModelScenario` object too big, returning `fpXflux` could be turned off by default. (Alternatively, the result could be returned "uncomputed", so that the computation would be triggered later.)

brendan-m-murphy:

Added option `calc_fp_x_flux` to `ModelScenario.footprints_data_merge`, which adds the fp x flux matrix (which must be calculated to calculate the modelled obs) to the combined data stored in `ModelScenario.scenario` under the data variable name `fp_x_flux`.

This is mostly a neutral change, but the return types of `calc_modelled_obs` and its helper functions have been changed to `xr.Dataset`, and the time series and/or fp_x_flux are returned as data variables in a single dataset. This means that the result of `calc_modelled_obs` can be merged into the combined data, regardless of the options selected.

- [x] Closes #1308

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- Option to return "fp x flux" from `ModelScenario.footprints_data_merge`. [PR #1328](https://github.com/openghg/openghg/pull/1328)
 - Function to compute baseline sensitivities for NESW. This is used in `calc_modelled_baseline` and will be useful for OpenGHG inversions. [PR #1326](https://github.com/openghg/openghg/pull/1326)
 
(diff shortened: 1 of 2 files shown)


EricSaboya (approved):
Happy with these changes!

PR #1348, opened 2025-05-30
Updating metadata merge functionality in Datasource.add_metadata to use central merge_dict() util function

rt17603:

### DIFF
--- a/openghg/store/base/_datasource.py
+++ b/openghg/store/base/_datasource.py
@@ -409,7 +409,7 @@ class Datasource:
         Returns:
             None
         """
-        from openghg.util import to_lowercase
+        from openghg.util import to_lowercase, merge_dict
 
         try:
             del metadata["object_store"]
@@ -419,7 +419,8 @@ class Datasource:
             logger.warning("object_store should not be added to the metadata, removing.")
 
         lowercased: dict = to_lowercase(metadata, skip_keys=skip_keys)
-        self._metadata.update(lowercased)
+        merged_metadata = merge_dict(self._metadata, lowercased, on_overlap="ignore", on_conflict="right")
+        self._metadata = merged_metadata
 
     def get_dataframe_daterange(self, dataframe: DataFrame) -> tuple[Timestamp, Timestamp]:
         """Returns the daterange for the passed DataFrame
--- a/openghg/util/_metadata_util.py
+++ b/openghg/util/_metadata_util.py
@@ -180,7 +180,7 @@ def merge_dict(
     keys_right: Iterable | None = None,
     remove_null: bool = True,
     null_values: Iterable = null_metadata_values(),
-    on_overlap: Literal["check_value", "error"] = "check_value",
+    on_overlap: Literal["check_value", "ignore", "error"] = "check_value",
     on_conflict: Literal["left", "right", "drop", "error"] = "left",
     relative_tolerance: float = 1e-3,
     lower: bool = True,
@@ -204,8 +204,9 @@ def merge_dict(
         remove_null: Before comparing and merging, remove keys from left and right which have null values.
         null_values: Values which are classed as null.
             See null_metadata_values() function for list of default values.
-        on_overlap: If keys overlap, can choose to check values using or raise an error.
-            Options: ["check_value", "error"]
+        on_overlap: If keys overlap, can choose to check values or raise an error.
+            See check_value_match() function for how these values will be compared and checked.
+            Options: ["check_value", "ignore", "error"]
         on_conflict: If there is a conflict between key values, choose how to resolve this.
             Options: ["left", "right", "drop", "error"]
         relative_tolerance: Tolerance between two numbers when checking values.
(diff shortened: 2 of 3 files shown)


openghg/util/_metadata_util.py
```diff
@@ -204,8 +204,9 @@ def merge_dict(
         remove_null: Before comparing and merging, remove keys from left and right which have null values.
         null_values: Values which are classed as null.
             See null_metadata_values() function for list of default values.
-        on_overlap: If keys overlap, can choose to check values using or raise an error.
-            Options: ["check_value", "error"]
+        on_overlap: If keys overlap, can choose to check values or raise an error.
```
brendan-m-murphy, line 207: Line 198 of the docstring is possible inaccurate?

openghg/util/_metadata_util.py
```diff
@@ -278,6 +279,18 @@ def merge_dict(
                     logger.warning(msg)
                 elif on_conflict == "error":
                     values_not_matching[key] = (value1, value2)
+    elif on_overlap == "ignore":
```
brendan-m-murphy, line 282: This could probably be simplified by making a `check_values` function, and if the option is "ignore", then it should just always return `True`, and if the option is "check_value" it should perform the checks from lines to 250-257. You can just make an "update dict" during this loop, and update "merged_dict" at the end.

openghg/util/_metadata_util.py
```diff
@@ -204,8 +204,9 @@ def merge_dict(
         remove_null: Before comparing and merging, remove keys from left and right which have null values.
         null_values: Values which are classed as null.
             See null_metadata_values() function for list of default values.
-        on_overlap: If keys overlap, can choose to check values using or raise an error.
-            Options: ["check_value", "error"]
+        on_overlap: If keys overlap, can choose to check values or raise an error.
+            See check_value_match() function for how these values will be compared and checked.
+            Options: ["check_value", "ignore", "error"]
```
brendan-m-murphy, line 209: Maybe the docstring for `on_conflict` should say "If there is a conflict between values" rather than "if there is a conflict between key values" (does this mean the values of the keys, or the values the keys point to? I think it is the latter, so maybe omit "keys" or say something more specific like "values at the same key")

openghg/store/base/_datasource.py
```diff
@@ -419,7 +419,8 @@ def add_metadata(self, metadata: dict, skip_keys: list | None = None) -> None:
             logger.warning("object_store should not be added to the metadata, removing.")
 
         lowercased: dict = to_lowercase(metadata, skip_keys=skip_keys)
-        self._metadata.update(lowercased)
+        merged_metadata = merge_dict(self._metadata, lowercased, on_overlap="ignore", on_conflict="right")
```
brendan-m-murphy, line 422: this would probably be clearer if the parameters were called `key_overlap` and `value_conflict`, although I may have insisted on `on_conflict` at some point...

brendan-m-murphy (changes requested):
Looks good. I found the parameter names slightly confusing (although I may have been in favour of `on_conflict` in the past). Also I think the long description in the docstring needs to be updated, since the left value isn't always taken on conflict.

openghg/util/_metadata_util.py
rt17603: Do you mean the last line of the following?
> - if the same keys are present in both dictionaries:
>            - if one of the two values is identified as a value which has not been explicitly set (e.g. "not_set") the
>            other value will be used in preference.
>            - otherwise, the value from left gets preference in the merged dictionary.

I believe this is correct since `on_conflict` is set to "left" by default
 - e.g. [`test_merge_not_set_values_ignore`](https://github.com/openghg/openghg/blob/devel/tests/util/test_metadata_util.py#L201) checks this.

That also seems to be true with the new code as the `on_conflict` input is used to determine whether to merge in "left" or "right" into the overall merged_dict - https://github.com/openghg/openghg/blob/centralise_datasource_meta_merge/openghg/util/_metadata_util.py#L284

openghg/util/_metadata_util.py
rt17603: Ok, I've added the doc-string

openghg/util/_metadata_util.py
rt17603: Ok, I can add a "by default" statement for this.

openghg/store/base/_datasource.py
rt17603: I don't think I was necessarily clear - my point was this starts to feel like this goes outside the scope of the PR as it starts impacting other areas of the code. When I was saying I wasn't keen to make the change I was also saying I'm not sure this update is needed.

brendan-m-murphy: I said it is fine to leave it. I mentioned that refactoring in case you're concern was having to change the variable names in other places.

rt17603: Ok thanks for the information on this - useful tool.

openghg/util/_metadata_util.py
rt17603: Is this necessary for the PR do you think? I feel like it's creeping out of scope (as these are areas of the code which aren't being updated here) but I could see that it could be being used as a way to simplify the clauses.

If so, I think I could do with a bit more clarity on what update is being suggested here - would it be checking all the values in one go in a new function (rather than looping over each value and using the `check_value_match` function individually)?

brendan-m-murphy (approved):
Looks good!

Issue #1252, opened 2025-03-17
Refactor read_file code within children to BaseData to centralise this


rt17603:
We want to ensure the same process happens when new data is added. This can be dependent on the `data_type` and `source_format` (and potentially some additional, type specific data) but we should be able to centralise this process through [BaseStore](https://github.com/openghg/openghg/blob/devel/openghg/store/base/_base.py#L29)

This impacts:
- [ObsSurface](https://github.com/openghg/openghg/blob/devel/openghg/store/_obssurface.py#L108)
- [ObsColumn](https://github.com/openghg/openghg/blob/devel/openghg/store/_obscolumn.py#L26)
- [Footprints](https://github.com/openghg/openghg/blob/devel/openghg/store/_footprints.py#L187)
- [FootprintTimeseries](https://github.com/openghg/openghg/blob/devel/openghg/store/_flux_timeseries.py#L52)
- [Flux](https://github.com/openghg/openghg/blob/devel/openghg/store/_flux.py#L59)
- [EulerianModel](https://github.com/openghg/openghg/blob/devel/openghg/store/_eulerian_model.py#L30)
- [BoundaryConditions](https://github.com/openghg/openghg/blob/devel/openghg/store/_boundary_conditions.py#L65)

rt17603:
PR #1405 has moved the tuple checks (for gcwerks) into standardise_surface rather than read_file and so thsi should now mean the input to read file can more consistently be `str | Path | list[str | Path]`

rt17603:
Comment from @brendan-m-murphy 

> Can we include `validate_data` as a class method so this could just be the default definition in BaseStore? I think it would be a class method...

```
@classmethod
def validate_data(cls, data: Dataset) -> None:
    data_schema = cls.schema()
    data_schema.validate_date(data)
```

I think we may want to keep this as a static method as the workflow we currently have allows users to call `ObsSurface.validate_data` on their own data so this update should only be included if this is still possible.

rt17603:
Linked to PR #1183

PR #1386, opened 2025-07-02
Update filepath types to be consistent in store and standardise

rt17603:

Updating the type for `filepath` to be consistently either `pathType` or `multiPathType` for modules within `store` and `standardise`.

This includes updating some standardise functions which included only `Path` and not the option of a `str` for consistency.

This feeds into #1252 where we need to ensure all data type classes have the same expected inputs and metadata types.

Not needed
- Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature

### DIFF
--- a/openghg/store/_footprints.py
+++ b/openghg/store/_footprints.py
@@ -6,7 +6,7 @@ import numpy as np
 from openghg.store import DataSchema
 from openghg.store.base import BaseStore
 from openghg.store.storage import ChunkingSchema
-from openghg.types import pathType
+from openghg.types import multiPathType
 from openghg.util import check_species_lifetime, check_species_time_resolved, synonyms
 from xarray import Dataset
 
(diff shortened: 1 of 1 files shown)


rt17603:
Should wait for 0.15.0 release to be completed before this is merged.

rt17603:
And I'm actually ok with `parse_` functions etc. being more generous in what they accept - I don't think they should necessarily insist on the input being a `Path`.

openghg/standardise/_standardise.py
```diff
@@ -376,7 +376,7 @@ def standardise_bc(
 
 
 def standardise_footprint(
-    filepath: str | Path | list,
+    filepath: multiPathType,
```
brendan-m-murphy, line 379: I prefer the more explicit option here, but it's not really a big deal.
```suggestion
    filepath: str | Path | list[str | Path],
```

brendan-m-murphy (approved):
Looks good! I would probably avoid multiPathType in standardise_footprint, but it's not a big deal.

openghg/standardise/_standardise.py
rt17603: For what I'm trying to do at the moment I could do with starting to make `Footprint` and `ObsSurface` be consistent in terms of the `filepath` type - however, we can see if this makes more sense as the more verbose type once that's ironed out a bit hopefully.

PR #1393, opened 2025-07-11
Use central open_nc function for relevant data types

### LINKED ISSUE #1169: Allow `check_function_open_nc` to be used for all data types
### What is your issue?

Originally a comment from @brendan-m-murphy on PR #1164 

Ideally we would use the  `check_function_open_nc` function with all data types. One option would be to add an argument like align_lat_lon: bool = True and depending on that flag, either use preprocess or not.

Another option is, add a preprocess: Optional[Callable] = None argument to this function. Then we can pass partial(align_lat_lon, domain=domain) to it for the cases where we have lat/lon.

So we probably want a helper function like def open_netcdf(filepath, data_type) and depending on data_type, we can choose what function to use for preprocess (so if data_type in ["footprints", "flux", "boundary_conditions"] we can pass align_lat_lon).

rt17603:

This PR is to apply the open netcdf functionality, which is currently applied only to footprints, to all suitable data types.

The current `footprint_open_nc_fn` can take a list of filepaths, apply pre-processing (in this case the `align_lat_lon` function) and open this using `open_mf_dataset` as a concatenated dataset. This PR:

 - Renames `footprint_open_nc_fn` --> `open_nc_fn` and added `open_time_nc_fn` as a small wrapper to this
 - Includes this functionality for:
    - `standardise_flux` - including applying  `align_lat_lon` function using `domain` input, check time value is a dimension
    - `standardise_bc` - including applying  `align_lat_lon` function using `domain` input, check time value is a dimension
    - `standardise_eulerian` - just checking time value is a dimension
    - `standardise_column` - just checking time value is a dimension
    - Note: this does not apply this functionality to `standardise_surface` or `standardise_flux_timeseries`.
 - Updates this to use explicit `str | Path | list[str | Path]` type hints rather than `multiPathType`

Two issues had to be addressed as a part of adding this functionality (using open_mfdataset more widely):
 - for some bc files, "time" is not properly registered as a coordinate
    - included extra step in open_nc_fn to check the dims are correctly registered and added new function `open_time_nc_fn` to check "time" specifically - [openghg/util/_file.py](https://github.com/openghg/openghg/pull/1393/files#diff-91bdf808583223cbde1a1da8b6b13cee025a64ac09d0aace64e4ba9be40939be)
 - for two values in joined inputs, pd.infer_freq raises as error as this says it needs 3 time points to infer the frequency - this is actually a more general problem which has perhaps not been picked up before
    - split out steps for inferring the frequency to a separate function and added in a special case for 2 values - [openghg/util/_time.py](https://github.com/openghg/openghg/pull/1393/files#diff-657d5ddc799f8742f34a228090cb9d29ca7d9d6d1dd09dbafe172569222c738f)

- [x] Closes #1169 ; also is a useful step towards #1252 by starting to make the types consistent

Needed?
- Tutorials added

Not needed

### DIFF
--- a/openghg/util/_time.py
+++ b/openghg/util/_time.py
@@ -872,9 +872,6 @@ def infer_frequency(timestamps: DatetimeIndex) -> str | None:
 
     timestamps = timestamps.sort_values()
 
-    if len(timestamps) >= 3:
-        return pd.infer_freq(timestamps)
-
     if len(timestamps) < 2:
         raise ValueError("Unable to infer frequency from <2 data points")
     elif len(timestamps) == 2:
@@ -913,6 +910,8 @@ def infer_frequency(timestamps: DatetimeIndex) -> str | None:
                     break
             else:
                 inferred_period = f"{(total_seconds / day):.1f}D"
+    else:
+        inferred_period = pd.infer_freq(timestamps)
 
     return inferred_period
 


openghg/util/_file.py
```diff
@@ -251,9 +253,51 @@ def get_logfile_path() -> Path:
     return Path("/tmp/openghg.log")
 
 
-def footprint_open_nc_fn(
-    filepath: multiPathType, realign_on_domain: str | None = None, sel_month: bool = False
-) -> tuple[Callable, multiPathType]:
+def _select_time(x: xr.Dataset) -> xr.Dataset:
+    # WARNING : designed for a specific case where a day from another month was present
```
SutarPrasad, line 257: I'd like to suggest converting this to a docstring for consistency and PEP-257 compliance.

SutarPrasad (changes requested):
Everything looks good, just a small tidy up change.

openghg/util/_time.py
```diff
@@ -844,6 +847,75 @@ def relative_time_offset(
...
+        elif total_seconds >= min_year:
+            for year in year_options:
+                year_seconds = year * day
+                rounded_seconds = total_seconds / year_seconds
+                if np.isclose(rounded_seconds, 1):
+                    inferred_period = "YS"
+                    break
+            else:
+                inferred_period = f"{(total_seconds / day):.1f}D"
+    else:
```
brendan-m-murphy, line 913: This could be a guard clause, e.g.
```
if len(timestamps) >= 3:
    return pd.infer_freq(timestamps)
```
after line 876.

brendan-m-murphy (approved):
A small suggestion about adding a "guard clause" to simplify the infer frequency function a bit... totally optional. Looks good overall!

openghg/util/_time.py
rt17603: Thanks for the suggestion - I did update this but they mypy starting complaining about this being `Any` type so I reverted it!

PR #1405, opened 2025-07-21
Moving gcwerks check into standardise_surface and out of ObsSurface.read_file

### LINKED ISSUE #1406: Allow precision_filepath to be passed directly for gcwerks source_format
For the (old) GCWERKS source format pairs of files must be provided (filepath and precision_filepath) for this to be standardised. At the moment this can be specified by passing these files as tuples to `filepath`.

This means the `ObsSurface.read_file` function has to accept this more complex file format (`multiPathType`) which makes `filepath` inconsistent with other data types.

Suggestion to improve this:
1. Allow `precision_filepath` to specified as it's own variable in `standardise_surface`
    - Still need to support old workflow for now, but can include a deprecation warning 
2. Move the logic to check this to `standardise_surface` rather than `ObsSurface.read_file`

This also ties into improvements towards applying changes for #1252 .

rt17603:

This PR moves the logic for specifying `precision_filepath` input needed for GCWERKS into `standardise_surface` from `ObsSurface.read_file`. This allows `precision_filepath` to be specified directly and checks that this has been specified correctly (either using this direct input or the previous `filepath` with tuples method).

- [x] Closes #1406 and step towards #1252 as this makes the input `ObsSurface.read_file` more consistent with other data types

Not needed


openghg/standardise/surface/_gcwerks.py
```diff
@@ -607,3 +610,57 @@ def _get_site_attributes(site: str, inlet: str, instrument: str, gc_params: dict
         raise KeyError(f"Invalid instrument {instrument} passed, valid instruments : {valid_instruments}")
 
     return attributes
+
+
+def check_gcwerks_input(
+    filepath: multiPathType, precision_filepath: str | Path | list[str | Path] | None
```
brendan-m-murphy, line 616: can `precision_filepath` be `multiPathType | None`?

rt17603: [`multiPathType` is actually `str | Path | list | tuple`](https://github.com/openghg/openghg/blob/devel/openghg/types/_types.py#L22) so I've started reserving that when a filepath may contain these (filepath, precision_filepath) pairs (I also have a feeling mypy would complain here if we allowed precision_filepath to be too broad).

rt17603: We can perhaps rename it, if we start doing that universally.

PR #1416, opened 2025-07-29
ensure flask variability is read in if present in AGAGE obs source format

qq23840:
Matt is producing files for flask data in the AGAGE format, and that parser currently doesn't accept `mf_variability` as a variable to read in. However, in some cases we are using proxy variabilities for the flask data (e.g. MCOH, BCOB), and so this should be an option. 

I've added a fix that means that if there is a `mf_variability` variable in the source files, it is standardised.


qq23840:
@SutarPrasad - I've added a test (which passes) and updated the changelog.

Issue #1341, opened 2025-04-28
Include a sanity check when grabbing different data types to be combined


rt17603:
Include an order of magnitude sanity check when grabbing different data types
    - Warning for whether these don't seem to be on the same scale

brendan-m-murphy:
Closing since this is covered by PR #1417 (unless the boundary conditions don't have units... but that will be part of #1338

PR #1417, opened 2025-08-05
Use pint in ModelScenario calculations

### LINKED ISSUE #889: Units for baseline not read in ModelScenario
### What happened?

When using `ModelScenario`, the `model.calc_modelled_baseline` method doesn't account for any different units in the boundary conditions other than the default ppb. Leads to the baseline being off by orders of magnitude

The boundary condition file I used doesn't have any mention of units in the attributes. It's in absolute units (e.g. 1.5e-11)s) but is read in as ppb, but when it's added to `ModelScenario` the units are assigned ppb automatically. It would be good to infer the standard units for a species or be able to specify them as you add the BCs. 

_No response_

### Relevant log output

_No response_

image

### Environment

<details>
Python 3.9.7, currently on agage_new_format branch

</details>

brendan-m-murphy:

The calculations for modelled obs and modelled baselines have been updated to use pint when different types of data are combined. This ensures that results like "footprint x flux" have consistent values and units.

An option has been added to `footprints_data_merge` to set the output units. If this is not specified, then the obs units are used. Any variable starting with `mf` will be converted (if possible), including `mf_mod`. Also `bc_mod` and `fp_x_flux` (and similar variables) are converted to be in the specified output units.

When pint is used to preserve units, it is done by quantifying inside a function and dequantifying before returning the result. This means that users do not need to worry about using the `.pint` accessor for things like `.pint.reindex` etc.

Besides the functions in `_modelled_obs.py` and `_modelled_baseline.py`, the util function for stacking fluxes was changed to use pint. This is a place where we probably could have just assigned the units `mol m-2 s-1` to the result, but now we can handle fluxes with differing units, if necessary. A set of tests was added for `stack_datasets`, which is the helper function that required modification.

Other changes not directly related to consistent units in `footprints_data_merge`:
- unit parsing has been set to case insensitive, since the `H_back` coordinate has units `Hours` instead of `hours`, and it seems like we should just accept both. (The case insensitive parsing doesn't apply to prefixes.)
- I fixed typos on lines 1095 and 1103 of `_scenario.py`, where there are loops `for source in sources: ...` but inside the loop, `sources` was used instead of `source`. This actually works, but instead of computing fp x flux for each sector separately, it stacks the fluxes (and does so once for each source).

- [x] Closes #889, and makes #1341 unnecessary

### DIFF
--- a/doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
+++ b/doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
@@ -255,6 +255,9 @@ Output for individual sources can also be created by specifying the
     modelled_obs_anthro = scenario.calc_modelled_obs(sources="anthro", recalculate=True)
     modelled_obs_anthro.plot()
 
+
+Note: units are automatically aligned for satellite data, as in the "Working with units" section of :ref:`Comparing observations to emissions`.
+
 *Plotting functions to be added for 2D / 3D data*
 
 4. Cleanup
--- a/doc/source/tutorials/local/Analysing_data/Comparing_with_emissions.rst
+++ b/doc/source/tutorials/local/Analysing_data/Comparing_with_emissions.rst
@@ -337,6 +337,7 @@ All of this data (except the baseline sensitivities) can be produced at once usi
 
 Notice that the units of all these data variables are compatible. We will say more about this in the next section.
 
+
 5. Working with units
 ---------------------
 


SutarPrasad (commented):
Looks good for me, will wait for others to provide their review. Just a small note to add link referring to `Working with units` in the https://docs.openghg.org/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.html

brendan-m-murphy:
Okay I added that link.

SutarPrasad (approved):

PR #1466, opened 2025-09-12
Direct standardising of the dataset

### LINKED ISSUE #1186: Add `standardise_data` method to data_type classes to allow dataset to be passed directly.

SutarPrasad:
The following PR is focused on standardising the dataset directly. Only the ObsSurface data is configured to be processed.

**API changes**

- `standardise` to accept dataset and filepath.
- Both are made None to make optional.
- Added logic to raise and ValueError when both filepath and dataset missing.
- `standardise_surface` accepts dataset parameter.

**BaseStore class updates**

- Added function called `_standardise_from_dataset` similar to `_standardise_from_file`.
- This function is called inside `read_dataset`, method created newly to process the datasets.

- A new utility function, build_metadata, has been introduced to centralise and standardise metadata preparation.

  Previously, parts of the metadata were constructed within individual parsers, while direct dataset ingestion paths (e.g. standardise_surface(dataset=...)) could bypass those updates. This created the risk of inconsistent or incomplete metadata across different standardisation routes.
  
  With build_metadata, all metadata handling is now consolidated in one place. The function:
  
  Normalises and validates metadata keys.
  
  Merges user-supplied parameters with dataset attributes.
  
  Ensures parser-specific and derived fields (e.g. site, species, inlet, network) are consistently applied.
  
  Provides a single point of truth for future metadata extensions.
  
  This change reduces duplication, prevents divergence between parsers and dataset workflows, and ensures that all standardisation paths generate consistent metadata.

**ObsSurface class updates**

- Added the method called `prepare_surface_gas_data`.
- Called `dataset_formatter` and `assign_attributes` to be used in `prepare_surface_gas_data`
- `prepare_gas_data` function can be expanded to be used in parser's to increase modularity. Add a task below to create a issue and resolve it under new PR.
- [ ] Use `prepare_surface_gas_data` in surface parsers.

- [ ] Closes #1186 (Replace xxxx with the Github issue number)
- [ ] Tutorials updated/added
- [ ] Wiki updated


brendan-m-murphy (commented):
Overall looks good! I made some suggestions to make it less surface specific (I know this is the first case you covered).

It would be really nice if some of the functions introduced to replicate parser behaviour were actually used in the parsers, that way there wouldn't be separate functions that need to be in-sync.

In general though, I think parsing support should be very light for the `read_dataset` function (e.g. just enough to make data variables have the right names and so on).

rt17603 (commented):
From a quick review - I concur with all of @brendan-m-murphy's points on generalisation (e.g. using `prepare_data` method name) to make sure this allows for easy integration of other data types in future. 

With this change, I would also like to be careful around duplication where possible either (as Brendan suggested) making sure the `parse_openghg` function in particular uses the same underlying functions OR updating the `parse_openghg` function to be able to accept data or filepath (and then follows the same steps).

For scoping purposes we'd also need to think carefully about how this model could map to any next steps for this but we can discuss in our catch up meeting.

rt17603:
For this PR, can I suggest going through the suggestions already made and asking for a second round of review when this is ready?

The only additional comment I have at the moment is what we discussed in the meeting around making sure [`parse_openghg` for surface](https://github.com/openghg/openghg/blob/datase/openghg/standardise/surface/_openghg.py#L80) and this new functionality for [build_metadata](https://github.com/openghg/openghg/blob/datase/openghg/util/_metadata_util.py#L390) are both following the same workflow and are not duplicating code/functionality. Ultimately, if we could leverage something like [merge_dict](https://github.com/openghg/openghg/blob/datase/openghg/util/_metadata_util.py#L175) this could make this more generic but we don't necessarily have to achieve that in this PR.

openghg/store/base/_base.py
```diff
@@ -111,6 +111,150 @@ def to_data(self) -> dict:
...
+        """
+
+        from openghg.store.spec import check_parser
+        from openghg.util import check_if_need_new_version
+
+        #  Verifies if the provided source_format is presently handled by openghg. This step can be removed to allow users to treat adding dataset as a custom dataset.
+        source_format = check_parser(self._data_type, source_format)
+
+        #  Handle legacy overwrite / force flags consistently
+        if overwrite and if_exists == "auto":
```
rt17603, line 223: If these lines of code around checking the flags are starting to be duplicated between `read_dataset` and `read_file` (and we're not able to merge the `read_file` and `read_dataset` functions) then we should split this out into another method rather than duplicating the steps.

This goes for other sections in here if the steps are the same between `read_file` and `read_dataset` and are more than 1 (or 2 at most) lines.

SutarPrasad:
As discussed in the group, we are moving forward with Prototype 2.

Issue #1425, opened 2025-08-20
Add `transform_data` to `BoundaryConditions`


brendan-m-murphy:
### What is your issue?

Currently only the `Flux` data class has the `transform_data` method implemented. We need to have this method for `BoundaryConditions` so that we can transform CAMS bc data.

Some issues with `Flux.transform_data` that we want to avoid with CAMS are:
1. the only flux transform parser (for EDGAR) assumes most of the metadata will be extracted from the filename. We want to be able to transform CAMS without necessarily having to extract metadata from the file names, so we need to use the `**kwargs` in `BoundaryConditions.transform_data` to pass args to the parser.
2. We can only standardise one year of EDGAR data at a time; we should make sure that if the CAMS parser will accept multiple years, that the `transform_data` method can too.

In addition to adding `BoundaryConditions.transform_data`, we need to have a method `transform_bc_data` in `_transform.py`. Here I think we should have more arguments in the signature, e.g. for `domain` and `bc_input` (if appropriate), so that we do not need to extract metadata from the filename.

Also, currently the parser in PR #1420 is in `standardise/boundary_conditions`; it needs to be moved to `transform/boundary_conditions`.

PR #1473, opened 2025-09-17
Add cams parser fix v2

### LINKED ISSUE #1486: Add parser for CAMS data

Create parser for CAMS (N2O and CH4) boundary conditions. This is porting over functionality from the acrg repository - https://github.com/ACRG-Bristol/acrg/blob/develop/acrg/BC/BC_CAMS.py

alexdanjou:

- [ ] Closes #1486 
- [ ] Tutorials updated/added
- [ ] Wiki updated

**TODO:** Please make sure to close PR #1420 once the changes from there are integrated / this PR is completed and closed.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,62 @@
+# Dependabot configuration for OpenGHG
+# 
+# What Dependabot does:
+# - Automatically checks for dependency updates across Python packages and GitHub Actions
+# - Opens pull requests when newer versions of dependencies are available
+# - Helps keep dependencies secure and up-to-date with minimal manual effort
+#
+# How maintainers should interact with Dependabot PRs:
+# 1. Review the PR description and changelog links for the updated dependency
+# 2. Check that all CI tests pass (GitHub Actions will run automatically)
+# 3. If tests pass: merge the PR to keep dependencies current
+# 4. If tests fail: investigate the breaking changes and either:
+#    - Fix any compatibility issues in the codebase, or
+#    - Add version constraints if the update introduces breaking changes
+# 5. Configuration adjustments can be made by editing this file (.github/dependabot.yml)
+#
+# References:
+# - Issue: #1461
+# - Documentation: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
+version: 2
+updates:
+  # Monitor Python dependencies from pip (pyproject.toml and requirements*.txt files)
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "09:00"
+      timezone: "UTC"
+    open-pull-requests-limit: 10
+    labels:
+      - "dependencies"
+      - "python"
+    commit-message:
+      prefix: "deps"
+      prefix-development: "deps-dev"
+      include: "scope"
+    # Group minor and patch updates to reduce PR volume
+    groups:
+      minor-and-patch:
+        patterns:
+          - "*"
+        update-types:
+          - "minor"
+          - "patch"
+
+  # Monitor GitHub Actions dependencies
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+      time: "09:00"
+      timezone: "UTC"
+    open-pull-requests-limit: 5
+    labels:
+      - "dependencies"
+      - "github-actions"
+    commit-message:
+      prefix: "ci"
+      include: "scope"
--- a/.github/workflows/conda.yaml
+++ b/.github/workflows/conda.yaml
@@ -26,7 +26,7 @@ jobs:
         os: ["ubuntu-latest"]
         python-version: ["3.10", "3.11", "3.12"]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v5
       - uses: mamba-org/setup-micromamba@v2
         with:
           environment-file: environment.yaml
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -26,9 +26,9 @@ jobs:
         python-version: ["3.10", "3.11", "3.12"]
     steps:
       - name: Checkout
-        uses: actions/checkout@v3
+        uses: actions/checkout@v5
       - name: Set up Python ${{ matrix.python-version }}
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v6
         with:
           python-version: ${{ matrix.python-version }}
           cache: "pip"
@@ -63,9 +63,9 @@ jobs:
     runs-on: ubuntu-latest
     needs: ["test"]
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v5
       - name: Set up Python
-        uses: actions/setup-python@v4
+        uses: actions/setup-python@v6
         with:
           python-version: "3.12"
           cache: "pip"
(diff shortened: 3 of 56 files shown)


openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+
+from openghg.util import find_domain, open_time_nc_fn, timestamp_now
+from openghg.store import infer_date_range, update_zero_dim
+from openghg.retrieve import get_footprint
+
+logger = logging.getLogger("openghg.standardise.boundary_conditions")
+logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
+
+
+def interp1d_np(data: xr.Dataset, x: xr.DataArray, xi: np.ndarray, **kwargs):
```
brendan-m-murphy, line 18: `data` and `x` should be `np.ndarray` here. The function `xr.apply_ufunc` converts functions that work on `np.ndarray` and makes them work on `xr.Dataset` and `xr.DataArray`, which is what happens in `xr_interp`.

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+) -> xr.DataArray:
+    """Iterpolate along given dim.
+
+    If 'coord' is passed, then an alternate coordinate can be used, but the interpolation happens
+    along 'dim'.
+
+    This is useful if there is a "physical" coordinate, like altitude, which depends on lat and lon.
+    So if data has dims: lat, level, and data.z is a coordinate with dims lat, level, which converts
+    level to a height, dependant on lat, then setting dim='level' and coord='z' will allow interpolation
+    with interp_vals that are on the same scale as 'z'.
```
brendan-m-murphy, line 33: Maybe add a note that this is based on the example here: https://tutorial.xarray.dev/advanced/apply_ufunc/example-interp.html

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+    """
+    dim_coord = data[dim] if coord is None else data[coord]
+    result = xr.apply_ufunc(
```
brendan-m-murphy, line 36: Try this to fix the mypy error:
```suggestion
    result: xr.DataArray = xr.apply_ufunc(
```

If that doesn't work, then at the end do `return cast(xr.DataArray, result)` (you need to import `cast` from `typing` to do this)

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+        filepath: list of filepath that will be standardised
+        cams_version: version of cams product that is used
+        secies: species of interest
+        input_observations: type of input observations (e.g. "surface_satellite_dm") that is used
+    Returns:
+        tuple containing 1. a string aggregating the cams versions used, 2. the species and 3. a string aggregating the input observations used.
+    """
+    new_filepath, cams_version_check, species_check, input_observations_check = [], [], [], []
+
+    if not isinstance(filepath, list):
```
brendan-m-murphy, line 174: There is a function in `openghg.util` called `normalise_to_filepath_list` that will convert `filepath` to a list of paths. This will also get rid of a mypy error.

brendan-m-murphy: Also maybe change the name, like
```
filepath_list = normalise_to_filepath_list(filepath)
```
to avoid the mypy issue on line 179 (mypy isn't sure that `filepath` is a list there...)

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+        filepath = [
+            filepath,
+        ]
+
+    for file in filepath:
```
brendan-m-murphy, line 179: try the suggestions above, but if that doesn't work, do this:
```suggestion
    for file in filepath:  # type: ignore
```

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+        make_climatology: If True climatologies will be created. Not implemented yet.
+        chunks: Chunking schema to use when storing data. It expects a dictionary of dimension name and chunk size,
+                for example {"time": 100}. If None then a chunking schema will be set automatically by OpenGHG.
+                See documentation for guidance on chunking: https://docs.openghg.org/tutorials/local/Adding_data/Adding_ancillary_data.html#chunking.
+                To disable chunking pass in an empty dictionary.
+    Returns:
+        Dict: Dictionary of "species_bc_input_domain" : data, metadata, attributes
+    """
+
+    filepath, cams_version, species, input_observations = _check_and_set_params(
```
brendan-m-murphy, line 313: mypy is being dumb and doesn't realise that filepath is a list of paths after this line, and raises an error on line 360 when you pass `filepath` to `make_metadata`. Using `filepath_list, cams_version, ... = _check_and_set_params(...)` should fix this.

An alternative would be to add the line `assert isinstance(filepath, list[Path])` after line 315, which would tell mypy that the type of `filepath` has been "narrowed" to `list[Path]`.

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,374 @@
...
+        CAMS_input_observations=input_observations,
+    )
+
+    bc_data.attrs.update(add_attrs)
+
+    # create metadata
+    metadata = make_metadata(
+        bc_data,
+        filepath,
+        period,
```
brendan-m-murphy, line 361: At this point, `period` can't be `None`.  If `None` is passed then you can use the `infer_frequency` function from `openghg.util` to try to guess it from the CAMS time coordinate. (Basically it uses `pd.infer_freq`, but adds some logic in case there are only two time values.)

brendan-m-murphy (commented):
I will finish the review later, but for now, here are some suggestions for clearing the mypy errors.

SutarPrasad (changes requested):

openghg/standardise/boundary_conditions/_cams.py
```diff
@@ -0,0 +1,371 @@
...
+        continuous: whether time stamps have to be continuous
+        make_climatology: If True climatologies will be created. Not implemented yet.
+        chunks: Chunking schema to use when storing data. It expects a dictionary of dimension name and chunk size,
+                for example {"time": 100}. If None then a chunking schema will be set automatically by OpenGHG.
+                See documentation for guidance on chunking: https://docs.openghg.org/tutorials/local/Adding_data/Adding_ancillary_data.html#chunking.
+                To disable chunking pass in an empty dictionary.
+    Returns:
+        Dict: Dictionary of "species_bc_input_domain" : data, metadata, attributes
+    """
+    filepath_list = normalise_to_filepath_list(filepath)
```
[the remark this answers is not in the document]
alexdanjou: With that, if I want to do only one file, `open_time_nc_fn(filepath)` will return `xr.open_dataset` and `str` or `Path` (depending on `filepath`). But it will be converted to a list of path by `normalise_to_filepath_list` and therefore `xr_open_fn(filepath_list)` will return an error (as it will try to open a list of path with `xr.open_dataset`).

So I pass mypy checks but I loose one important ability ^^ (and that's why the test fail).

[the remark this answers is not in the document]
brendan-m-murphy: I would just do a type ignore then... we should probably just use `open_mfdataset` all the time... but until then it doesn't seem worth making the code worse to make mypy happy.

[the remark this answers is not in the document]
SutarPrasad: yes, like Brendan said, add '# type: ignore' for the line

rt17603:
Needs to integrate changes from PR #1427 to move this over from standardise to transform before this can be completed, This will then close Issue #1425

brendan-m-murphy (commented):
Looks good @alexdanjou, nice work! I'm not approving because @SutarPrasad will merge his changes into this branch.

PR #1447, opened 2025-09-02
Drop pin for icoscp and update authentication details in the docs

rt17603:
* **Summary of changes** (Bug fix, feature, docs update, ...)

The authentication steps for icoscp have been updated and this PR is to bring the docs in line with this. This also drops the current pinning of `icoscp` to 0.17.0 within the requirements as this no longer works with the latest authentication method.

Updated to also ensure `numpy>=2.0` as `icoscp` is currently pinned to `numpy<2.0` but our current code seems to work ok with this.

*As long as this works ok for others, this supercedes PR #1139 - here we move to putting the onus on the icoscp docs to tell us how to authenticate rather than needing to maintain an internal authentication method in openghg. If this PR is merged, that PR can be closed.*

* **Please check if the PR fulfills these requirements**

- [x] Closes #612 
- [x] [Tests added and passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature
- [x] All code checks passing - `black --line-length 110` run over code, `mypy` and `flake8` not showing any errors
- [x] Tutorials updated/added
- [x] Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature
- [x] Added any new requirements to `requirements.txt`, `environment.yaml` and `pyproject.toml`

Not needed
- Documentation updated/added
- Wiki updated

PR #1474, opened 2025-09-19
Added SPARQL search and direct download for ICOS

brendan-m-murphy:

This adds new features for working with ICOS:
1. Search methods that create Sparql queries, use the queries to search the ICOS Carbon Portal, then return the results as a pandas DataFrame
    - The main query is `data_query`, which creates a Sparql query string to search for site, inlet, species, data level, file name, ICOS "spec", and the ICOS dobj uri. This search can be filtered by passing a (list of) site(s), data level, (list of) species, (list of) inlet(s), plus filtering the ICOS spec label by a substring/regex, and including/excluding NRT data, as well as a less strict mode.
    - There is a query to get attributes for the data downloaded, including units
    - There is a query to list ICOS "specs" and "formats", which can be used with the `data_query` to find the format for a particular dobj.
    - If you print the queries, you can run them on the ICOS Sparql query endpoint: https://meta.icos-cp.eu/sparqlclient/?type=CSV
    - The method `make_query_df` should product a similar result as a pandas DataFrame
2. Methods for directly downloading the ICOS text files for ATC times series and ATC flask time series, as well as directly downloading netCDF time series. 
    - Functions for getting raw data: `get_icos_text_file` and `get_icos_nc_file`, both take a dobj uri string
    - `get_data_attrs` gets info on columns/data vars from the ICOS CP
    - Extra functions for processing text files:
        - `get_icos_text_header` gets lines from the text file starting with `#` and does some extra parsing of the "COMMENT" section
        - `get_attrs_from_header` gets attrs for columns, and `get_full_attrs` combines these with the attrs from `get_data_attrs`
        - `parse_date_columns` probably could be a more general util function, it is used for "ATC Product Time Series" format text files, since they don't have a combined datetime column
        - `dtypes_dict` and `icos_format_to_dtype`: used in conjunction with the attrs to find the data type for columns in the text file 
        - `parse_icos_atc_time_series_text` parses text files with ICOS format http://meta.icos-cp.eu/ontologies/cpmeta/asciiAtcProductTimeSer
        - `parse_icos_atc_flask_time_series_text` parses text files with ICOS format http://meta.icos-cp.eu/ontologies/cpmeta/asciiAtcFlaskTimeSer
     - `get_icos_data` takes a dict or row from the "data query" DataFrame and produces a lightly formated xr.Dataset with attributes. Only works for ATC Product/Flask time series and netCDF time series formats. Data vars are all "snake case" (e.g. "nb_points" instead of "NbPoints"). The attributes present should be fairly consistent for a given format.
3. Methods for accessing the ICOS station details - `openghg.retrieve.icos._stations.py`

TODO:
- [x] docstrings
- [ ] tests:
    - for parsers, we just need to download some data, save it, and check that parsing works like expected (so this will work without ICOS auth) - can add a mock for this if needed
    - for queries, this is more complicated because we need auth
    - for data download, again, we need auth
- [ ] further dataset formatting
- [ ] metadata extraction: I haven't touched this, the old methods should work but going through `icoscp_core.icos.meta` is probably preferred. For instance `dobj = meta.get_dobj_meta(dobj_uri)` produces an object with a `specificInfo` attribute. (To see all attributes you can do `from dataclasses import asdict` then `asdict(dobj).keys()`. The `dobj` is a Python `dataclass`.)
- [x] mypy checks

- [ ] Closes #1447 (Replace xxxx with the Github issue number)
- [ ] Tutorials updated/added
- [ ] Wiki updated


brendan-m-murphy:
> I'm starting just by testing the query function. Is it possible to separate true ICOS NRT files (e.g. https://meta.icos-cp.eu/objects/8mcL3Hw2_o1HeuSu7OBpx_HI) from the enormous quantity of "Copernicus" NRT files (e.g. https://meta.icos-cp.eu/objects/Acm6yCSA1OPTcKTn3Z03kwi2)? I'm not sure what these Copernicus files are - they never used to come up on the search. I don't think we'll ever want them.

Yes this is supposed to happen by default but I accidentally inverted the logic: for `data_query`, set `nrt=True` for now, and I'll fix it so that `nrt=False` (which is the default) omits the NRT files.

SutarPrasad (changes requested):
just 1 formatting and 1 path change, rests looks good.

SutarPrasad (approved):

PR #1475, opened 2025-09-22
Dataset prototype parser framework

### LINKED ISSUE #1186: Add `standardise_data` method to data_type classes to allow dataset to be passed directly.

SutarPrasad:
 This PR contains standardization of dataset using the existing parser framework. The ability is added to **`agage`, `co2_games` and `openghg`** source_formats for surface data_types. 
The changes I had to do were

- [x] Add "dataset" parameter to the `standardise_surface`, `standardise`, `read_file`, `_standardise_from_file` and `parse_openghg`, `parse_agage` and `parse_co_games`.
- [x] The "dataset" parameter and "filepath" are now made optional.
- [x] Unlike the other prototype I am yet to add the check if both are supplied None the relevant error to be raised.
- [x] The _standardise_from_file is called separately for dataset inside the _read_file function, this is done much earlier and avoids the hashes too.
- [x] Created a separate util called get_dataset which has context_manager to properly open and close the datasets from file.
- [x] If we decide to go with this PR we will require to add more tests.
- From scoping perspective if we are to extend this funcitonality to other source_formats.
- [x] This has been applied inside get_dataset. Inorder to process mulitple_concatenated dataset, a util function similar to `open_time_nc_fn`.is required.
- [x] The input API's remain similar to the user with additional "data" parameter.
- [x] Renamed read_file to read_data to standardse_and_store

- [x] Closes #1186 (Replace xxxx with the Github issue number)
- [x] Tutorials updated/added

### DIFF
--- a/openghg/store/base/_base.py
+++ b/openghg/store/base/_base.py
@@ -383,9 +383,6 @@ class BaseStore:
 
             filepaths = list(unseen_hashes.values())
 
-            if not filepaths:
-                return [{}]
-
             # Check if filepaths are all netcdf files
             file_extensions = [fp.suffix for fp in filepaths]
             nc_extensions = [".nc", ".nc4"]


openghg/store/base/_base.py
```diff
@@ -118,8 +118,9 @@ def read_data(
 
     def _standardise_from_file(
```
rt17603, line 119: Would want to rename this function e.g. `_standardise_from_data`

openghg/store/base/_base.py
```diff
@@ -233,26 +237,31 @@ def _standardise_from_file(
...
-            if isinstance(filepath, list) and len(filepath) == 1:
-                filepath = filepath[0]
-
-            if isinstance(filepath, Path):
-                x.update({"file": filepath.name})
-                logger.info(f"Completed processing: {filepath.name}.")
-            elif isinstance(filepath, list):
-                filepath_str = ", ".join([fp.name for fp in filepath])
+            if not filepaths:
+                continue
```
rt17603, line 246: May be better to include `break`? Or include this check outside the loop instead?

openghg/store/base/_base.py
```diff
@@ -233,26 +237,31 @@ def _standardise_from_file(
...
+                x.update({"file": fp.name})
+                logger.info(f"Completed processing: {fp.name}.")
+            else:
+                filepath_str = ", ".join(fp.name for fp in filepaths)
                 x.update({"files": filepath_str})
                 logger.info(f"Completed processing files: {filepath_str}.")
 
         return datasource_uuids
 
     def read_file(
```
rt17603, line 258: Would likely want to rename this as well - `read_data` is a suggestion but if there is already a function named this we may need to rename the other function to something else.

openghg/store/base/_base.py
```diff
@@ -494,7 +523,7 @@ def _validate_datasources(
         self,
         data: list[MetadataAndData],
         fn_input_parameters: dict,
-        filepath: Path | list[Path] | None = None,
+        filepath: str | Path | list[str] | list[Path] | None = None,
```
rt17603, line 526: Do these additional types need to be added back in? With some careful prior arrangement before calling this function it would be good to ensure this is still `Path | list[Path] | None` at this point.

openghg/standardise/_standardise.py
```diff
@@ -52,16 +61,22 @@ def standardise(
         pass
 
     with dclass(bucket=bucket) as dc:
-        result = dc.read_file(filepath=filepath, **kwargs)
-
+        if dataset is not None:
+            logger.info(
+                "The ability to pass dataset directly to the standardise function is currently availble for {data_type = ['surface'], source_format=['agage', 'openghg', 'co2_games']} "
```
rt17603, line 66: I think we probably want to avoid hard-coding this here as this is likely to change and we'd have to make sure this line stayed up-to-date. May be best to remove this logging message and just allow the individual parsers to raise the error (or could add a check on BaseStore for if the parse function beng called accepts `dataset`).

openghg/store/base/_base.py
```diff
@@ -509,16 +538,18 @@ def _validate_datasources(
         """
         validate_params = self.find_data_schema_inputs()
 
+        if filepath is not None:
```
rt17603, line 541: If the `filepath` input was already `Path | list[Path] | None` we wouldn't need this.

openghg/standardise/surface/_agage.py
```diff
@@ -44,14 +45,13 @@ def parse_agage(
...
     """
-    filepath = Path(filepath)
 
     network = clean_string(network)
     instrument = clean_string(instrument)
 
     # get the parameters from the file metadata, as opposed to from the .json file
 
-    with xr.load_dataset(filepath) as ds:
+    with xr.load_dataset(filepath) as ds:  # type: ignore
```
rt17603, line 54: If `filepath` is None would this not produce an error here?

openghg/standardise/surface/_agage.py
```diff
@@ -84,7 +84,7 @@ def parse_agage(
     species = str(filepath).split(sep="_")[-2]
```
rt17603, line 84: Similarly here, if `filepath` is None I believe this would produce an error. Were we able to extract the `species` from the `file_params` details (`file_params = ds.attrs`) instead?

openghg/standardise/surface/_openghg.py
```diff
@@ -69,170 +71,168 @@ def parse_openghg(
...
-
-    # Search attributes sources (in order) and populate metadata
-    for param in attributes_needed:
-        for source in attribute_sources:
-            if param in source:
-                metadata[param] = source[param]
+    with get_dataset(dataset=dataset, filepath=filepath) as data:
+
+        # Extract current attributes from input data
+        attributes = data.attrs
```
rt17603, line 77: Something like [build_surface_metadata](https://github.com/openghg/openghg/blob/461a15a14ec1173f596f5f1b126cc4fb64e1b82f/openghg/util/_metadata_util.py#L399) from the other prototype could still be incorporated into this workflow. May be something we could approach on a separate PR though as this could also make use of other pre-existing functions like [`merge_dict`](https://github.com/openghg/openghg/blob/devel/openghg/util/_metadata_util.py#L175).

tests/standardise/test_standardise.py
```diff
@@ -114,6 +115,58 @@ def test_standardise_obs_openghg():
...
+            site="TAC",
+            network="DECC",
+            inlet=185,
+            instrument="picarro",
+            source_format="openghg",
+            sampling_period="1h",
+            force=True,
+            store="user",
+            update_mismatch="metadata",
+        )
```
rt17603, line 169: Would be good to add quick source_format="agage" and source_format="co2_games" tests as well in the appropriate test files. They can just use the same data, just checks that the `dataset` input can be used successfully for these source_format inputs.

doc/source/tutorials/local/Adding_data/Adding_observation_data.rst
```diff
@@ -321,6 +321,43 @@ created when we run ``openghg --quickstart``.
 The ``store`` argument can be passed to any of the ``standardise`` functions in OpenGHG and is required if you have write access
 to more than one store.
 
+Adding In-memory dataset to the object store
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Up to this point, we have seen how to specify a file path and add data to the object store.
+In many workflows, however, you may already have your data available in memory as an ``xarray.Dataset``.
+The ``standardise_surface`` function also supports this workflow, allowing you to pass a dataset directly without needing to first write it to disk.
```
rt17603, line 329:
```suggestion
For the appropriate `source_format` inputs, the ``standardise_surface`` function also supports this workflow, allowing you to pass a dataset directly without needing to first write it to disk.
```

rt17603 (changes requested):
Looks good for now - a few small suggested changes here including:
- Renaming `read_file` and `_standardise_from_file`.
- Tests for the additional source_formats would also be valuable here
   - I'd like to check `parse_agage` runs with `dataset` input as there's a few places I think this may fall over when `filepath` is not specified.

rt17603:
Also a quick thought -  are we commiting to the input always being a `dataset` (i.e. an xarray Dataset object) or would we want to call this `data` to leave this open to other data inputs later if we decide?

openghg/standardise/surface/_agage.py
brendan-m-murphy: Can you use the `get_dataset` context manager here? If not, maybe extend it to work in more cases.

openghg/util/_file.py
```diff
@@ -405,3 +407,67 @@ def open_time_nc_fn(
...
+        xr_open_fn, filepath = open_time_nc_fn(
+            filepath=filepath,
+            realign_on_domain=realign_on_domain,
+            sel_month=sel_month,
+            check_coords=check_coords,
+        )
+        ds = xr_open_fn(filepath).chunk(chunks)
+        yield ds
+    finally:
+        if ds is not None:
```
brendan-m-murphy, line 443: Maybe write this like:
```
...
ds = xr_open_fn(filepath).chunk(chunks)
try:
    yield ds
finally:
    ds.close()
```
This is what the docs say to do: https://docs.python.org/3.10/library/contextlib.html#contextlib.contextmanager

You don't need to do `ds = None` this way. And if something goes wrong before `ds` is defined, then there isn't anything to close.

SutarPrasad: done

brendan-m-murphy (changes requested):
Still a few things to address.

openghg/util/_file.py
brendan-m-murphy: Yes I understand why we want to close the dataset, but we don't need to do that if an error occurs before it is opened. The issue here is that there are too many statements in the `try` block.

You can rewrite this like:
```
    if filepath is not None:
        if isinstance(filepath, list):
             filepath = [Path(f) for f in filepath]
        elif isinstance(filepath, str):
            filepath = Path(filepath)

    xr_open_fn, filepath = open_time_nc_fn(
            filepath=filepath,
            realign_on_domain=realign_on_domain,
            sel_month=sel_month,
            check_coords=check_coords,
        )
    ds = xr_open_fn(filepath)

    try:
        ds = ds.chunk(chunks)
        yield ds
    finally:
        ds.close()
```

so there are fewer possible errors to catch, and if the dataset is never opened, the error will happen before you try to close the dataset.

tests/standardise/test_standardise.py
SutarPrasad: this was added on another branch, I am quickly adding those changes.

openghg/standardise/surface/_agage.py
rt17603: In principle, all of the `file_params` checking looks like it could be included within the `with` statement. At the moment we're using two different methods to open the file - `xr.load_dataset` and `get_data(dataset=dataset, filepath=filepath)` which seems unecessary. But I also think we could argue this is outside of the scope of minimal changes for this PR.

rt17603 (changes requested):
Looks nearly ready to go - just a few small comments to address.

openghg/store/base/_base.py
```diff
@@ -222,37 +226,45 @@ def _standardise_from_file(
...
+                    logger.info(f"Completed processing: {fp.name}.")
+                else:
+                    filepath_str = ", ".join(fp.name for fp in filepaths)
+                    x.update({"files": filepath_str})
+                    logger.info(f"Completed processing files: {filepath_str}.")
 
         return datasource_uuids
 
-    def read_file(
+    def read_data(
```
brendan-m-murphy: I think `standardise_and_store` is a better name here.

SutarPrasad:
Addressed the recent discussion, and ready for review

brendan-m-murphy (approved):
There should be an else clause to handle the case when both filepath and data are missing

PR #1480, opened 2025-09-24
Added check for negative uncertainties

### LINKED ISSUE #1479: Check for negative uncertainty values before resampling
### What is your issue?

Some ICOS data uses -9.99 as a NaN flag for variability and we missed this while parsing the data. This kind of problem isn't caught anywhere else, so we should add a check to `get_obs_surface` to fill these values with NaNs before resampling (since they won't be negative after resampling and will be hard to remove).

brendan-m-murphy:
They are converted to NaN before resampling. This only happens if you are resampling the data (I could apply it even if the data isn't resampled, but then the data would need to be loaded into memory.)

- [x] Closes #1479
- [x] [Tests added and passed](https://docs.openghg.org/development/python_devel.html#testing) if fixing a bug or adding a new feature (tests passed, but no new test added; the change seems to fix the problem with ICOS data with -9.99 in an uncertainty variable)
- Documentation updated/added
- Tutorials updated/added
- Wiki updated
- Added any new requirements to `requirements.txt` and `recipes/meta.yaml`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Bug causing error due to resampling non-numeric data. [PR #1478](https://github.com/openghg/openghg/pull/1478)
 - Fixed bug in `merge_and_extend_dict` to make sure repeated values in `left` and `right` produce unique values in output and to ensure `left` and `right` are not modified in place. [PR #1477](https://github.com/openghg/openghg/pull/1477)
 
+## Added
+
+- Check for negative uncertainty values (these are converted to NaN). [PR #1480](https://github.com/openghg/openghg/pull/1480)
 
 ## [0.16.0] - 2025-08-29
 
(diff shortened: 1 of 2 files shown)


alexdanjou (approved):

rt17603:
Did we need a CHANGELOG entry for this? Could the PR description be updated here just to give a bit more context and how this solves the problem?

brendan-m-murphy:
Yea I haven't finished this, we said yesterday it needs tests. Although it is a pretty simple change.

rt17603:
> Yea I haven't finished this, we said yesterday it needs tests. Although it is a pretty simple change.

Ah apologies, I thought this was ready for review and potential merge so just wanted to highlight a few details to add. Let me know when you're ready on this one.

brendan-m-murphy:
If we could add data directly, it would be pretty easy to make a test, but since this goes through `get_obs_surface`, at the moment I would have to create a new file with negative uncertainty values.

I could split out this addition into a function and test that, but all I have done is add
`ds = ds.where(ds[dv] >= 0.0)` where `dv` is either `..._repeatability` or `..._variability`. I guess it is slightly complicated because there can be multiple matches for `dv`. But the filtering itself is pretty basic.

@rt17603 up to you if you're happy to merge without new tests. It seems that it worked for @alexdanjou and @hdelongueville for the ICOS data with this problem, and that's the only case we know of so far.

rt17603 (approved):
All sounds fine to me.

PR #1522, opened 2025-11-15
Document build failing reforms

### LINKED ISSUE #1523: Failure for retrieve_all on document building process
### What happened?

At the moment the retrieve_all is not able to find the latest version on the directly dataset added. This is causing the document building process to fail

retrieve_all() should be able to find the latest version for all of the data

```Python

```

### Relevant log output

```Python

```

_No response_

### Environment

<details>

</details>

SutarPrasad:

This PR fixes the documentation build issues for Release 0.17.0. Since the branch was created from the latest devel, it was also used for the 0.17.0 release, but all release-specific changes have now been removed. This PR can be merged into devel as a one-off without affecting ongoing development.

The main fix addresses a bug in retrieve_all(), where the value of version="latest" was being overwritten. This caused retrieval errors when datasets had mixed version availability—for example, CO₂ data where some source_format entries had multiple versions while others only had v1. The loop would incorrectly stay locked on a previous version (e.g., v2), resulting in failures.

Additional cleanup includes:
 - resequencing and adjusting tutorial sections for improved clarity.
 - Additional workflow which tests document building and generates artificat to be downloaded inorder to explore the build documents. The retention of artifacts in github by default is 90 days, but we can reduce it upon discussion. As well as we can discuss on how we want to manage this document build workflow.

- [x] Closes #1523  (Replace xxxx with the Github issue number)
- [x] Tutorials updated/added

### DIFF
--- a/openghg/dataobjects/_searchresults.py
+++ b/openghg/dataobjects/_searchresults.py
@@ -176,8 +176,11 @@ class SearchResults:
                 version_to_use = metadata["latest_version"]
             else:
                 if version not in metadata["versions"]:
-                    raise ValueError(f"Invalid version {version} for UUID {uuid}")
-
+                    possible_versions = ", ".join(metadata["versions"])
+                    raise ValueError(
+                        f"Invalid version {version} for UUID {uuid} possible versions: {possible_versions}"
+                    )
+                version_to_use = version
             results.append(
                 ObsData(
                     uuid=uuid,


openghg/dataobjects/_searchresults.py
```diff
@@ -173,15 +173,15 @@ def _retrieve_by_uuid(self, uuids: Iterable, version: str, sort: bool = True) ->
...
+                version_to_use = metadata["latest_version"]
             else:
                 if version not in metadata["versions"]:
                     raise ValueError(f"Invalid version {version} for UUID {uuid}")
 
             results.append(
                 ObsData(
                     uuid=uuid,
-                    version=version,
+                    version=version_to_use,
```
brendan-m-murphy: What if `version` is `"v1"`? Then `version_to_use` will be unbound. I think you need to set `version_to_use = version` in the else clause, after the guard clause for an invalid version

PR #1535, opened 2025-12-02
Fix CAMS parser

### LINKED ISSUE #1556: Ensure the cams units are captured during the transform workflow
### What happened?

Currently the units for cams data_vars vmr_[n,e,s,w] are stored. They are required to be fetched from the species data_var and stored for these variables

_No response_

```Python

```

### Relevant log output

```Python

```

_No response_

### Environment

<details>

</details>

alexdanjou:
* **Summary of changes** (Bug fix, feature, docs update, ...) *

Used parse_expression in cams to parse the units and store the value.
updated unit_mapping to have 1e-09 and 1e-06 values for ppb and ppm respectively.
This ensures when the scalar value is parsed it looks for these values by updated lambda function in the _unit.py

- [x] Closes #1556 (Replace xxxx with the Github issue number)
- [X] All code checks passing - `black --line-length 110` run over code, `mypy` and `flake8` not showing any errors
- [X] Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature

### DIFF
--- a/openghg/transform/boundary_conditions/_cams.py
+++ b/openghg/transform/boundary_conditions/_cams.py
@@ -32,14 +32,13 @@ def get_cams_data_units(ds: xr.DataArray, species: str) -> str:
     """
     species = species.upper()
     if species not in ds:
-        raise KeyError(f"Species '{species}' not found in dataset.")
+        raise ValueError(f"Species '{species}' not found in dataset.")
     try:
         units_from_data = ds[species].attrs["units"]
         parsed_units = cf_ureg.parse_expression(units_from_data)
         if not parsed_units.dimensionless:
-            units = str(parsed_units.units)
-        else:
-            units = str(parsed_units.magnitude)
+            raise ValueError(f"Invalid units for '{species}'.  Not dimensionless: {units_from_data}")
+        units = str(parsed_units.magnitude)
     except Exception as e:
         raise ValueError(f"Could not parse units: {units_from_data}") from e
     return units


openghg/transform/boundary_conditions/_cams.py
```diff
@@ -337,9 +345,23 @@ def parse_cams(
...
+
+        try:
+            cf_units = cf_ureg.parse_expression(units_from_data)
+        except Exception as e:
+            raise ValueError(f"Could not parse units: {units_from_data}") from e
+
+        if cf_units is None:
+            raise ValueError(f"Parsed units are None for: {units_from_data}")
+
+        if getattr(cf_units, "dimensionless", False):
```
brendan-m-murphy: The line `cf_ureg.preprocessors.append(lambda x: unit_mapping.get(x, x))` already does this when you use `cf_ureg.parse_expression` (or should do something like it... if the casting to `str` is necessary, we should add it to the preprocessor). Anyway, I don't think we need to do this here (especially if the checks above worked).

openghg/transform/boundary_conditions/_cams.py
```diff
@@ -337,9 +345,23 @@ def parse_cams(
 
     with xr_open_fn(filepath).chunk(chunks) as ds:
 
-        units = get_cams_data_units(ds)
+        units_from_data = get_cams_data_units(ds, species=species)
+
+        try:
```
brendan-m-murphy: Maybe put this check into `get_cams_data_units`, same with the check on line 355.

openghg/transform/boundary_conditions/_cams.py
brendan-m-murphy: I realise you might have added this because Alex's code was returning "ppb", but I don't think we need to worry about that. It's good to check that we *can* parse the units, but I don't think we need to convert them here.

openghg/transform/boundary_conditions/_cams.py
```diff
@@ -343,25 +342,16 @@ def parse_cams(
...
         units_from_data = get_cams_data_units(ds, species=species)
 
         try:
-            cf_units = cf_ureg.parse_expression(units_from_data)
+            units = str(cf_ureg.parse_expression(units_from_data).units)
         except Exception as e:
             raise ValueError(f"Could not parse units: {units_from_data}") from e
 
-        if cf_units is None:
-            raise ValueError(f"Parsed units are None for: {units_from_data}")
```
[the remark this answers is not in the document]
SutarPrasad: removed this separate check

openghg/util/_units.py
```diff
@@ -1,20 +1,24 @@
...
 # convert scientific notation to volume ratios
-unit_mapping = {"1e-6": "ppm", "1e-9": "ppb", "1e-12": "ppt", "1e-15": "ppq", "1e-09": "ppb"}
-cf_ureg.preprocessors.append(lambda x: unit_mapping.get(x, x))
-
 # remove spaces from some non-standard units ("per mil", "per meg", etc.)
-cf_ureg.preprocessors.append(lambda x: x.replace("per m", "per_m"))
+unit_mapping = {"1e-6": "ppm", "1e-9": "ppb", "1e-12": "ppt", "1e-15": "ppq", "1e-09": "ppb"}
+cf_ureg = UnitRegistry(
+    preprocessors=[
+        lambda x: next((v for k, v in unit_mapping.items() if k in x), x),
```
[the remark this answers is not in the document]
SutarPrasad: yes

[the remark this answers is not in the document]
SutarPrasad: will revisit this in a separate PR, as this might require updating once I start ensure units are ensured across repo.

openghg/transform/boundary_conditions/_cams.py
```diff
@@ -8,25 +8,33 @@
...
+        species: species of the data
     Returns:
         unit
     """
-    if "CH4" in ds and ds.CH4.units == "1e-9":
-        return "ppb"
-    elif "N2O" in ds and ds.N2O.units == "1e-9 mol mol-1":
-        return "ppb"
+    species = species.upper()
+    units = getattr(ds[species], "units", None)
```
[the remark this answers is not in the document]
SutarPrasad: added the check as well as shifted the parse_expression inside the get_cams as discussed earlier

openghg/util/_units.py
```diff
@@ -1,17 +1,20 @@
...
 logger = logging.getLogger("openghg.util")
 logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
 
 
 # convert scientific notation to volume ratios
-unit_mapping = {"1e-6": "ppm", "1e-9": "ppb", "1e-12": "ppt", "1e-15": "ppq"}
-cf_ureg.preprocessors.append(lambda x: unit_mapping.get(x, x))
+unit_mapping = {"1e-6": "ppm", "1e-9": "ppb", "1e-12": "ppt", "1e-15": "ppq", "1e-09": "ppb", "1e-06": "ppm"}
+
+cf_ureg.preprocessors.append(lambda x: next((v for k, v in unit_mapping.items() if k in x), x))
```
brendan-m-murphy: Is this necessary? It would parse "1e-9 m" to "ppb", which is incorrect. So if we need this to parse dimensionless units, we need another approach. Are you sure that our current units registry doesn't consider "ppb" and "1e-09 mol/mol" equivalent?

openghg/transform/boundary_conditions/_cams.py
```diff
@@ -8,14 +8,43 @@
...
+    Returns:
+        unit
+    """
+    species = species.upper()
+    if species not in ds:
+        raise KeyError(f"Species '{species}' not found in dataset.")
+    try:
+        units_from_data = ds[species].attrs["units"]
+        parsed_units = cf_ureg.parse_expression(units_from_data)
+        if not parsed_units.dimensionless:
```
brendan-m-murphy: I think this should just be an error because a mole fraction must be dimensionless

Issue #1366, opened 2025-06-24
Code+Tutorial: Comparing_with_emissions fails to plot_comparison correctly


SutarPrasad:
### What happened?

scenario.plot_comparison() from the tutorial generated a plot with modelled_CH4 values near zero, showing a large difference from the observed values.

<img width="1114" alt="Image" src="https://github.com/user-attachments/assets/b8a7ab4b-2e3a-4fc7-baa8-c40f9433bf2e" />

 

### What did you expect to happen?

The difference between the modelled and the observed values should be less.
The issue appears to be with the boundary conditions units.

### Minimal Complete Verifiable Example

```Python
Please run the "Comparing_with_emissions" tutorial
```

### Relevant log output

```Python

```

### Environment

<details>

</details>

SutarPrasad:
looks like this issue is now resolved

<img width="1250" height="550" alt="Image" src="https://github.com/user-attachments/assets/509ed282-6c70-4ba0-a1c5-842f47eaec5f" />

Some updates were put in this PR: https://github.com/openghg/openghg/pull/1611

PR #1611, opened 2026-04-21
Plot comaprison for hires

### LINKED ISSUE #1612: Function `plot_comparison` is failing to generate the expected outputs
### What happened?

Upon calling scenario.plot_comaprison() comment from user
"The modelled CO2 mf based on hourly footprint should have the same amount of  points with the obsevation sata, but now only one data points left after filtering based on observation time, so the time precision between observation mf and modelled mf based hourly fp was suspected different. The differences on time precision maybe stem from standardization or calculation of modelled mf in openghg.
In addition, the scenario function doesn't seem to work for OCO-2 high-resolution data. Although no errors are reported, there is only one filtered point in the flux plot, whereas it should be consistent with the observations (more than 5000 points)."
[](url)

Image

to get final plot with all data points

Image

```Python

```

### Relevant log output

```Python

```

_No response_

### Environment

<details>

</details>

SutarPrasad:
The "hires" data was unable to handle the varying timestamps for example 
  Without fix:  xarray aligns by label → almost no matches → NaN everywhere
  With fix:     ffill maps 00:00 → 00:37, 14:00 → 14:23 etc. 

Incase of HiRes data `plot_comaprision` was not looking for "mf_mod_high_res" it is now added.

The calibration_scale is not available so KeyError was being raised it is now handled.

- [x] Closes #1612  (Replace xxxx with the Github issue number)
- [x] Tutorials updated/added
- [x] Wiki updated

### DIFF
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -183,6 +183,9 @@ def _make_high_freq_flux(flux: xr.DataArray, fp: xr.DataArray | xr.Dataset) -> x
     # create rolling windows
     flux_high_freq = _make_hf_flux_rolling_avg_array(flux_high_freq, fp)
 
+    # reindex to align with fp time coordinates (after creating rolling windows to avoid NaN values at start of time series)
+    flux_high_freq = flux_high_freq.reindex({"time": fp.time}, method="ffill")
+    
     flux_high_freq.attrs["units"] = flux.attrs.get("units")
 
     return flux_high_freq
@@ -234,7 +237,7 @@ def fp_x_flux_time_resolved(
 
     # create high frequency flux (resampled to gcd of footprint time and H_back frequencies) with H_back dim
     flux_high_freq = _make_high_freq_flux(flux, fp)
-    flux_high_freq = flux_high_freq.reindex({"time": fp_time_resolved.time}, method="ffill")
+    
     fp_x_flux = (flux_high_freq.pint.quantify() * fp_time_resolved.pint.quantify()).sum(
         "H_back"
     ) + fp_x_flux_residual
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -238,5 +238,5 @@ def fp_x_flux_time_resolved(
     fp_x_flux = (flux_high_freq.pint.quantify() * fp_time_resolved.pint.quantify()).sum(
         "H_back"
     ) + fp_x_flux_residual
-    
+
     return cast(xr.DataArray, fp_x_flux.pint.dequantify())
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -234,9 +234,9 @@ def fp_x_flux_time_resolved(
 
     # create high frequency flux (resampled to gcd of footprint time and H_back frequencies) with H_back dim
     flux_high_freq = _make_high_freq_flux(flux, fp)
-
+    flux_high_freq = flux_high_freq.reindex({"time": fp_time_resolved.time}, method="ffill")
     fp_x_flux = (flux_high_freq.pint.quantify() * fp_time_resolved.pint.quantify()).sum(
         "H_back"
     ) + fp_x_flux_residual
-
+    
     return cast(xr.DataArray, fp_x_flux.pint.dequantify())
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -184,7 +184,16 @@ def _make_high_freq_flux(flux: xr.DataArray, fp: xr.DataArray | xr.Dataset) -> x
     flux_high_freq = _make_hf_flux_rolling_avg_array(flux_high_freq, fp)
 
     # reindex to align with fp time coordinates (after creating rolling windows to avoid NaN values at start of time series)
-    flux_high_freq = flux_high_freq.reindex({"time": fp.time}, method=None)
+    fp_index = pd.DatetimeIndex(fp.time.values)
+    flux_index = pd.DatetimeIndex(flux_high_freq.time.values)
+    need_second_reindex = (flux_index.get_indexer(fp_index) < 0).any()
+
+    if need_second_reindex:
+        flux_high_freq = flux_high_freq.reindex(
+            {"time": fp.time},
+            method="ffill",
+            tolerance=pd.Timedelta(hours=fp_highest_res_hours),
+        )
 
     flux_high_freq.attrs["units"] = flux.attrs.get("units")
 
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -184,8 +184,8 @@ def _make_high_freq_flux(flux: xr.DataArray, fp: xr.DataArray | xr.Dataset) -> x
     flux_high_freq = _make_hf_flux_rolling_avg_array(flux_high_freq, fp)
 
     # reindex to align with fp time coordinates (after creating rolling windows to avoid NaN values at start of time series)
-    flux_high_freq = flux_high_freq.reindex({"time": fp.time}, method="ffill")
-    
+    flux_high_freq = flux_high_freq.reindex({"time": fp.time}, method=None)
+
     flux_high_freq.attrs["units"] = flux.attrs.get("units")
 
     return flux_high_freq
@@ -237,7 +237,7 @@ def fp_x_flux_time_resolved(
 
     # create high frequency flux (resampled to gcd of footprint time and H_back frequencies) with H_back dim
     flux_high_freq = _make_high_freq_flux(flux, fp)
-    
+
     fp_x_flux = (flux_high_freq.pint.quantify() * fp_time_resolved.pint.quantify()).sum(
         "H_back"
     ) + fp_x_flux_residual
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -156,7 +156,12 @@ def _make_hf_flux_rolling_avg_array(
     return flux_hf_rolling
 
 
-def _make_high_freq_flux(flux: xr.DataArray, fp: xr.DataArray | xr.Dataset) -> xr.DataArray:
+def _make_high_freq_flux(
+    flux: xr.DataArray,
+    fp: xr.DataArray | xr.Dataset,
+    *,
+    align_to_fp_time: bool = True,
+) -> xr.DataArray:
     fp_highest_res_hours = _fp_time_and_h_back_freq_gcd(fp)
     start, end = _padded_flux_slice_start_and_end(fp)
     offset = time_of_day_offset(start)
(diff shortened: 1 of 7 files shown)


openghg/plotting/_timeseries.py
```diff
@@ -156,6 +156,11 @@ def _plot_single_timeseries(
 
     species_string = _latex2html(species_info[synonyms(species, lower=False)]["print_string"])
 
+    if existing_calibration_scale is None:
+        logger.warning("'calibration_scale' not found in metadata.")
+        existing_calibration_scale = ""
+        metadata["calibration_scale"] = existing_calibration_scale
+
```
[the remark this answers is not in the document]
SutarPrasad: @copilot apply changes based on [this feedback](https://github.com/openghg/openghg/pull/1611#discussion_r3117265346)

openghg/analyse/_modelled_obs.py
```diff
@@ -234,7 +234,7 @@ def fp_x_flux_time_resolved(
 
     # create high frequency flux (resampled to gcd of footprint time and H_back frequencies) with H_back dim
     flux_high_freq = _make_high_freq_flux(flux, fp)
-
+    flux_high_freq = flux_high_freq.reindex({"time": fp_time_resolved.time}, method="ffill")
     fp_x_flux = (flux_high_freq.pint.quantify() * fp_time_resolved.pint.quantify()).sum(
```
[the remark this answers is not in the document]
SutarPrasad: @copilot apply changes based on [this feedback](https://github.com/openghg/openghg/pull/1611#discussion_r3117265400) look for already existing tests use the same data to test this. we also have oco2 data file

openghg/analyse/_scenario.py
```diff
@@ -1511,7 +1511,10 @@ def plot_comparison(
             sources=sources, resample_to=resample_to, platform=platform, cache=cache, recalculate=recalculate
         )
         x_data = modelled_obs["time"]
-        y_data = modelled_obs["mf_mod"]
+        if "mf_mod_high_res" in modelled_obs:
+            y_data = modelled_obs["mf_mod_high_res"]
+        else:
+            y_data = modelled_obs["mf_mod"]
```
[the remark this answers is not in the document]
SutarPrasad: @copilot apply changes based on [this feedback](https://github.com/openghg/openghg/pull/1611#discussion_r3117265419)

openghg/analyse/_modelled_obs.py
```diff
@@ -234,7 +234,7 @@ def fp_x_flux_time_resolved(
 
     # create high frequency flux (resampled to gcd of footprint time and H_back frequencies) with H_back dim
     flux_high_freq = _make_high_freq_flux(flux, fp)
-
+    flux_high_freq = flux_high_freq.reindex({"time": fp_time_resolved.time}, method="ffill")
```
brendan-m-murphy: The high frequency flux is reindexed with forward fill on line 181. The times it is filled to there should include any times the in the footprint time coordinate.

My guess is that the footprint is not aligned to `full_dates` index in `_make_high_freq_flux`. The intention is for `full_dates` to contain all time points in both the flux and fp time coordinates.

brendan-m-murphy (commented):
Can I have more context for this PR? Is there a related issue? Where did this issue arise? What data was used?

This sounds like a bug, so it should have an issue with an example that reproduces.

I was planning to re-write some of this code to be faster. It's possible the alignment scheme isn't that great.

As I noted in my comment, we already reindex the high freq flux before the point where you added a reindex. Reindexing the same data twice is not ideal, so I'd like to have a better idea of what is going wrong.

openghg/analyse/_modelled_obs.py
SutarPrasad: Thank you for pointing this out. In `_make_high_freq_flux`, the original reindexing step assumes regularly spaced time points. The additional reindexing introduced here addresses cases with irregular time intervals. I have removed "ffill" and kept None to avoid missing value masking, bias incase of large gaps etc.

To improve clarity and maintain consistency, I have moved the reindexing logic inside the function so that all related handling is centralised. The initial reindexing remains necessary for generating the `_make_hf_flux_rolling_avg_array`, while the subsequent reindexing ensures proper handling of irregular time points.

Please refer to this issue https://github.com/openghg/openghg/issues/1612

brendan-m-murphy: > Thank you for pointing this out. In `_make_high_freq_flux`, the original reindexing step assumes regularly spaced time points. The additional reindexing introduced here addresses cases with irregular time intervals. I have removed "ffill" and kept None to avoid missing value masking, bias incase of large gaps etc.
> 
> To improve clarity and maintain consistency, I have moved the reindexing logic inside the function so that all related handling is centralised. The initial reindexing remains necessary for generating the `_make_hf_flux_rolling_avg_array`, while the subsequent reindexing ensures proper handling of irregular time points.
> 
> Please refer to this issue #1612

I think maybe you want to do reindex with `method="ffill"` but `limit=1` (I think that is an option) and maybe add a tolerance.

Actually... the tolerance thing sounds familiar. Is this problem satellite specific? I thought we had to make some choices about reindexing/alignment for satellite last year, and tolerance was discussed.

OTOH, it seems like there is a general problem, and your fix solves it. But this probably isn't going to be an issue for in-situ data/footprints unless something is wrong with the footprints.

openghg/analyse/_modelled_obs.py
```diff
@@ -183,6 +183,9 @@ def _make_high_freq_flux(flux: xr.DataArray, fp: xr.DataArray | xr.Dataset) -> x
     # create rolling windows
     flux_high_freq = _make_hf_flux_rolling_avg_array(flux_high_freq, fp)
 
+    # reindex to align with fp time coordinates (after creating rolling windows to avoid NaN values at start of time series)
+    flux_high_freq = flux_high_freq.reindex({"time": fp.time}, method=None)
```
brendan-m-murphy: Maybe instead of `method=None` we want to limit the number of matches, either using `limit=1` or a tolerance, or both. The `limit` option is available on the pandas `reindex` method: https://pandas.pydata.org/pandas-docs/version/2.3/reference/api/pandas.Index.reindex.html#pandas.Index.reindex

And the limit argument should go through `indexer_kwargs` in xarray's `reindex`: https://docs.xarray.dev/en/stable/generated/xarray.DataArray.reindex.html but I'm not certain about that.

brendan-m-murphy (changes requested):
Can you look into the tolerance and limit options?

Also, can you remember/look up conversations from last year where you were adding satellite support to `ModelScenario`? I thought we discussed tolerance for alignment back then. Maybe that is having an effect here. I think this is probably only going to affect satellite data.

If this only affects satellite data, we should at least add a comment explaining this, because I want to make some changes to speed this code up, and it is critical enough that I would consider separate paths for satellite if we need to do extra alignment.

openghg/analyse/_modelled_obs.py
SutarPrasad: Based on this it looks to be issue for other types too:https://github.com/openghg/openghg/issues/1366

I am uncertain about the tradeoff with tolerance and limit.

tests/analyse/test_modelled_obs.py
brendan-m-murphy, line 64: This should say "Hourly" not "Daily"

tests/analyse/test_modelled_obs.py
```diff
@@ -331,6 +335,27 @@ def test_modelled_obs_co2_consistency(model_scenario_co2_dummy, model_scenario_p
...
+    fp = footprint_paris_co2_dummy.data.copy(deep=True)
+    fp = fp.isel(time=slice(0, 2), lat=slice(0, 1), lon=slice(0, 1), H_back=slice(0, 2))
+    fp["fp_time_resolved"][:] = 1.0
+    fp["fp_residual"][:] = 0.0
+    fp = fp.assign_coords(time=pd.to_datetime(["2012-01-01 00:37:00", "2012-01-01 14:23:00"]))
+
+    flux = flux_co2_dummy.data.copy(deep=True).isel(lat=slice(0, 1), lon=slice(0, 1))
+    flux["flux"][:] = 2.0
+
+    result = fp_x_flux_time_resolved(fp, flux)
```
brendan-m-murphy, line 349: The full date range (used to make the high freq flux) is going to include the time "2012-01-01 00:37:00" (I think), so the flux is going to be forward filled by the first reindex to include that time. It will not be forward filled to include "2012-01-01 14:23:00" because that is not off of the start date by some number of hours.

That is why the result has 2.0. I'm not completely sure why it has 0.0, but maybe it dropped NaNs and computed an empty sum.

If the issue is that a small number of footprint times are not regular, this won't be a huge problem, but if none of the footprint times are regularly spaced, then reindexing with method=None will still be dropping a lot of data.

I think reindexing with a limit of 1, a method of ffill or nearest, and a tolerance that would at least prevent multiple flux times from mapping to the same fp time (e.g. slightly less than half of the frequency used for the "full dates" datetime index).

I think the right way to think about this is:

1) we need the high freq flux to have a frequency matching the H_back frequency so we can construct the H_back dim for the high freq flux (there are other ways to do the calculation, but they all rely on the flux going backwards in time in step with H_back)

2) after we have H_back for the high freq flux, we want to align the times to the footprint, in some sensible way.

I think the way the calculation is done not isn't really ideal, and it definitely assumes that the footprint time is "continuous" i.e. has a regular frequency. If that is not true for satellite data, then we need to do more alignment.


Maybe you or Bowei should just check the plotting example again with this branch, but my guess is that the fp_x_flux will still be missing a lot of data if method=None. (Please check before adding a tolerance! If I'm wrong then that is good to know.)

openghg/analyse/_modelled_obs.py
brendan-m-murphy: Is that the same issue? The first plot in #1366 looks like either the units are wrong or the baseline is missing.

With reindex and method=None, it's kind of assuming that the flux and footprint times are aligned, but there might be gaps.

I'm kind of confused about how the test in `test_modelled_obs.py` is still passing with `method=None`. That would only re-arrange the flux time index to match the fp time index, and would add NaNs for any missing values. 

From what I can tell, the flux and footprint are aligned, the flux is restricted to a subset of the footprint times, and then the footprint times are changed. If the flux times don't match the footprint times before the reindex with method=None, then they won't match after (reindex with method=None can only fix an index that is out of order).

That suggests that the test doesn't actually recreate the original bug. (I added a more detailed comment on the test itself.)

SutarPrasad: This is the method=ffill plot:
ffill


This is the method=None plot:
not much of difference is observed
methodNone

brendan-m-murphy (approved):
I think this PR is a good pragmatic fix. The fill tolerance is a bit loose (line 201 of `_modelled_obs.py`), and the NaN created by `ffill` is probably still a problem right? 

Long term, we might need to think about how this should work for satellite data (or other irregular data). I also think the fill should be tightly bounded. Time-resolved footprints are specifically meant to preserve temporal variation in fluxes, for example the CO2 diurnal cycle from photosynthesis and respiration. We want to be careful that filling is not shifting the flux by too much. The tolerance helps with this, but we might want to test its effect.

I think we probably also need alignment before making the rolling window since `H_back` is hours back from the *footprint's* time points.

Maybe make a follow up issue tracking:
1. the NaN from ffill (this is not the only place this happens)
2. the impact of the tolerance (probably minimal here, but would be good to know)
3. alignment when making the rolling window

I've tested a few alternative algorithms to the rolling window approach we're currently using, so I could look into how they would handle irregular footprint times.

Issue #1591, opened 2026-03-10
Standardising footprint fails when files are added in loop but works with file list


SutarPrasad:
### What happened?

for fp in filepaths:
    standardise_footprint(
        filepath=fp,
        site=site,
        domain=domain,
        model="NAME",
        inlet=inlet,
        species="inert",
        source_format="paris",
        store="shared_store_zarr",
        if_exists="new",
        inner_domain=inner_domain,
        chunks={"time": 600, "lat": 600, "lon": 600},
        tag="paris6km",
        info_metadata={"openghg_versions":"0.18.0"},
    )
After using above loop the last file that gets standardise, the data will be stored from that only.

### What did you expect to happen?

The data from each file should get stored and can be retrieved using different values of  start & end date for the available data
data = get_footprint(site='MHD',                            
                             domain='europe-6km', height='10m',                 
                             start_date='2023-01-01',                           
                             end_date='2024-03-01',                             
                             species='inert',                                   
                             store='shared_store_zarr')

### Minimal Complete Verifiable Example

```Python

```

### Relevant log output

```Python

```

### Environment

<details>

</details>

brendan-m-murphy:
`if_exists="new"` will either overwrite the existing data or create a new version with only the most recently added data. So I think that setting should be checked first. `if_exists="auto"` will work if the files being added do not overlap each other.

I think making the arguments for updating and adding data clearer might help in general...

PR #1614, opened 2026-05-04
Clarify datasource versioning and conflict planning

### LINKED ISSUE #1613: Clarify datasource update/versioning semantics before inlining VersionedZarrStore
## Background

While reviewing #1555, the logic around `if_exists`, `save_current`, derived `new_version`, `force`, and overlap detection turned out to be difficult to reason about. Some of the surprising behaviour in #1555 appears to preserve existing behaviour from `devel` / `Iss881-part-E-expose-update-via-datasource`, rather than being introduced by the refactor.

This issue is intended as a precursor to #1555: clarify and test the existing update/versioning semantics first, so that #1555 can focus on inlining `LocalZarrStore`/`VersionedZarrStore` operations without also having to decide what the public flags should mean.

### 1. `if_exists="combine"` with non-overlapping data can create a latest version containing only the new data

`save_current="auto"` currently derives `new_version=True` whenever `if_exists != "auto"`. This happens before the code knows whether the new data overlaps existing data.

On the existing base/devel implementation, `Datasource.add_timed_data()` checks for overlap and then takes the no-overlap branch before it reaches the `if_exists == "combine"` branch:

```python
if not self._store or not overlapping:
    self._store.add(version=version_str, dataset=data, ...)
    date_keys.append(new_daterange_str)
elif if_exists == "combine":
    self._store.update(version=version_str, dataset=data, ...)
```

So for `if_exists="combine"`, `save_current="auto"`, and no overlap, the code creates a new version via `LocalZarrStore.add()`. `LocalZarrStore.add()` creates the new version without copying the current version first, then inserts only the new data.

This is surprising because the documented meaning of `combine` is effectively "replace and insert new data into the current timeseries". A user would likely expect the latest version to contain current data plus new non-overlapping data, with overlapping values replaced by the new input. Instead, for non-overlap, the latest version can contain only the new chunk.

#1555 preserves this behaviour with equivalent direct `VersionedZarrStore` logic (`copy_current` is only true for `combine` when overlap is detected). That makes #1555 look suspicious, but the underlying behaviour is already present.

### 2. New non-overlapping versions can mutate old `date_keys` metadata

`date_keys` is initialised as a direct reference to the previous version's list:

```python
date_keys = self._data_keys[self._latest_version] if self._data_keys else []
```

When `new_version=True` and the new data does not overlap, the no-overlap branch appends to this list before assigning it to the new version. That means the previous version's `_data_keys` metadata can be updated even though the previous zarr data was not changed.

This is metadata-only, and `date_keys` itself is legacy metadata from an older backend, but it can still make version metadata misleading and complicates reasoning about #1555.

#1555 appears to fix this particular aliasing bug by resetting `date_keys` when `new_version` is true, but this would be a good small fix to land before #1555 if we keep the current behaviour for now.

### 3. `force=True` is coupled to replacement/versioning semantics

The store layer uses `force` to bypass hash checks for previously seen files. However, `standardise_and_store()` also does this:

```python
if force and if_exists == "auto":
    if_exists = "new"

new_version = check_if_need_new_version(if_exists, save_current)
```

So a flag documented as "process this input even if it looks identical / previously seen" also changes the storage update policy from `auto` to `new`, and with `save_current="auto"` it also derives `new_version=True`.

This means `force=True` can create a replacement-style new version containing only the forced input. That coupling is surprising from the public `standardise_*` APIs and makes it hard to explain what `force` actually does.

## Suggested precursor PR

The goal of the precursor PR should be to separate public option parsing from the concrete storage operation selected for the current input.

A suggested approach:

1. Keep the public API flags for now: `if_exists`, `save_current`, `force`, and the internal `new_version` parameter where needed for compatibility.
2. Stop deciding the final versioning/storage action before overlap detection.
3. Add a small internal planning helper that takes:
   - requested `if_exists`
   - requested `save_current`
   - whether this is a forced reprocessing
   - whether there is existing data
   - whether the new data overlaps existing data
4. Return an explicit action object/enum describing what to do, for example:
   - `insert_current`: add non-overlapping data to the current version
   - `copy_then_insert`: create a new version copied from current, then insert non-overlapping data
   - `replace_current`: overwrite the current version with only the new data
   - `new_replace`: create a new version containing only the new data
   - `upsert_current`: update overlapping values and insert non-overlapping values in the current version
   - `copy_then_upsert`: copy current to a new version, then upsert the new data
   - `error_overlap`: raise `DataOverlapError`
5. Make `Datasource.add_timed_data()` execute that plan rather than re-deriving meanings from `if_exists` and `new_version` inline.
6. Keep `force` limited to the file/hash processing decision unless there is an explicit reason to preserve the current replacement behaviour. If backward compatibility requires preserving it, document it as such and add tests that pin the behaviour.
7. Add focused tests for the option matrix, especially:
   - `if_exists="auto"`, no overlap, `save_current="auto"` -> append/insert into current version
   - `if_exists="auto"`, overlap -> error without creating an empty/orphan version
   - `if_exists="combine"`, overlap, `save_current="auto"` -> new version containing current data with overlapping values replaced and non-overlaps inserted
   - `if_exists="combine"`, no overlap, `save_current="auto"` -> decide and test whether this should copy current then insert, or intentionally create a new version containing only new data
   - `if_exists="new"`, overlap/non-overlap -> new version containing only the new data when `save_current="auto"`
   - `save_current="n"` variants update/replace the current version rather than preserving it
   - `force=True` with default `if_exists` -> explicitly test whether it should only bypass hash checks or continue to imply replacement semantics
8. Fix the `date_keys` aliasing issue by copying the list before mutation or by constructing fresh metadata for the target version.

## Why this simplifies #1555

#1555 currently has to inline `LocalZarrStore` behaviour into `Datasource` while also preserving, interpreting, and exposing a confusing set of interactions between public flags and lower-level zarr operations. That makes review difficult: it is hard to tell whether a surprising branch in #1555 is a refactor bug or a faithful reproduction of existing behaviour.

A precursor PR that introduces an explicit update/versioning plan would give #1555 a stable contract to preserve. Then #1555 can mechanically replace calls like `LocalZarrStore.add()`, `LocalZarrStore.update()`, and `LocalZarrStore.overwrite()` with the corresponding `VersionedZarrStore` operations, while tests assert the intended semantics independently of the storage wrapper removal.

This should make #1555 smaller, easier to review, and less likely to accidentally lock in behaviour that only exists because `if_exists`, `save_current`, `force`, and overlap detection are currently entangled.

brendan-m-murphy:
## Summary
- Added an explicit timed-data planning helper so `Datasource.add_timed_data()` decides storage behavior after overlap detection.
- Kept version metadata isolated by copying `date_keys` before mutation.
- Tightened Zarr update handling so empty non-overlap selections do not flow into `upsert` as a bogus update.
- Added tests covering the planning matrix, non-overlap combine behavior, metadata isolation, and no-orphan-version overlap handling.

- [x] Closes #1613 
- Documentation updated/added
- Tutorials updated/added
- Wiki updated
- Added any new package requirements to `pyproject.toml` [dependencies section] and `recipes/meta.yaml`

## Testing
- `pytest tests/store/test_datasource.py`
- `pytest tests/db_integrity/test_data_updates.py -k "combine or new_version or new_no_version or overlapping"`
- `pytest tests/storage/test_store.py`
- `flake8 openghg/ tests/store/test_datasource.py --count --statistics`
- `mypy --python-version 3.12 openghg/`
- `black --check --line-length 110 openghg/objectstore/_legacy_datasource.py openghg/storage/_zarr_store.py tests/store/test_datasource.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Fixed unit parsing ambiguities by keeping the Pint registry case-sensitive for SI abbreviations and adding explicit case-variant aliases for coordinate and time units (e.g. `Degrees_North`, `Hours`). [PR #1599](https://github.com/openghg/openghg/pull/1599)
 - Fixed `convert_to_slice` to use `abs(input)` when computing the relative tolerance range, ensuring that negative inlet values (eg. for sites below sea level) are correctly matched during data retrieval. [PR #1605](https://github.com/openghg/openghg/pull/1605)
 - Fixed "xarray fails to decode time" by using pandas datetime conversion and storing as np.datetime64[ns].[PR #1608](https://github.com/openghg/openghg/pull/1608)
+- Clarified datasource update/versioning behavior for `if_exists`, `save_current`, and overlap handling, including fixes for copied-version metadata and non-overlapping combine updates. [PR #1614](https://github.com/openghg/openghg/pull/1614)
 
 ### Added
 
--- a/openghg/objectstore/_legacy_datasource.py
+++ b/openghg/objectstore/_legacy_datasource.py
@@ -367,7 +367,7 @@ class Datasource(AbstractDatasource[xr.Dataset]):
         elif plan.action == "upsert":
             logger.info("Updating store by combining new data with existing.")
             self._store.update(version=version_str, dataset=data, compressor=compressor, filters=filters)
-            date_keys = [get_representative_daterange_str(self.get_data())]
+            date_keys = [get_representative_daterange_str(self.get_data(version=version_str))]
         else:
             raise DataOverlapError(
                 "Unable to add new data, because it overlaps with current data and `if_exists` is set to 'auto'. "
--- a/openghg/objectstore/_legacy_datasource.py
+++ b/openghg/objectstore/_legacy_datasource.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 from collections import defaultdict
+from dataclasses import dataclass
 from typing import Any, cast, Literal
 from typing_extensions import Self
 from types import TracebackType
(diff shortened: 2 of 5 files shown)


SutarPrasad (approved):
looks good

SutarPrasad:
might be worth verifying if this is also resolved https://github.com/openghg/openghg/issues/1591

brendan-m-murphy:
> might be worth verifying if this is also resolved #1591

It doesn't make the user interface any clearer, so I guess it doesn't totally fix it. OTOH, now it should probably be easier to change the options we give to user, since there is an intermediate layer.

PR #1678, opened 2026-06-29
OCO2 process based on integrated footprints

### LINKED ISSUE #1688: OCO2 process based on integrated footprints
### What is your issue?

When using instantaneous fluxes together with integrated footprints, the modeled CO₂ mole fractions differ substantially from those obtained using hourly footprints with H_back = 24 h. This discrepancy is expected because the current implementation applies the instantaneous flux at the release time to the entire 720 h integrated footprint, effectively assuming a constant emission rate throughout the backward integration period. This assumption is not appropriate for species with strong diurnal emission cycles (e.g., CO₂), because emissions at a single time of day cannot represent emissions over such a long period.

LiBowei9:
* **Summary 
This PR updates the forward model to process flux as monthly means instead of instantaneous values for CO2.
* **Changes
- **Flux Calculation**: Changed instantaneous flux to monthly mean for the CO2 model.
- **Footprints**: Implemented logic based on integrated footprints.
- **Functionality**: Enabled the integrated footprint processing path specifically for CO2 simulations.


SutarPrasad:
Updated the PR in OpenGHG style to handle integrated co2 species pipeline. Earlier co2 species pipeline defaulted to time_resolved option only.

openghg/analyse/_modelled_obs.py
```diff
@@ -9,23 +9,41 @@
...
+        flux_low_freq = (
+            flux_var.resample(time="1MS")
+            .mean()
+            .reindex_like(footprint.fp, method="ffill")
+        )
+        flux = flux_low_freq.to_dataset(name="flux")
+    else:
+        # align separately on time
+        # TODO: if method="nearest" was acceptable, then we could align all coordinates at once with reindex_like
+        flux = flux.reindex_like(footprint, method="ffill")
```
brendan-m-murphy: Both if and else branches do the same `.reindex_like`, so that can be done outside the if/else.

openghg/analyse/_modelled_obs.py
```diff
@@ -9,23 +9,41 @@
...
         DataArray containing footprint times flux.
 
     """
     flux = reindex_on_dims(flux, footprint, ["lat", "lon"])
-
-    # align separately on time
-    # TODO: if method="nearest" was acceptable, then we could align all coordinates at once with reindex_like
-    flux = flux.reindex_like(footprint, method="ffill")
-
+    if use_low_freq_flux:
```
brendan-m-murphy: This code could be put in a helper function and added e.g. in `ModelScenario` before `fp_x_flux_integrated` is called, e.g. in `calc_modelled_obs`. I don't think this code should go here though.

brendan-m-murphy (changes requested):
Move the resampling outside of `fp_x_flux_integrated` into e.g. `ModelScenario.calc_modelled_obs`. The resampling to monthly is independent of the fp_x_flux part.

openghg/analyse/_modelled_obs.py
SutarPrasad: separated out into new function

openghg/analyse/_modelled_obs.py
SutarPrasad: removed branches

SutarPrasad:
separated `make_integrated_low_freq_flux` from `fp_x_flux_integrated`

LiBowei9:
This now matches my intended use case: integrated CO2 footprints use the monthly mean of the hourly prior flux rather than the instantaneous flux at each observation time. Time-resolved CO2 footprints remain on the high-frequency pathway.

openghg/analyse/_scenario.py
```diff
@@ -699,17 +703,21 @@ def convert_units(
...
+        # if output_units is None:
+        #     output_units = self.units or "mol/mol"  # use mol/mol if obs units are not available
         if output_units is None:
-            output_units = self.units or "mol/mol"  # use mol/mol if obs units are not available
+            output_units = self.units or "mol/mol"
 
         to_convert = []
         data_vars = data_vars or ds.data_vars
         for dv in data_vars:
+            unit = ds[dv].attrs.get("units")
```
brendan-m-murphy, line 714: This fallback corrupts unrelated unitless variables selected later by substring. For example, `mf_number_of_observations=2` is assigned `mol/mol` here and becomes `2_000_000` when output units are ppm. Please keep `convert_units` limited to variables that already declare units (its documented contract), or restrict any fallback to known mole-fraction outputs. A regression test should assert that observation counts remain unchanged.

openghg/analyse/_scenario.py
```diff
@@ -1266,8 +1293,15 @@ def _calc_modelled_obs_HiTRes(
 
         if "fp_HiTRes" in self.scenario:
             fp = self.scenario.fp_HiTRes
-        else:
+        elif "fp_time_resolved" in self.scenario.data_vars:
             fp = self.scenario[["fp_time_resolved", "fp_residual"]]
+        else:
+            return self._calc_modelled_obs_integrated(
```
brendan-m-murphy, line 1299: The integrated fallback drops `ts_name` and `fp_x_flux_name`. With integrated CO2 and `split_by_sectors=True`, the sector calculation therefore returns `mf_mod`/`fp_x_flux`, and `modelled_obs.update()` replaces the totals with source-dimensional arrays instead of adding the sectoral variables. I reproduced this locally. Dispatching from the available footprint variables (rather than species alone) would keep the integrated total and sector paths consistent.

openghg/standardise/footprints/_acrg_org.py
```diff
@@ -73,7 +74,10 @@ def parse_acrg_org(
 
     fp_data = xr_open_fn(filepath)
 
-    time_resolved = check_species_time_resolved(species, time_resolved)
+    if time_resolved is None:
```
brendan-m-murphy, line 77: Explicit `time_resolved=False` is recorded in metadata but is not applied to ACRG data containing both `fp` and `fp_HiTRes`. The parser retains `fp_HiTRes` and `H_back`, so `ModelScenario` later chooses the HiTRes calculation despite the integrated request. Please drop the time-resolved variable/orphan dimension for the integrated path and add an end-to-end ACRG regression test.

CHANGELOG.md
```diff
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - Converted negative surface observation uncertainty values to NaN before optional averaging. [PR #1544](https://github.com/openghg/openghg/pull/1544)
+- Ensured explicitly integrated CO2 site and satellite footprints are standardised and modelled through the integrated-footprint pathway, while preserving the time-resolved default for CO2 footprints. [PR #1678](https://github.com/openghg/openghg/pull/1678)
```
brendan-m-murphy, line 31: This entry (and the new tutorial entry below) is under the already released `0.19.0` section. They should be moved to the corresponding `Unreleased` sections.

brendan-m-murphy (commented):
Thanks — my two earlier suggestions have been applied: the monthly resampling is now outside `fp_x_flux_integrated`, and the duplicate reindex branches are gone. On re-review I found three correctness regressions below. I’m preparing a small follow-up PR based on this branch with red/green commits. Two related concerns also need attention: `use_low_freq_flux` is not part of the modelled-observation cache identity, and `ModelScenario` currently cannot select integrated versus time-resolved footprints when both are stored.

brendan-m-murphy:
I opened #1698 as the upstream replacement for this PR. It targets `openghg:devel`, carries forward the commits from #1678, and adds the red-test/green-fix follow-up commits for the review findings. The earlier PR on the contributor fork has been closed.

SutarPrasad:
Due to forking and no write access the changes from this PR are migrated all together into **PR #1698**. Closing this PR

PR #1687, opened 2026-07-22
file hash with datasource uuid, force removes shifting to new

### LINKED ISSUE #1676: Re-standardising deleted data is blocked by stale file hashes, and force=True can collapse timed data to the last file
### What happened?

When a datasource is deleted with DataManager.delete_datasource, re-standardising the same source files does not work as expected. After deleting boundary condition data from an object store, re-running standardise_bc on the same files skips them as “previously standardised” and does not recreate the datasource. The only current workaround is force=True, but for timed data that can produce incorrect results because the latest version ends up containing only the final file processed.

There appear to be two separate problems:

Previously standardised file hashes are stored at the data-class level and are consulted before datasource lookup/update logic runs.
Deleting a datasource does not clear those stored file hashes, so the files are still treated as already seen.
force=True bypasses the file-hash gate, but it also changes the later write behaviour for timed data in a way that can route processing into the “new data only” path, which is not appropriate for rebuilding a deleted multi-file timeseries.

This makes datasource deletion non-reversible from a user workflow perspective. After deleting data, users cannot simply re-import the same files. The obvious workaround, force=True, is unsafe for timed multi-file re-imports because it can silently produce an incomplete latest version.

For boundary condition rebuilds, the least-bad workaround I found is:

```
force=True, if_exists="combine", save_current="n"
```

That works better for rebuilding a timed series, but it is still a workaround for stale file-hash state that probably should not survive datasource deletion.

Suggested fix

One of these should happen:

Deleting a datasource should also remove any corresponding stored file hashes.
Or the file-hash skip logic should be tied to existing datasource state, not only to historical file hashes.
Or there should be an explicit supported “re-import deleted files” path that bypasses stale hashes without changing timed-data version semantics.

Deleting a datasource should leave the store in a state where the same input files can be standardised again without requiring force=True.

If force=True is required as a re-import mechanism, it should not implicitly lead to “latest version contains only the final file” for timed multi-file inputs.

- Standardise a set of timed boundary condition files with standardise_bc.
- Delete the resulting datasource with DataManager.delete_datasource.
- Re-run standardise_bc on the same input files.
- Observe that the files are skipped as already standardised.
- Re-run with force=True.
- Observe that the latest version can end up containing only the final file, rather than the rebuilt combined series.

There appear to be two separate problems:

Previously standardised file hashes are stored at the data-class level and are consulted before datasource lookup/update logic runs.
Deleting a datasource does not clear those stored file hashes, so the files are still treated as already seen.
force=True bypasses the file-hash gate, but it also changes the later write behaviour for timed data in a way that can route processing into the “new data only” path, which is not appropriate for rebuilding a deleted multi-file timeseries.

This makes datasource deletion non-reversible from a user workflow perspective. After deleting data, users cannot simply re-import the same files. The obvious workaround, force=True, is unsafe for timed multi-file re-imports because it can silently produce an incomplete latest version.

For boundary condition rebuilds, the least-bad workaround I found is:

```
force=True, if_exists="combine", save_current="n"
```

That works better for rebuilding a timed series, but it is still a workaround for stale file-hash state that probably should not survive datasource deletion.

Suggested fix

One of these should happen:

- Deleting a datasource should also remove any corresponding stored file hashes.
- Or the file-hash skip logic should be tied to existing datasource state, not only to historical file hashes.
- Or there should be an explicit supported “re-import deleted files” path that bypasses stale hashes without changing timed-data version semantics.

### Environment

<details>

</details>

SutarPrasad:

- Invalidated file hashes when a linked datasource is deleted, allowing re-import without `force=True`.
- Preserved per-datasource hash provenance so shared-file re-imports recreate only missing datasources and leave surviving ones unchanged.
- Added migration handling for stores created before datasource hash ownership was recorded.
- Restored `force=True` to only bypass duplicate-file checks; overlap handling remains controlled by `if_exists`.
- Fixed combined updates retaining complete data, date ranges, and version metadata.
- Prevented empty Zarr inserts from leaving orphaned versions.
- Preserved public positional compatibility for datasource update methods.
- Updated tutorials and changelog with deletion, re-import, combine, and `force` behavior.

- [x] Closes #1676 #1591 (Replace xxxx with the Github issue number)


ben-adam63 (dismissed):
I don't know about the failing tests, but on my end this fixes the bug so I'm happy.

ben-adam63 (dismissed):
Still looks good - not having any problems with deleting and re-standardising.

openghg/objectstore/_legacy_datasource.py
```diff
@@ -312,6 +316,7 @@ def add_data(
                    - raises DataOverlapError if there is an overlap
                 - "new" - creates new version with just new data
                 - "combine" - replace and insert new data into current timeseries
+            force: Force adding overlapping data when if_exists is "auto".
```
brendan-m-murphy: I don't really like `if_exists = "auto"` because it is vague, but if the data is overlapping, then `if_exists == 'auto'` should raise an error.

brendan-m-murphy (commented):
Having to use `force=True` when data is deleted is definitely a bug and should be fixed. But I'm not sure all of the changes here are necessary.

The issue states:

> The obvious workaround, force=True, is unsafe for timed multi-file re-imports because it can silently produce an incomplete latest version.

I'm not sure this is correct. Checks for data overlaps are independent from hash checks, so if there is overlapping data, then `if_exists="auto"` should raise an error.

The new `force` option seems to change this contract.

openghg/objectstore/_legacy_datasource.py
SutarPrasad: I agree, it seems vague, we can decide on removing it and keeping the other two options.

if_exists="auto" + overlap + force=False: raises DataOverlapError
if_exists="auto" + overlap + force=True: currently does not raise; it inserts as a new version
if_exists="combine" + overlap: combines
if_exists="new" + overlap: stores only the new data as the latest version

SutarPrasad: I have updated this combination 
if_exists="auto" + overlap + force=True: currently does not raise; it inserts as a new version
is updated to raise the error.

SutarPrasad:
superseded by https://github.com/openghg/openghg/pull/1689

PR #1689, opened 2026-07-24
Remove file hashing from standardisation

### LINKED ISSUE #1676: Re-standardising deleted data is blocked by stale file hashes, and force=True can collapse timed data to the last file

brendan-m-murphy:

  **Decision:** this PR intentionally removes file hashing and retrieved-data hashing. Hash removal is the chosen design, not an intermediate step toward a more elaborate hash-ownership system. Making hashes reliable across multi-file inputs, multiple datasources, deletion, legacy stores, remote retrieval, overlap handling, and version creation adds more complexity than the duplicate-file quick check is worth. The existing `if_exists` policy is now the authoritative way to handle incoming data that overlaps stored data.

  - Fixes #1676 so deleted datasources can be recreated from the same source files.
  - Implements the removal tracked in #1690 across standardisation, ICOS retrieval, conversion, utility exports, tests, and documentation.
  - Discards legacy `_file_hashes` and `_retrieved_hashes` when old store state is loaded, and never persists those fields again.
  - Routes every input through the normal parser/store path and existing `if_exists` overlap policy instead of using a hash-based skip gate.
  - Retains ignored `force` parameters temporarily for call compatibility. Every ignored `force=True` entry point now emits one caller-attributed `DeprecationWarning` with guidance to use `if_exists`; default/`force=False` calls remain warning-free. Unrelated `force` controls that still perform real work are unchanged.
  - Exposes `if_exists` explicitly for ICOS retrieval.
  - Replaces the classic metastore's internal string hash with a canonical serialized-state comparison so concurrent-write protection remains type-sensitive without hashing.
  - Adds a red/green regression for #1676 using two boundary-condition files: before the implementation, re-standardising after datasource deletion returned `[{}]`; after the implementation, both months are recreated and checked against the source data.
  - Preserves all files during looped `if_exists="combine"` overlap updates, including fully overlapping Zarr updates, and retains the hash-independent date-range fixes from #1687.
  - Reworks tests that used hashes as a proxy for insertion to assert datasource contents, overlap behavior, explicit version creation, and deprecation behavior directly.

- [x] Closes #1676 and #1690
- [x] Tutorials updated/added

*Not applicable*

- Wiki update
- New package requirements

*Validation*

- Ignored-`force` red phase: 10 expected warning-test failures and 20 passing default/`force=False` cases
- Ignored-`force` green phase: 30 passed
- Broader standardise/BaseStore/ObsSurface suite: 111 passed, 6 skipped
- ICOS warning/overlap regression plus repeated-flux and multi-file combine regressions: 3 passed
- Earlier branch validation: store suites 90 passed, 7 skipped, 1 xfailed; retrieval access 27 passed, 1 xfailed, 1 xpassed; standardisation 39 passed; database integrity 16 passed, 1 skipped
- Commit hooks passed: Black, Flake8, MyPy, whitespace, line-ending, and debug-statement checks
- Full project MyPy, production Flake8, Black, `git diff --check`, and Sphinx dummy build passed during branch validation
- Ruff was also attempted but is not installed in the current uv environment


openghg/retrieve/icos/_retrieve.py
```diff
@@ -54,7 +55,10 @@ def retrieve_atmospheric(
                 - "never" - don't update mismatches and raise an AttrMismatchError
                 - "from_source" / "attributes" - update mismatches based on attributes from ICOS Header
                 - "from_definition" / "metadata" - update mismatches based on input metadata
-        force: Force adding of data even if this is identical to data stored (checked based on previously retrieved file hashes).
+        force: Deprecated compatibility argument. This value is ignored.
```
brendan-m-murphy: This function should probably emit a deprecation warning if we are ignoring input parameters.

SutarPrasad (approved):

PR #1693, opened 2026-08-03
added more test and info in tutorial

### LINKED ISSUE #1591: Standardising footprint fails when files are added in loop but works with file list

SutarPrasad:
This is intended behaviour, added more tests and explanation in tutorial.

- [x] Closes #1591 (Replace xxxx with the Github issue number)
- [x] Tutorials updated/added
- [x] Wiki updated

### DIFF
--- a/doc/source/tutorials/local/Adding_data/Updating_existing_data.rst
+++ b/doc/source/tutorials/local/Adding_data/Updating_existing_data.rst
@@ -7,13 +7,18 @@ OpenGHG categorises data based on the supplied necessary keywords and stores the
 When adding data to the object store, two checks will be made against currently stored data:
 
 1. Whether data has the same set of distinct keywords.
-2. Whether the time range for the data being added overlaps with the current time range for that data.
+2. Whether any time-coordinate values in the data being added match values
+   already stored for that data.
 
-If the data exists but the time range does not overlap, this data will be added,
-grouped with the previous data and associated with the same keywords when using
-the default ``if_exists="auto"`` policy.
+By default, this is an exact check for matching time-coordinate values rather
+than a check that the start and end date ranges intersect.
 
-By default, if data exists and the time range *does* overlap with existing data, the data will not be added and this will produce a ``DataOverlapError``.
+If the data exists but there are no matching time-coordinate values, this data
+will be added, grouped with the previous data and associated with the same
+keywords when using the default ``if_exists="auto"`` policy.
+
+By default, if data exists and matching time-coordinate values are found, the
+data will not be added and this will produce a ``DataOverlapError``.
 
 Updating data
 -------------
@@ -22,19 +27,25 @@ To add updated data to the object store, when using the ``standardise_*``
 functions the user can specify what action to perform using the ``if_exists``
 input. This provides the options:
 
-1. ``"auto"`` - add to the current data if there are no overlapping data points;
-   raise ``DataOverlapError`` otherwise (default).
-2. ``"new"`` - create a new latest version containing the newly added data only.
-   Previous versions are retained unless ``save_current="n"`` is also used.
+1. ``"auto"`` - add to the current data if there are no matching
+   time-coordinate values; raise ``DataOverlapError`` otherwise (default).
+2. ``"new"`` - make the latest version contain only the newly added data.
 3. ``"combine"`` - combine the new and previous data and prefer the new data where
-   the time range overlaps.
+   time-coordinate values match.
+
+``if_exists`` determines the contents of the resulting latest version;
+``save_current`` determines whether that result is written to a new version or
+into the current latest version.
 
 These choices are deliberate versioning behaviour, not a special case of how
-files are passed in. Passing a list of filepaths to a ``standardise_*`` function
-and calling the same function repeatedly in a loop both apply this same
-``if_exists`` policy. For a growing time series, use ``if_exists="auto"`` for
-non-overlapping files or ``if_exists="combine"`` when new files should be merged
-with current data.
+files are passed in. When a list of filepaths is successfully concatenated by
+the parser, the ``if_exists`` policy is applied once to the combined dataset.
+Calling the same function repeatedly in a Python loop applies the policy once
+per call. Both approaches use the same policy, but the resulting version history
+can differ depending on how the parser groups the inputs. For a growing time
+series, use ``if_exists="auto"`` for files without matching time-coordinate
+values or ``if_exists="combine"`` when new files should be merged with current
+data.
 
 Managing versions
 -----------------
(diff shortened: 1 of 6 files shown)


doc/source/tutorials/local/Adding_data/Updating_existing_data.rst
```diff
@@ -9,31 +9,47 @@ When adding data to the object store, two checks will be made against currently
...
+To add updated data to the object store, when using the ``standardise_*``
+functions the user can specify what action to perform using the ``if_exists``
+input. This provides the options:
 
-1. "auto" - combine with previous data if no overlapping data points, raise ``DataOverlap`` error otherwise (default).
-2. "new" - store the newly added data (only)
-3. *"combine" - combine the new and previous data and prefer the new data in the case where the time range overlaps. - to be implemented.*
+1. ``"auto"`` - add to the current data if there are no overlapping data points;
+   raise ``DataOverlapError`` otherwise (default).
+2. ``"new"`` - create a new latest version containing the newly added data only.
```
brendan-m-murphy: A new latest version is only created if save_current is "y". So maybe it should say something like: "make the latest version contain only the newly added data. If save_current="n", the existing latest version is overwritten. If save_current="y", a new latest version is created, preserving the existing latest version."

doc/source/tutorials/local/Adding_data/Updating_existing_data.rst
```diff
@@ -9,31 +9,47 @@ When adding data to the object store, two checks will be made against currently
...
+   Previous versions are retained unless ``save_current="n"`` is also used.
+3. ``"combine"`` - combine the new and previous data and prefer the new data where
```
brendan-m-murphy, line 33: This also interacts with with save_current. If save current is "y", then a new version is created with the existing version copied in, then the new data is combined there. If save current is "n", then the data is just combined into the latest version (overwriting any data at overlapping times).

doc/source/tutorials/local/Adding_data/Updating_existing_data.rst
```diff
@@ -9,31 +9,47 @@ When adding data to the object store, two checks will be made against currently
...
+   the time range overlaps.
 
-By default, using the "new" option will also create a new version of the data. In this way, the previous data will be retained (saved) but the new data will become the details which are accessed by default.
+These choices are deliberate versioning behaviour, not a special case of how
+files are passed in. Passing a list of filepaths to a ``standardise_*`` function
+and calling the same function repeatedly in a loop both apply this same
+``if_exists`` policy. For a growing time series, use ``if_exists="auto"`` for
+non-overlapping files or ``if_exists="combine"`` when new files should be merged
+with current data.
```
brendan-m-murphy: Crucially, use `save_current="n"` here to avoid creating a new version for each iteration. If you default to `if_exists="combine"` because the data might overlap, that can create new versions unless you say "do not save the current latest version".

brendan-m-murphy (commented):
I made some corrections on the wording... maybe I will have codex double check because these options are super confusing...

brendan-m-murphy:
Follow-up after checking these comments against `check_if_need_new_version()` and `plan_timed_data_update()`, and running the new footprint test matrix (all 6 cases pass):

- Correction to my first inline comment: a new version is **not** created only when `save_current="y"`. With `if_exists="new"` or `"combine"`, the default `save_current="auto"` also creates a new version. Better wording for `"new"` would be: “Make the latest version contain only the newly added data. With `save_current="auto"` (default) or `"y"`, create a new latest version and preserve the current latest version; with `"n"`, overwrite the current latest version.”
- The second comment is correct for explicit `"y"`/`"n"`, but should say that `save_current="auto"` behaves like `"y"` for `if_exists="combine"`.
- The third comment is correct for repeated calls: `if_exists="combine", save_current="auto"` creates a new version on each update after the first; use `save_current="n"` to combine in place.

The clearest description seems to be: **`if_exists` determines the contents of the resulting latest version; `save_current` determines whether that result is written to a new version or into the current latest version.** The full behavior with existing data is:

| `if_exists` | `save_current="auto"` | `save_current="y"` | `save_current="n"` |
|---|---|---|---|
| `"auto"` | append non-overlapping points in place | copy current data to a new version, then append non-overlapping points | append non-overlapping points in place |
| `"new"` | new version containing only new data | new version containing only new data | replace current latest version with new data |
| `"combine"` | new version containing old + new, preferring new at matching times | same as `"auto"` | update current latest version in place, preferring new at matching times |

For every `"auto"` case, matching time-coordinate values still raise `DataOverlapError`.

A few adjacent documentation problems also turned up:

1. The “Managing versions” bullets are currently inaccurate: under `save_current="auto"`, `"new"` and `"combine"` create a new version even when the data do not overlap; and the generic `"yes"`/`"no"` descriptions ignore the selected `if_exists` behavior. The table above would avoid those ambiguities.
2. “Time range overlaps” is misleading. The implementation checks matching time-coordinate values (exact matches by default), not merely intersecting start/end ranges.
3. Section 4 still uses `save_current=False`. On this PR head, `check_if_need_new_version("new", False)` raises `UnboundLocalError`; this must be `save_current="n"`.
4. `Adding_ancillary_data.rst` still shows the old “file has been uploaded previously … skipping” warning immediately after the new paragraph. File-hash skipping has been removed; repeating the input now follows `if_exists` and normally overlaps.
5. The generic list-vs-loop wording could use a batching qualification: when a filepath list is successfully concatenated, the policy is applied once to the combined dataset; a Python loop applies it once per call. The added PARIS-footprint cases do pass and demonstrate equivalence for those fixtures, but this should not promise identical version histories for every parser/input grouping.

Minor nearby fixes: “may different species” should be “many”, the example says `CF_6` but searches for `cf4`, and “we should not see this only includes” should be “we should now see this includes”.

SutarPrasad:
> Follow-up after checking these comments against `check_if_need_new_version()` and `plan_timed_data_update()`, and running the new footprint test matrix (all 6 cases pass):
> 
> * Correction to my first inline comment: a new version is **not** created only when `save_current="y"`. With `if_exists="new"` or `"combine"`, the default `save_current="auto"` also creates a new version. Better wording for `"new"` would be: “Make the latest version contain only the newly added data. With `save_current="auto"` (default) or `"y"`, create a new latest version and preserve the current latest version; with `"n"`, overwrite the current latest version.”
> * The second comment is correct for explicit `"y"`/`"n"`, but should say that `save_current="auto"` behaves like `"y"` for `if_exists="combine"`.
> * The third comment is correct for repeated calls: `if_exists="combine", save_current="auto"` creates a new version on each update after the first; use `save_current="n"` to combine in place.
> 
> The clearest description seems to be: **`if_exists` determines the contents of the resulting latest version; `save_current` determines whether that result is written to a new version or into the current latest version.** The full behavior with existing data is:
> 
> `if_exists`	`save_current="auto"`	`save_current="y"`	`save_current="n"`
> `"auto"`	append non-overlapping points in place	copy current data to a new version, then append non-overlapping points	append non-overlapping points in place
> `"new"`	new version containing only new data	new version containing only new data	replace current latest version with new data
> `"combine"`	new version containing old + new, preferring new at matching times	same as `"auto"`	update current latest version in place, preferring new at matching times
> For every `"auto"` case, matching time-coordinate values still raise `DataOverlapError`.
> 
> A few adjacent documentation problems also turned up:
> 
> 1. The “Managing versions” bullets are currently inaccurate: under `save_current="auto"`, `"new"` and `"combine"` create a new version even when the data do not overlap; and the generic `"yes"`/`"no"` descriptions ignore the selected `if_exists` behavior. The table above would avoid those ambiguities.
> 2. “Time range overlaps” is misleading. The implementation checks matching time-coordinate values (exact matches by default), not merely intersecting start/end ranges.
> 3. Section 4 still uses `save_current=False`. On this PR head, `check_if_need_new_version("new", False)` raises `UnboundLocalError`; this must be `save_current="n"`.
> 4. `Adding_ancillary_data.rst` still shows the old “file has been uploaded previously … skipping” warning immediately after the new paragraph. File-hash skipping has been removed; repeating the input now follows `if_exists` and normally overlaps.
> 5. The generic list-vs-loop wording could use a batching qualification: when a filepath list is successfully concatenated, the policy is applied once to the combined dataset; a Python loop applies it once per call. The added PARIS-footprint cases do pass and demonstrate equivalence for those fixtures, but this should not promise identical version histories for every parser/input grouping.
> 
> Minor nearby fixes: “may different species” should be “many”, the example says `CF_6` but searches for `cf4`, and “we should not see this only includes” should be “we should now see this includes”.

Added more detailed ValueError's for invalid values supplied for save_current etc. Also changed the wordings in the tutorials.

doc/source/tutorials/local/Adding_data/Updating_existing_data.rst
```diff
@@ -7,13 +7,18 @@ OpenGHG categorises data based on the supplied necessary keywords and stores the
...
+   already stored for that data.
 
-If the data exists but the time range does not overlap, this data will be added,
-grouped with the previous data and associated with the same keywords when using
-the default ``if_exists="auto"`` policy.
+By default, this is an exact check for matching time-coordinate values rather
+than a check that the start and end date ranges intersect.
 
-By default, if data exists and the time range *does* overlap with existing data, the data will not be added and this will produce a ``DataOverlapError``.
+If the data exists but there are no matching time-coordinate values, this data
```
brendan-m-murphy: maybe replace "this data" with something more specific like "the new data"?

brendan-m-murphy (commented):
Just a small change, everything else looks good.

brendan-m-murphy (approved):

PR #1694, opened 2026-08-05
1515 allow data input for all data types

### LINKED ISSUE #1515: Allow `data` input for all data_types
At the moment, `data` can be passed directly for obs_surface but this should be applied (for some selected parsers) for all data_types.

**ADD MORE DETAILS**

SutarPrasad:
Added ability to add in-memory data to all the datatype parsers accepting netcdf input format files.

- [x] Closes #1515  (Replace xxxx with the Github issue number)
- [x] Tutorials updated/added
- [x] Wiki updated


ben-westhenry (dismissed):
Looks fine to me.

openghg/standardise/footprints/_acrg_org.py
```diff
@@ -69,9 +70,17 @@ def parse_acrg_org(
...
-    xr_open_fn, filepath = open_time_nc_fn(filepath, domain, sel_month=True)
+    if domain is None or model is None or inlet is None:
+        raise ValueError("`domain`, `model`, and `inlet` must be specified.")
+    species = species or "inert"
 
-    fp_data = xr_open_fn(filepath)
+    if data is None:
+        if filepath is None:
+            raise ValueError("Please specify either `filepath` or `data`.")
+        xr_open_fn, filepath = open_time_nc_fn(filepath, domain, sel_month=True)
```
brendan-m-murphy, line 90: Could the `data=` path go through the same preprocessing as the filepath path? At present it skips domain coordinate validation/alignment and `sel_month=True` (and the PARIS, flux, and BC direct paths have the same underlying parity problem). I reproduced a Dataset with latitudes offset by 0.1° being accepted and labelled `EUROPE`, whereas the filepath route canonicalizes or rejects it. Please share the preprocessing between both input modes and add file-vs-Dataset parity/invalid-domain tests.

openghg/standardise/footprints/_acrg_org.py
```diff
@@ -69,9 +70,17 @@ def parse_acrg_org(
...
+        fp_data = xr_open_fn(filepath)
+    else:
+        fp_data = data
```
brendan-m-murphy: This aliases the caller-owned Dataset. Later, `fp_data["release_height"] = ...` mutates that original object; I reproduced `release_height` appearing in the caller after parsing. Please use a derived copy or `assign()` and add an input-immutability assertion.

openghg/standardise/eulerian_model/_openghg.py
```diff
@@ -36,15 +38,20 @@ def parse_openghg(
...
     species = clean_string(species)
     start_date = clean_string(start_date)
     end_date = clean_string(end_date)
     setup = clean_string(setup)
 
-    with xr_open_fn(filepath).chunk(chunks if chunks is not None else {}) as em_data:
+    if isinstance(filepath, list) and len(filepath) == 1:
+        filepath = filepath[0]
+
+    with get_data(dataset=data, filepath=filepath) as em_data:
```
brendan-m-murphy: `get_data()` applies coordinate preprocessing only for filepath input. A direct Dataset with scalar `time` therefore remains 0-D and later raises `TypeError: len() of unsized object`, while the same file-backed data is expanded to a size-one time dimension. Please normalize direct Dataset coordinates identically and cover scalar-time parity.

openghg/transform/flux/_edgar.py
```diff
@@ -162,15 +163,119 @@ def assemble_edgar_metadata(
...
+    date: str | None,
+    species: str | None = None,
+    domain: str | None = None,
+    lat_out: ArrayType = None,
+    lon_out: ArrayType = None,
+    source: str | None = None,
+    edgar_version: str | None = None,
+) -> dict:
+    """Transform an already opened EDGAR dataset using the normal EDGAR path."""
+    attrs = data.attrs.copy()
```
brendan-m-murphy: These attrs flow into metastore metadata without the native-scalar conversion used by the filepath branch. Values such as `np.int64`/`np.float32` then fail JSON serialization during persistence. Please share the `.item()` sanitization path and add an end-to-end `transform_flux_data(data=...)` test with NumPy-valued attrs.

openghg/transform/flux/_edgar.py
```diff
@@ -162,15 +163,119 @@ def assemble_edgar_metadata(
...
+        try:
+            date = str(pd.Timestamp(data.time[0].values).year)
+        except (AttributeError, IndexError):
+            raise ValueError("Date must be specified when direct data has no time coordinate.")
+
+    domain = domain or "globaledgar"
+    source = source or attrs.get("source", "anthro")
+    edgar_version = edgar_version or attrs.get("database_version", "direct")
+
+    if "flux" in data:
```
brendan-m-murphy: This treats any variable named `flux` as already converted and even defaults missing units to `mol/m2/s`. Raw mass-flux values can therefore be stored unchanged or mislabeled. Please validate canonical molar units, convert recognized mass-flux units using the species molar mass, or reject missing/incompatible units; a numerical parity test against the `emi_<species>` branch would catch this.

openghg/transform/flux/_edgar.py
```diff
@@ -162,15 +163,119 @@ def assemble_edgar_metadata(
...
+
+    prior_info_dict = {
+        "EDGAR": {
+            "version": f"EDGAR {edgar_version}",
+            "filename": attrs.get("filename", "direct.nc"),
+            "raw_resolution": "0.1 degrees x 0.1 degrees",
+            "reference": attrs.get("source", ""),
+        }
+    }
+    key = "_".join((species_label, source, domain, date))
```
brendan-m-murphy, line 368: `date` is not normalized here: the new tutorial passes `date=2014` as an int, which makes this `join` raise `TypeError`. Also, a supplied year is not checked against an existing time coordinate, so a 2016 Dataset can receive a 2015 key. Please normalize to a validated four-digit string and validate/infer it from `time` when present.

brendan-m-murphy (commented):
Inline findings from the review. The preprocessing-parity and EDGAR-units comments are merge-blocking in my view; the remaining comments cover correctness, persistence, API contracts, and test reliability. When rebasing onto current `devel`, please also preserve PR #1698's `time_resolved: bool | None = None` semantics and explicit integrated-CO₂ path.

openghg/store/_flux.py
```diff
@@ -160,7 +161,12 @@ def transform_data(
 
         # Format input parameters (specific to data_type)
         fn_input_parameters = self.format_inputs(**kwargs)
-        fn_input_parameters["datapath"] = Path(datapath)
+        if data is not None:
```
brendan-m-murphy, line 183: Supplying both `datapath` and `data` silently ignores the path, despite the error text promising exactly one input. Separately, omitted `database` reaches `database.upper()` and raises `AttributeError` after the writable store is opened. Please validate the XOR input contract and required database at the public transform boundary before side effects, with the same defensive check for boundary conditions.

openghg/standardise/_standardise.py
```diff
@@ -253,6 +253,7 @@ def standardise_column(
     chunks: dict | None = None,
     info_metadata: dict | None = None,
     concat_nc_files: bool | None = None,
+    data: xr.Dataset | None = None,
```
brendan-m-murphy, line 293: Please document the new `data` parameter and the exactly-one-of `data`/`filepath` contract in this public API docstring. The same omission occurs across the other standardisation and transform entry points, so generated API docs expose the signature without explaining accepted formats, ownership, or required metadata.

tests/transform/flux/test_edgar.py
```diff
@@ -87,6 +87,24 @@ def test_parse_edgar_raw(folder, version, species, mean_raw_flux):
     assert metadata.items() >= expected_metadata.items()
 
 
+def test_parse_edgar_data():
+    """Test transforming an EDGAR dataset supplied directly."""
+    folder = get_flux_datapath("EDGAR/yearly/v6.0_CH4")
+    filepath = next(folder.glob("*.nc"))
```
brendan-m-murphy: This directory contains both 2015 and 2016 NetCDF files, so `next(glob(...))` is filesystem-order dependent and can select 2016 while the test hard-codes 2015. Please select the 2015 fixture explicitly and assert output time, values, and units so year/unit regressions are not masked.

openghg/transform/flux/_edgar.py
brendan-m-murphy: Note: I think we implicitly do not allow the `emi` type EDGAR data files. A more robust process could be useful here, but this is meant to process very specific files.

openghg/transform/flux/_edgar.py
brendan-m-murphy: This is a general problem; I'm not sure if it is addressed in the filepath route either (I don't think we check JSON serialisability globally; it should probably be a metastore responsibility, or a plugin option for the metastore).

openghg/standardise/eulerian_model/_openghg.py
brendan-m-murphy: While we don't currently use this data type, I think the general principle is correct: data vs. datapath input shouldn't remove all validations, many apply equally once the data is in xr.Dataset format.

brendan-m-murphy (changes requested):
Codex found a bunch of issues. The most important is keeping lat/lon alignment correction (or error for out of tolerance coordinates). In general, it seems like we should try to share more of the processing/validation code between the two paths.

openghg/standardise/_standardise.py
```diff
@@ -756,16 +785,12 @@ def standardise_eulerian(
...
-            - "auto" - checks new and current data for timeseries overlap
-                - adds data if no overlap
-                - raises DataOverlapError if there is an overlap
-            - "new" - just include new data and ignore previous
-            - "combine" - replace and insert new data into current timeseries
-        save_current: Whether to save data in current form and create a new version.
-            - "auto" - this will depend on if_exists input ("auto" -> False), (other -> True)
-            - "y" / "yes" - Save current data exactly as it exists as a separate (previous) version
-            - "n" / "no" - Allow current data to updated / deleted
+        if_exists: What to do if existing data is present: ``"auto"`` checks
```
brendan-m-murphy: This change removed some details that might be useful. OTOH these details are verbose and repeated in other docstrings. Put more detailed documentation for the if_exists and save_current options in a module level docstring for `_standardise.py`.

openghg/store/_boundary_conditions.py
```diff
@@ -142,13 +142,44 @@ def transform_data(
...
+        """Transform raw boundary conditions and assign them to the store.
+
+        The database selects a parser such as
+        :func:`openghg.transform.boundary_conditions.parse_cams`. Exactly one
+        of ``datapath`` and ``data`` must be provided.
+
+        Args:
+            datapath: Path to raw boundary-condition data.
+            database: Name of the boundary-condition transform parser.
+            if_exists: Action to take when matching stored data exists.
```
brendan-m-murphy, line 154: Since the type of these arguments is `str` rather than e.g. `Literal`, there is not enough info here for a user to know how these work. A short list of accepted options in the `Args:` section is sufficient; again, we might want to reference more detailed documentation for these settings (like I suggested for the module level docstring in _standardise.py, although now we are crossing multiple submodules).

openghg/transform/boundary_conditions/_cams.py
```diff
@@ -350,33 +350,42 @@ def parse_cams(
...
+
     Returns:
         Dict: Dictionary of "species_bc_input_domain" : data, metadata, attributes
+
+    Raises:
+        ValueError: If exactly one input is not supplied or the species or
+            units cannot be determined from direct data.
     """
 
+    if (datapath is None) == (data is None):
```
brendan-m-murphy: Is `and` more appropriate than `==` here? Or is there some reason that won't work?

openghg/transform/flux/_edgar.py
```diff
@@ -163,43 +163,166 @@ def assemble_edgar_metadata(
     return metadata
 
 
+def _serialisable_attrs(attrs: dict[Any, Any]) -> dict[Any, Any]:
```
brendan-m-murphy: This function is not edgar specific and would be useful in general (provided you modify the docstring to remove the EDGAR part). It should be in the `util` submodule.

openghg/transform/flux/_edgar.py
```diff
@@ -163,43 +163,166 @@ def assemble_edgar_metadata(
...
-    else:
-        name = "fluxes" if "fluxes" in data else f"emi_{species_label}"
-        if name not in data:
-            raise ValueError(f"Data variable {name} not present.")
-        flux_da = data[name] * 1e3 / molar_mass(species_label)
-        units = "mol/m2/s"
+    source = str(source or attrs.get("source", "anthro"))
+    input_edgar_version = edgar_version or attrs.get("database_version")
+    flux_da = _select_raw_edgar_variable(data, species_label, input_edgar_version)
+    flux_da = flux_da * 1e3 / molar_mass(species_label)
```
brendan-m-murphy, line 306: Add short inline comment saying this is converting kg/m2/s to mol/m2/s

brendan-m-murphy (changes requested):
Documentation fixes needed.

brendan-m-murphy (dismissed):
Looks good!

PR #1696, opened 2026-08-05
1695 agents instructions for codex and claude

### LINKED ISSUE #1695: Agents instructions for CODEX, CLAUDE and gitleaks for security

Project needs instructions for codex and claude.

Token utilisation is increasing problem, thus requires an optimal solution to reduce the token utilisation whilst providing beneficial AI agent usage across OpenGHG repo.
Having AGENTS.md will work across almost all of the frontier models and work as a clear set of guides when interacting with OpenGHG codebase

_No response_

### Additional context

_No response_

SutarPrasad:
Added AGENTS.md and CLAUDE.md. Reduced some texts and added more explicit instructions in some place. Used vscode's inbuit analyse feature to enhance the written .md files which anlyses the .md files and suggests more optimisation across the instructions written in the files.

- [x] Closes #1695  (Replace xxxx with the Github issue number)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,122 @@
+# Agent Instructions for OpenGHG
+
+Scope: the whole repository.
+
+OpenGHG is a Python package for greenhouse gas data processing, standardisation,
+storage, retrieval, analysis, and plotting. Limit each change to the minimum
+files required to satisfy the task. Do not refactor unrelated code in the same
+edit. Keep changes scientific-data safe and consistent with nearby code and
+tests.
+
+## Token-Efficient Workflow
+
+1. Start with `graphify-out/` for orientation before broad source searches.
+  Query it surgically; do not load or paste the full JSON files.
+2. Use `graphify-out/manifest.json` to find candidate files. Top-level keys are
+   repository paths, and values contain generated hashes/metadata.
+3. Use `graphify-out/graph.json` to inspect modules, symbols, imports, and
+   likely impact paths. Useful fields are `nodes`, `links`, `source_file`,
+   `source_location`, `label`, `relation`, `source`, and `target`.
+4. After graphify orientation, read only the source, tests, and config directly
+   relevant to the task. Use `rg` inside likely paths instead of scanning the
+   whole repository by default.
+5. Treat `graphify-out/` as generated context. Do not edit it unless explicitly
+   asked.
+6. This file distills `.github/copilot-instructions.md`; do not re-read that
+   file, `README.md`, or `CHANGELOG.md` during routine orientation unless the
+   task specifically concerns them.
+
+Useful graphify queries:
+
+```bash
+jq -r 'keys[] | select(test("standardise|store|retrieve"))' graphify-out/manifest.json
+jq -r '.nodes[] | select(.source_file | test("^openghg/standardise/")) | [.id, .label, .source_file, .source_location] | @tsv' graphify-out/graph.json
+jq -r '.links[] | select(.source_file == "openghg/store/_obssurface.py") | [.relation, .source, .target, .source_location] | @tsv' graphify-out/graph.json
+```
+
+## Project Map
+
+- `openghg/`: package source.
+- `tests/`: pytest suite mirroring package areas; test data lives under
+  `tests/data/`.
+- `openghg/standardise/`: incoming scientific dataset parsing and
+  standardisation.
+- `openghg/store/`, `openghg/storage/`, `openghg/objectstore/`: stored data,
+  metadata, indexing, local object stores, and zarr handling.
+- `openghg/dataobjects/`: user-facing data object classes.
+- `openghg/retrieve/`, `openghg/analyse/`, `openghg/transform/`,
+  `openghg/util/`: retrieval, analysis, transformation, and shared helpers.
+
+## Coding Rules
+
+- Preserve public APIs, metadata keys, file formats, and xarray/pandas behaviour
+  unless the task explicitly changes a contract.
+- Match existing helper APIs, data conventions, validation style, and test
+  fixtures before introducing new patterns.
+- Prefer typed, explicit helpers over duplicated parsing or validation logic.
+- Keep package and test data paths relative to package/test helper APIs. Avoid
+  hard-coded absolute paths.
+- Use Black formatting with `line-length = 110`.
+
+- Add or update focused tests beside the affected code when behaviour changes.
+- Do not include generated caches, local object stores, virtual environments, or
+  `graphify-out/` changes unless explicitly requested.
+
+## Environment
+
+Prefer `uv` for agent-run setup and commands. Fallback order:
+
+1. Use `uv` (preferred).
+2. If `uv` is unavailable, use an already-active Conda or Micromamba environment.
+3. Do not create new Pixi environments.
+
+```bash
+uv sync --extra dev
+uv run python -c "import openghg"
+```
+
+For optional documentation work:
+
+```bash
+uv sync --all-extras
+uv run python -c "import openghg"
+```
+
+Do not create Pixi environments during automated agent runs; Pixi setup can halt
+or take too long. Use Conda or Micromamba only if `uv` is unavailable and the
+task cannot be validated otherwise.
+
+## Validation
+
+Run, at minimum, `compileall` and the single test file most directly covering
+the changed code. Add `black --check` and `flake8` if formatting or style was
+touched. If a change affects typing, include `mypy` for the affected module.
+
+```bash
+uv run python -m compileall -q openghg
+uv run black --check openghg
+uv run flake8 openghg
+uv run mypy openghg
+uv run pytest tests/path/to/test_file.py
+```
+
+If `uv run` is unavailable but an environment is already active, run the same
+tools directly. If a validation check reports failures unrelated to the
+current change (pre-existing issues), note them in the report but do not fix
+them unless the task explicitly asks for it.
+
+Do not trigger GitHub Actions CI workflows. Run only the local validation
+commands listed in the Validation section.
+
+Special test handling:
+
+- Tests marked `cfchecks` require `--run-cfchecks`.
+- Tests marked `icos` require `--run-icos` and may use the network.
+- Tests marked `xesmf` require optional regridding dependencies.
+- Pytest has a default timeout of 300 seconds in `pyproject.toml`.
+
+## Reporting
+
+When finishing, state the focused files changed and the exact checks run. If a
+relevant check could not run because of dependencies, network access, or compiled
+system libraries, say so explicitly.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -70,6 +70,8 @@ Prefer `uv` for agent-run setup and commands. Fallback order:
 2. If `uv` is unavailable, use an already-active Conda or Micromamba environment.
 3. Do not create new Pixi environments.
 
+Never run `sudo` commands.
+
 ```bash
 uv sync --extra dev
 uv run python -c "import openghg"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added ability to save output of plot_comparison using save_path and ability to pass additional figure modifications.[PR #1672](https://github.com/openghg/openghg/pull/1672)
 - Added more tests and explanation in the tutorial regarding if_exists behaviour.[PR #1693](https://github.com/openghg/openghg/pull/1693)
+- Added instructions for agentic ai efficient and optimal use in AGENTS.md, CLAUDE.md.[PR #1695](https://github.com/openghg/openghg/issues/1695)]
 
 ## [0.19.0] - 2026-06-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added ability to save output of plot_comparison using save_path and ability to pass additional figure modifications.[PR #1672](https://github.com/openghg/openghg/pull/1672)
 - Added more tests and explanation in the tutorial regarding if_exists behaviour.[PR #1693](https://github.com/openghg/openghg/pull/1693)
-- Added instructions for agentic ai efficient and optimal use in AGENTS.md, CLAUDE.md.[PR #1695](https://github.com/openghg/openghg/issues/1695)]
+- Added instructions for agentic ai efficient and optimal use in AGENTS.md, CLAUDE.md.[PR #1695](https://github.com/openghg/openghg/issues/1695)
 
 ## [0.19.0] - 2026-06-11
 
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,100 @@
+# Claude Guidance for OpenGHG
+
+OpenGHG is a Python package for greenhouse gas data processing, standardisation,
+storage, retrieval, analysis, and plotting. Limit each change to the minimum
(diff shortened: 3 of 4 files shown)


AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+
+OpenGHG is a Python package for greenhouse gas data processing, standardisation,
+storage, retrieval, analysis, and plotting. Limit each change to the minimum
+files required to satisfy the task. Do not refactor unrelated code in the same
+edit. Keep changes scientific-data safe and consistent with nearby code and
+tests.
+
+## Token-Efficient Workflow
+
+1. Start with `graphify-out/` for orientation before broad source searches.
```
brendan-m-murphy, line 13: Is there a way to create graphify-out? I usually put each agent in a new environment, so hardcoding a graphify-out location will be fiddly. It seems like using graphify should be a skill, rather than a generic instruction.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+
+- `openghg/`: package source.
+- `tests/`: pytest suite mirroring package areas; test data lives under
+  `tests/data/`.
+- `openghg/standardise/`: incoming scientific dataset parsing and
+  standardisation.
+- `openghg/store/`, `openghg/storage/`, `openghg/objectstore/`: stored data,
+  metadata, indexing, local object stores, and zarr handling.
+- `openghg/dataobjects/`: user-facing data object classes.
+- `openghg/retrieve/`, `openghg/analyse/`, `openghg/transform/`,
```
brendan-m-murphy, line 47: transform is more aligned with standardise than these other submodules. (It would make sense for e.g. applying basis functions to be a "transformation", but that is not how it works in OpenGHG.)

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+## Coding Rules
+
+- Preserve public APIs, metadata keys, file formats, and xarray/pandas behaviour
+  unless the task explicitly changes a contract.
+- Match existing helper APIs, data conventions, validation style, and test
+  fixtures before introducing new patterns.
+- Prefer typed, explicit helpers over duplicated parsing or validation logic.
+- Keep package and test data paths relative to package/test helper APIs. Avoid
+  hard-coded absolute paths.
+- Use Black formatting with `line-length = 110`.
```
brendan-m-murphy, line 59: We should probably switch to ruff.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+- Do not include generated caches, local object stores, virtual environments, or
+  `graphify-out/` changes unless explicitly requested.
+
+## Environment
+
+Prefer `uv` for agent-run setup and commands. Fallback order:
+
+1. Use `uv` (preferred).
+2. If `uv` is unavailable, use an already-active Conda or Micromamba environment.
+3. Do not create new Pixi environments.
```
brendan-m-murphy, line 71: Maybe add "without asking permission"... pixi is useful if you're getting HDF errors, but if you say "never" use it, then it might be difficult to tell an agent to use it.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+- Prefer typed, explicit helpers over duplicated parsing or validation logic.
+- Keep package and test data paths relative to package/test helper APIs. Avoid
+  hard-coded absolute paths.
+- Use Black formatting with `line-length = 110`.
+
+- Add or update focused tests beside the affected code when behaviour changes.
+- Do not include generated caches, local object stores, virtual environments, or
+  `graphify-out/` changes unless explicitly requested.
+
+## Environment
```
brendan-m-murphy, line 65: We should probably have example env set-up scripts. Here are some examples I use; I'm not saying these are ideal, but I set up the codex/chatgpt app to run these commands whenever I create a new worktree in a project.

For codex and openghg I use:
```bash
cd "$CODEX_WORKTREE_PATH"
uv sync --all-extras
uv run pre-commit install
```

For openghg inversions, I use:

```bash
cd "$CODEX_WORKTREE_PATH"

git fetch --no-tags origin devel

if ! git diff --quiet || ! git diff --cached --quiet; then
    echo "Refusing to change the worktree base: tracked changes already exist." >&2
    exit 1
fi

# A Codex-created worktree initially inherits the canonical checkout's HEAD.
# That commit may legitimately be a feature branch not contained in devel; the
# commit remains referenced by that branch, so retargeting this disposable
# worktree does not lose it. Use the named remote-tracking ref rather than the
# repository-wide FETCH_HEAD, which another concurrent fetch can overwrite.
git switch --detach refs/remotes/origin/devel

uv sync
uv tool install tox --with tox-uv --force
```

I should probably add the "fetch origin/devel and branch off of that" option to the openghg env... it depends how heavily you use worktrees whether this is useful.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+
+Prefer `uv` for agent-run setup and commands. Fallback order:
+
+1. Use `uv` (preferred).
+2. If `uv` is unavailable, use an already-active Conda or Micromamba environment.
+3. Do not create new Pixi environments.
+
+Never run `sudo` commands.
```
brendan-m-murphy, line 73: This is probably unnecessary. The codex app has specific permissions and sandbox settings, and I don't know if they could run a sudo command without you entering a password, since they're usually in a new terminal sessions.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+
+```bash
+uv sync --extra dev
+uv run python -c "import openghg"
```
brendan-m-murphy, line 77: Do we need to do this uv run?

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+For optional documentation work:
+
+```bash
+uv sync --all-extras
+uv run python -c "import openghg"
+```
+
+Do not create Pixi environments during automated agent runs; Pixi setup can halt
```
brendan-m-murphy, line 87: I have not experienced this problem with pixi

AGENTS.md
```diff
@@ -0,0 +1,124 @@
...
+or take too long. Use Conda or Micromamba only if `uv` is unavailable and the
+task cannot be validated otherwise.
+
+## Validation
+
+Run, at minimum, `compileall` and the single test file most directly covering
+the changed code. Add `black --check` and `flake8` if formatting or style was
+touched. If a change affects typing, include `mypy` for the affected module.
+
+```bash
```
brendan-m-murphy, line 97: How does this interact with our pre-commit? When should this be run? mypy is slow, running it a lot would be bad. Upgrading to ruff and ty would probably help here.

The single test file limitation doesn't seem like a useful limitation; it can be restricted to "relevant tests".

CHANGELOG.md
```diff
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Added ability to save output of plot_comparison using save_path and ability to pass additional figure modifications.[PR #1672](https://github.com/openghg/openghg/pull/1672)
 - Added more tests and explanation in the tutorial regarding if_exists behaviour.[PR #1693](https://github.com/openghg/openghg/pull/1693)
+- Added instructions for agentic ai efficient and optimal use in AGENTS.md, CLAUDE.md.[PR #1695](https://github.com/openghg/openghg/issues/1695)
```
brendan-m-murphy, line 25: [P2] This is PR #1696, not #1695, and the current URL points to issue #1695. Please change this to `[PR #1696](https://github.com/openghg/openghg/pull/1696)`.

AGENTS.md
brendan-m-murphy: Suggested grouping:

```markdown
- `openghg/standardise/`, `openghg/transform/`, and `openghg/data_processing/`: ingestion-time parsing, standardisation, and dataset transformations.
- `openghg/retrieve/` and `openghg/analyse/`: data access and analysis.
- `openghg/util/`: shared helpers.
```

That avoids implying that `transform` contains inversion-style basis-function transformations.

AGENTS.md
```diff
@@ -0,0 +1,124 @@
+# Agent Instructions for OpenGHG
```
brendan-m-murphy, line 1: [P1] This PR closes #1695, whose title explicitly includes gitleaks/security, but none of the changed files adds gitleaks configuration or security guidance. Please either add enforceable gitleaks integration (prefer repository tooling such as pre-commit and CI over prose an agent may ignore), or split that scope into a follow-up and stop closing the issue. As written, merging silently closes unimplemented scope.

AGENTS.md
brendan-m-murphy: Agreed, but I would keep that migration out of this PR. Instructions should describe the toolchain that actually exists, so for now they should invoke `pre-commit` rather than enumerate Black/flake8/mypy individually. A separate Ruff/ty PR can update configuration, CI, pre-commit, and the one instruction line atomically.

AGENTS.md
brendan-m-murphy: Suggested replacement: “Do not create or switch environments routinely. Use Pixi when explicitly requested or when `uv` cannot provide required compiled/HDF dependencies; explain the reason before switching.” This preserves the useful escape hatch without making Pixi the default. The later duplicate paragraph about Pixi can then be deleted.

AGENTS.md
brendan-m-murphy: Yes, I would keep `uv run`. `uv sync` creates/synchronises the project environment but does not guarantee that later terminals have it activated; `uv run python ...` deterministically uses that environment. The more useful question is whether this import smoke test needs to be prescribed at all—it should be run only after setup or dependency/package changes, not every task.

AGENTS.md
brendan-m-murphy: I would delete this line. Sandbox/permission policy belongs to the agent host, and a repository instruction cannot reliably enforce it. If the concern is dependency installation, the narrower useful rule is “Do not install system packages unless the task requires them; prefer the project environment.”

AGENTS.md
brendan-m-murphy: I suggest replacing the whole fixed minimum with proportional validation:

```markdown
## Validation

Run checks relevant to the files and behaviour changed.

- For Python changes, run the smallest relevant pytest selection.
- Before handoff, run `uv run pre-commit run --files <changed files>` once for changed Python/configuration files.
- For documentation changes, run the affected documentation build or link check.
- Broaden testing for shared/core code, public API changes, or failures suggesting wider impact.
- Do not rerun unchanged passing checks without a reason.
- Report the exact checks run and anything that could not run.
```

That delegates Black/flake8/mypy selection to the repository's pre-commit configuration, avoids prescribing a single test file, and removes unconditional `compileall`/mypy cost. I would also replace “Do not trigger GitHub Actions CI workflows. Run only…” with “Do not manually dispatch workflows unless asked; normal PR CI is authoritative.”

CLAUDE.md
```diff
@@ -0,0 +1,102 @@
+# Claude Guidance for OpenGHG
```
brendan-m-murphy, line 1: [P2] Please avoid maintaining a second ~100-line copy of `AGENTS.md`. Claude Code's own documentation recommends that repositories which already use `AGENTS.md` make `CLAUDE.md` import it. Replace this file with `@AGENTS.md`, followed only by genuinely Claude-specific overrides if any. That gives one source of truth and immediately reduces maintenance and context cost: https://code.claude.com/docs/en/memory#agentsmd

AGENTS.md
brendan-m-murphy: I would make Graphify a repo-scoped, opt-in skill rather than an unconditional startup rule. Keep only a short pointer in `AGENTS.md`, for example: “For orientation in an unfamiliar area or dependency/impact analysis, use the `graphify-navigation` skill when `graphify-out/graph.json` and `manifest.json` exist.”

A minimal `.agents/skills/graphify-navigation/SKILL.md` could be:

```markdown
---
name: graphify-navigation
description: Use tracked Graphify output to orient in unfamiliar OpenGHG code or trace imports, callers, and likely impact paths. Use for repository exploration and impact analysis when graphify-out/graph.json and manifest.json exist; do not use for known-file edits or ordinary text search.
---

# Graphify navigation

1. Check that both tracked output files exist. If either is absent, skip Graphify and use `rg --files` and `rg`; do not regenerate output unless asked.
2. Query `manifest.json` to identify candidate files.
3. Query `graph.json` only for the relevant modules, symbols, or relations; never load it wholesale.
4. Treat results as hints. Verify every conclusion against source code and `rg` before editing.
5. Never edit generated Graphify output unless explicitly requested.
```

Start with only this `SKILL.md`; add a query script later only if repeated use shows that the `jq` expressions are brittle or routinely rewritten. Repo-scoped skills are supported under `.agents/skills`: https://learn.chatgpt.com/docs/build-skills

AGENTS.md
brendan-m-murphy: I would separate automatic bootstrap from agent decision-making.

For Codex, check in the native `.codex` local-environment setup so every new worktree runs:

```bash
cd "$CODEX_WORKTREE_PATH"
uv sync --extra dev
uv run pre-commit install
```

Codex setup scripts are specifically intended to run automatically for new worktrees and can be checked into the repository: https://learn.chatgpt.com/docs/environments/local-environment

Then keep `AGENTS.md` short and conditional:

> Use the environment already provisioned for the worktree; do not recreate it routinely. Run project commands with `uv run`. If dependencies are missing, run `uv sync --extra dev`; add `--extra doc` only for documentation work. If `uv` is unavailable, use the active Conda/Micromamba environment. Use Pixi only when explicitly requested or when `uv` cannot provide required compiled/HDF dependencies, and explain why before switching.

This avoids paying for `--all-extras` on every task and still gives agents a clear recovery path. Branch/worktree retargeting is a separate policy and should only go in the setup script if every Codex-created worktree is intentionally based on `origin/devel`.

AGENTS.md
brendan-m-murphy: ^^^That is from codex... I should have been more specific: do we need to import openghg as a "smoke test"?

AGENTS.md
brendan-m-murphy: Codex:

Correction to my setup suggestion above: `uv run pre-commit install` is unnecessary for disposable Codex worktrees if agents will use `uv run pre-commit run ...` explicitly. Manual runs do not require the Git hook to be installed.

The minimal automatic setup should therefore be:

```bash
cd "$CODEX_WORKTREE_PATH"
uv sync --extra dev
```

Install the Git hook only in developer checkouts where commits should automatically invoke pre-commit.

AGENTS.md
brendan-m-murphy: Codex:

`pre-commit` can be run manually without committing, and files do not need to be staged:

```bash
# All configured hooks on selected files
uv run pre-commit run --files path/to/file.py path/to/other.py

# One configured hook on selected files
uv run pre-commit run flake8 --files path/to/file.py
uv run pre-commit run mypy --files path/to/file.py

# All hooks across the repository
uv run pre-commit run --all-files
```

`pre-commit run` with no options checks staged files but still does not commit. `pre-commit install` only installs the automatic Git hook and is not required for manual runs. Hooks such as Black may modify files, but they never create a commit.

I suggest this exact replacement:

```markdown
## Validation

Run checks proportionate to the change.

- For changed Python or configuration files, run:
  `uv run pre-commit run --files <changed files>`.
- During iteration, a single hook may be run with:
  `uv run pre-commit run <hook-id> --files <changed files>`.
- Run the smallest relevant pytest selection:
  `uv run pytest <test paths or node IDs>`.
- Run `uv run mypy openghg` when changing shared types, public interfaces,
  or code whose typing impact extends beyond the changed files.
- Run all hooks or the full test suite only for broad changes, when requested,
  or when narrower checks reveal wider problems.
- Report the exact checks run and any checks that could not run.
```

This keeps `.pre-commit-config.yaml` as the source of truth for Black, flake8, and mypy while keeping pytest separate because it is not currently a pre-commit hook.

PR #1703, opened 2026-08-07
Add time-resolved Numba fp x flux operator

brendan-m-murphy:
## Summary of changes

- add a public, lazy `fp_x_flux_time_resolved_numba` operator in `openghg.analyse`
- retain `source`, `lat`, and `lon` for later basis projection
- use a Numba/Dask block kernel for regular hourly footprints
- support source selection and relabelling, single-source flux, and regular coarse-frequency flux
- apply an explicit observation-time selector only after constructing the full calculation
- preserve source metadata coordinates
- add optional atomic float32 ppm Zarr persistence with a provenance manifest
- add an optional distributed-worker Numba warm-up helper
- expose Numba through the optional `fp-x-flux` dependency extra
- install the optional `fp-x-flux` extra in GitHub Actions so mypy and tests exercise the Numba path

## Motivation and impact

Verification-games needs a source-resolved footprint-times-flux field without an automatic latitude/longitude reduction. Earlier `forward_model_tests` applied that reduction downstream, so it does not define a separate spatial-retention variant. The explicit `fp_x_flux_time_resolved_numba` name makes this optimized implementation opt-in alongside the existing xarray paths while allowing callers to sum or project the spatial dimensions later.

The native result is dimensionless mole fraction. The persistence helper converts it to ppm per unit source scaling.

## Validation

- 49 focused and regression tests passed, including:
  - direct parity with the existing time-resolved implementation
  - hourly multi-source and single-source inputs
  - 2-hour and 24-hour coarse flux
  - existing integrated CO2 behavior
  - existing irregular-time behavior on the legacy path
  - lazy import coverage
- Black passed
- Flake8 passed
- `uv sync --extra dev --extra fp-x-flux` passed (Numba 0.66.0)
- `uv lock --check` passed
- `git diff --check` passed

## Checklist

- [ ] Closes an existing issue
- [x] Tests added and passed
- [x] Black and Flake8 checks passed
- [x] Documentation updated
- [ ] Tutorial updated — not required for this low-level analysis API
- [ ] Wiki updated — not required
- [x] Changelog entry added
- [x] New package requirement added to `pyproject.toml`
- [ ] `recipes/meta.yaml` updated — not required because Numba is an optional extra and the recipe derives only the base runtime environment


openghg/analyse/_fp_x_flux.py
```diff
@@ -0,0 +1,732 @@
...
+        if len(source_labels) != flux_da.sizes["source"]:
+            raise ValueError("source_labels must contain one label per selected source.")
+        flux_da = flux_da.assign_coords(source=list(source_labels))
+    if _regular_time_step_hours(fp_time_resolved["time"], label="Footprint") != 1:
+        raise ValueError("Footprint time coordinate must be a regular hourly grid.")
+    flux_step_hours = _regular_time_step_hours(flux_da["time"], label="Flux")
+    low_frequency_flux = _low_frequency_flux(flux_da, fp_ds)
+    flux_metadata = flux_da
+
+    if flux_step_hours > _h_back_window_hours(fp_time_resolved):
```
SutarPrasad, line 558: **P1 — Validate flux coverage in the low-frequency branch**

The low-frequency path skips the coverage validation performed by `_flux_with_halo`. For example, when the available flux data is limited to January but the footprints are in March, this branch returns a plausible-looking result by forward-filling the January monthly mean instead of raising an error. That can silently produce scientifically incorrect values.

Please validate that coarse flux covers the required footprint period, including any required lag window, as the high-frequency path does.

SutarPrasad (commented):
I found a coverage-validation issue in the low-frequency flux path.

openghg/analyse/_fp_x_flux.py
brendan-m-murphy: Implemented in 5ab12d93a. Flux interval coverage is now validated before kernel selection, from the earliest release minus the maximum H_back through the latest release. Added low-frequency regression cases for a missing lag halo and coarse flux ending well before the footprint period; all 13 focused tests pass.

SutarPrasad (approved):

PR #1708, opened 2026-08-07
Support irregular times in Numba fp x flux

### LINKED ISSUE #1704: Support irregular footprint release times in fp_x_flux_keep_space
## Context

Draft PR #1703 adds `openghg.analyse.fp_x_flux_keep_space`, using a Numba/Dask block kernel when footprint release times form a regular hourly grid. The optimized kernel deliberately does not yet handle gapped or off-hour release times.

OpenGHG's current time-resolved calculation handles off-grid releases by forward-filling a regularized flux array with a one-grid-step tolerance. That behavior works for existing tests, but the scientific alignment contract is implicit and may associate a release with a flux value in ways callers do not expect.

A loop-over-`H_back` implementation exists in `forward_model_tests`. It avoids the large rolled `time × H_back × lat × lon × source` intermediate and benchmarked better than the existing rolled algorithm, though worse than the regular-grid Numba kernel. As written, it uses exact reindexing and therefore also needs an explicit irregular-time alignment policy.

## Objective

Support irregular footprint release times in `fp_x_flux_keep_space` while preserving original timestamps, spatial/source dimensions, lazy execution, and scientifically explicit flux-time alignment.

## Design work required

Before implementing the fallback, define what flux value represents `flux[release_time - H_back]` when the target is not on the flux grid:

- whether flux timestamps denote interval starts, interval centres, or instantaneous values
- whether alignment should use interval membership, forward-fill, nearest-neighbour, interpolation, or another rule
- the allowed tolerance and behavior when no flux value is valid
- behavior for coarse-frequency flux and targets crossing month boundaries
- whether gapped but hour-aligned releases should share the same policy as genuinely off-hour releases

This policy should be implemented in one alignment helper and tested independently rather than embedded as an incidental `xarray.reindex` option.

## Proposed implementation

1. Add reference tests covering the alignment policy.
2. Add a lazy keep-space fallback that loops over `H_back`:
   - compute target times for each lag
   - align flux using the explicit policy
   - restore the original release-time coordinate
   - multiply and accumulate without constructing the rolled five-dimensional array
3. Dispatch automatically:
   - contiguous regular hourly releases → Numba block kernel
   - gapped/off-hour releases → per-lag xarray/Dask fallback
4. Compute the low-frequency residual through the same shared helper as the regular path.
5. Apply `time_selector` only after the complete calculation.
6. Record the selected implementation in output/cache metadata.
7. Benchmark the fallback against the existing rolled implementation and the regular-grid Numba path.

## Test matrix

- off-hour releases matching the existing `00:37` / `14:23` regression
- gapped, hour-aligned releases
- unsorted and duplicate timestamps
- missing flux before/after the permitted tolerance
- hourly and regular coarse-frequency flux
- single-source and multi-source inputs
- month-boundary residual behavior
- source metadata preservation
- lazy Dask graph and requested output chunks
- post-computation observation-time filtering
- direct parity with the legacy implementation where the chosen alignment semantics overlap

## Acceptance criteria

- Alignment semantics are documented and covered by small numerical reference cases.
- Irregular input preserves the exact original release timestamps.
- Output remains `(source, lat, lon, time)` and lazy.
- Regular hourly inputs continue to use the Numba kernel without performance regression.
- The fallback avoids constructing the full rolled `H_back` flux array.
- Missing/out-of-tolerance flux is handled explicitly rather than silently.

## Out of scope

An irregular-time Numba kernel. That would require per-chunk target-index maps and variable-width flux slabs; it should only be considered after profiling the loop fallback.

brendan-m-murphy:
* **Summary of changes**

This stacked PR depends on #1715. It extends the corrected legacy time-resolved alignment with a source-resolved Numba implementation for genuinely irregular and non-monotonic footprint release times.

- add a public prepared-data `fp_x_flux_time_resolved_numba_core` that assumes monotonic time and suitable chunks
- keep sorting, rechunking, source handling, and output-order restoration in the public wrapper
- use the same native, left-labelled flux intervals as #1715 for every release/lag target
- preserve exact footprint release timestamps in the output
- preserve explicit post-computation observation-time filtering
- preserve prepared footprint time chunks through the residual calculation
- add parity tests for irregular sub-hour release times with hourly and two-hourly flux
- retain tests for irregular ordering, duplicate times, single-source flux, wrapper/core parity, and invalid prepared inputs
- document every production, benchmark, and focused-test function; use full Google-style docstrings for the public API and nontrivial private helpers
- add reproducible OCO2 benchmark and SLURM harnesses

### Relationship to #1715

#1715 defines the corrected behavior of the existing xarray operator:

- the resampling phase comes from the native flux timestamps;
- the regular grid extends across the padded footprint range;
- lag windows are built on that flux-aligned grid; and
- completed windows are assigned to the exact original footprint timestamps.

This PR uses direct interval lookup instead of constructing that intermediate grid, but implements the same mapping. For a release `r` and lag `h`, both operators use the flux interval containing `r - h`.

The parity tests now compare the Numba output directly with `fp_x_flux_time_resolved` from #1715 for:

- sub-hour irregular releases with hourly flux; and
- sub-hour irregular releases with two-hourly flux.

The results agree to the Numba operator's documented float32 precision. Tiny bit-level differences can remain because the xarray and Numba implementations accumulate floating-point terms in different orders; there is no interval-selection or scientific-value difference.

The Numba wrapper additionally supports non-monotonic and duplicate release coordinates by sorting internally and restoring positional order. Those cases have no reliable direct legacy comparison, so they retain independent interval-membership reference tests.

### Performance evidence

Using the January 2023 OCO2-EASTASIA footprint (5,830 irregular releases; 340 × 391 × 24):

- the public Numba wrapper completed the month in 109.96 s with approximately 10.0 GiB MaxRSS
- runtime scaled approximately linearly from 1, 3, 7, and 14 days through the month
- a repaired, sorted, time-chunked Zarr plus the prepared core completed in 58.97 s
- a pure-xarray H_back loop matched checksums but exceeded 16 GiB during the 7-day stage

The earlier pre-#1715 claim that the rolled legacy method differed by 1.30% is intentionally removed: that measurement used the footprint-shifted alignment which #1715 corrects.

- [x] Closes #1704
- [x] All code checks passing - Black, mypy, and Flake8
- [x] Added an entry in the latest `CHANGELOG.md`

Not required for this low-level analysis API:

- Tutorials
- Wiki
- Additional package requirements beyond the optional Numba extra introduced by #1703

### Validation

- `pytest tests/analyse/test_fp_x_flux.py tests/test_modelled_obs_alignment.py tests/analyse/test_modelled_obs.py -q` — 40 passed
- direct corrected-legacy parity for irregular hourly and two-hourly flux
- `mypy --python-version 3.11 openghg/` — no issues in 143 files on the previous revision; production code is unchanged by this restack
- Black and Flake8 checks
- AST docstring audit — every function in the new production, benchmark, and focused-test files is documented
- `git diff --check`

This PR should merge after #1715, then be retargeted to `devel` once #1715 merges.


openghg/analyse/_fp_x_flux.py
```diff
@@ -448,12 +743,197 @@ def _attach_source_coordinates(result: xr.DataArray, flux: xr.DataArray) -> xr.D
...
+    Raises:
+        ValueError: If the selector includes a timestamp absent from the result.
+    """
+    if time_selector is None:
+        return result
+    selected_times = _normalise_time_selector(time_selector)
+    missing = pd.DatetimeIndex(selected_times).difference(pd.DatetimeIndex(result["time"].values))
+    if len(missing):
+        raise ValueError(f"time_selector contains {len(missing)} timestamp(s) outside the result time grid.")
+    return result.sel(time=selected_times)
```
SutarPrasad, line 809: Using .sel() on non-unique index will cause InvalidIndexError. 
ChatGpt Sol 5.6 suggestion: Use positional/boolean selection instead.(caveman output)

openghg/analyse/_fp_x_flux.py
SutarPrasad, line 1215: this is not changed in this PR but spotted and can be updated to below suggestion

time_values = result["time"].values

"time_coverage": {
    "start": str(time_values.min()),
    "end": str(time_values.max()),
    "count": result.sizes["time"],
},

SutarPrasad (changes requested):

SutarPrasad:
feel free to make the changes altogether after merging if that's easier.

openghg/analyse/_fp_x_flux.py
brendan-m-murphy: Maybe we should just raise this error instead (can catch it and emit a more detailed message). I don't think we would want to duplicate a fp_x_flux row.

I suppose implicitly we are assuming fp and obs are already aligned?

openghg/analyse/_fp_x_flux.py
brendan-m-murphy, line 1001: As pointed out above, if `time_selector` does not have unique values then this could raise an error. If used in the way suggested here, then the observations and footprints would already be aligned (presumably) so this would not be a problem. I think catching and re-raising the non-unique index error might be best. Users ought to check this alignment issue before calling this function to avoid wasted compute.

openghg/analyse/_fp_x_flux.py
brendan-m-murphy: Implemented in 7c48a6dd8 using the explicit-error approach discussed here. time_selector now rejects duplicate selector values and raises an actionable ValueError when the result time index is non-unique, advising callers with already-aligned footprint/observation inputs to omit the selector. Duplicate release times remain supported when no selector is applied, with regression tests covering both cases.

openghg/analyse/_fp_x_flux.py
brendan-m-murphy: Implemented in 7c48a6dd8. Manifest time coverage now uses the minimum and maximum coordinate values rather than positional first/last values, so restored non-monotonic release order is represented correctly. The Zarr/manifest test now writes reversed time order and verifies the reported bounds.

brendan-m-murphy: Handled in 7c48a6dd8 together with the live thread above. We now catch this condition before xarray selection and raise a detailed ValueError; duplicate selector values are also rejected rather than silently collapsed. Callers whose footprints and observations are already positionally aligned should omit time_selector.

PR #1715, opened 2026-08-21
Fix irregular time-resolved flux alignment

brendan-m-murphy:
* **Summary of changes** (Bug fix)

This separates the irregular-time alignment correction discussed in #1708 from the new Numba implementation. It changes only the existing xarray-based `fp_x_flux_time_resolved` path.

### Previous method

The existing workaround derived the resampling offset from the padded start of the footprint releases. For a satellite footprint released at `00:37`, it therefore built the high-frequency flux grid at `00:37` too.

That avoids losing irregular footprint timestamps during xarray's exact coordinate alignment, but it changes what the flux timestamps mean: native left-labelled flux intervals such as `[01:00, 02:00)` and `[02:00, 03:00)` are effectively shifted onto a footprint-aligned grid. With time-varying flux, the lagged footprint can then receive a value from the wrong averaging interval.

### Corrected method

Flux timestamps remain the starts of their native, half-open averaging intervals. The calculation now:

1. anchors resampling to the first flux timestamp rather than the first footprint release;
2. includes the flux interval containing the padded start needed for the rolling lag window;
3. constructs lag windows on that flux-aligned grid;
4. extends that grid through the padded footprint end so coarse flux supplies a completed window for every release; and
5. forward-fills each completed lag window to the exact irregular footprint release time, with a one-grid-step tolerance.

For example, for a release at `02:37` and hourly flux labelled on the hour, `H_back=0` uses the interval starting at `02:00`, while `H_back=1` uses the interval starting at `01:00`. The flux intervals are not relabelled to `00:37`.

### Footprint/observation timestamps

The footprint is not resampled. After selecting the correct flux intervals, the lagged flux array is assigned the exact original `fp.time` coordinate before multiplication. The returned `fp_x_flux` therefore retains the footprint timestamps, and observations aligned to those footprint timestamps remain unchanged. The regression test asserts this coordinate identity explicitly.

### Coarse-flux end coverage caught by CI

The first draft correctly changed the grid phase, but generated `full_dates` only through the final timestamp emitted by the resampler. When two-hourly flux had its final relevant label at `06:00` and the footprint release was at `07:00`, the upsampled grid therefore stopped at `06:00`. Reindexing the `07:00` release then reused the completed `06:00` lag window:

- correct targets for release `07:00`: `07:00`, `06:00`, `05:00`;
- targets represented by the reused `06:00` window: `06:00`, `05:00`, `04:00`.

The Numba operator uses direct interval membership and correctly exposed this through `test_regular_coarse_flux_matches_existing_time_resolved_method[2]`; its expected values were not changed.

Commit `7e8d34c46` fixes the edge case by retaining the native flux phase while extending `full_dates` through the padded footprint end. This supplies the `07:00` lag window without shifting the native flux intervals. The complete Numba parity file and both irregular-alignment suites now pass locally.

Initial local verification ran only `tests/test_modelled_obs_alignment.py` and `tests/analyse/test_modelled_obs.py`. Those seven tests passed, but that selection omitted the PR #1703 Numba/legacy parity suite in `tests/analyse/test_fp_x_flux.py`, which is why this coarse-flux case was not caught before the first push.

### Scope

This PR deliberately contains no Numba operator or new public API, so the legacy alignment correction can be reviewed independently from #1708.

- [ ] Closes #xxxx (no separate issue)
- [x] [Tests added and passed](https://docs.openghg.org/development/python_devel.html#testing) locally
- [x] Black and flake8 pass locally; mypy passed on the previous CI revision
- [ ] Tutorials updated/added
- [ ] Wiki updated
- [x] No new package requirements

Not needed for this focused internal correction: user documentation, tutorials, wiki, and package requirements.

### Verification

- `.venv/bin/pytest tests/analyse/test_fp_x_flux.py tests/test_modelled_obs_alignment.py tests/analyse/test_modelled_obs.py -q` — 20 passed
- Black and flake8 on the changed implementation
- `git diff --check`
- GitHub Actions for `7e8d34c46` — documentation and both test matrices passed

### DIFF
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -187,9 +187,7 @@ def _make_high_freq_flux(
         flux_high_freq = flux_resampler.ffill()
 
     # Fill gaps on the regular, flux-aligned grid before constructing lag windows.
-    full_dates = pd.date_range(
-        flux_high_freq.time.values[0], flux_high_freq.time.values[-1], freq=freq
-    ).to_numpy()
+    full_dates = pd.date_range(flux_high_freq.time.values[0], end, freq=freq, inclusive="left").to_numpy()
     flux_high_freq = flux_high_freq.reindex({"time": full_dates}, method="ffill")
 
     # create rolling windows
--- a/openghg/analyse/_modelled_obs.py
+++ b/openghg/analyse/_modelled_obs.py
@@ -165,19 +165,18 @@ def _make_hf_flux_rolling_avg_array(
 def _make_high_freq_flux(
     flux: xr.DataArray,
     fp: xr.DataArray | xr.Dataset,
-    *,
-    align_to_fp_time: bool = True,
 ) -> xr.DataArray:
     fp_highest_res_hours = _fp_time_and_h_back_freq_gcd(fp)
     start, end = _padded_flux_slice_start_and_end(fp)
-    offset = time_of_day_offset(start)
-
-    # select values with start time padded to include first rolling window backwards in time
-    flux_high_freq = flux.sel(time=slice(start, end))
-
-    # resample flux
-    freq = f"{fp_highest_res_hours}h"
     flux_res_hours = calc_hourly_freq(flux.time, input_nanoseconds=True)
+    offset = time_of_day_offset(flux.time.values[0])
+
+    # Include the flux interval containing the padded start of the rolling window.
+    flux_slice_start = start - np.timedelta64(flux_res_hours, "h")
+    flux_high_freq = flux.sel(time=slice(flux_slice_start, end))
+
+    # Resample on the flux interval grid, not the offset of the first footprint release.
+    freq = f"{fp_highest_res_hours}h"
     flux_resampler = flux_high_freq.resample({"time": freq}, offset=offset)
 
     if flux_res_hours <= fp_highest_res_hours:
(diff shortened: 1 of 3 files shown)


brendan-m-murphy:
## Why the regression test shows the previous alignment is wrong

The conclusion depends on the explicit time-coordinate contract used here: a flux value timestamped `t` represents the left-labelled, half-open interval `[t, t + cadence)`. If the flux timestamps instead represented instantaneous samples or interval centres, we would need a different alignment rule.

The regression test deliberately removes other effects:

- one spatial cell;
- `fp_time_resolved = 1` for `H_back = [0, 1]`;
- `fp_residual = 0`;
- hourly flux increasing by one, so every interval is identifiable; and
- irregular releases at `00:37`, `01:44`, `02:51`, and `03:58`.

The relevant flux intervals are:

| Flux timestamp | Represented interval | Value |
|---|---|---:|
| `23:00` | `[23:00, 00:00)` | 3 |
| `00:00` | `[00:00, 01:00)` | 4 |
| `01:00` | `[01:00, 02:00)` | 5 |
| `02:00` | `[02:00, 03:00)` | 6 |
| `03:00` | `[03:00, 04:00)` | 7 |

For the release at `00:37`, `H_back=0` targets `00:37` and must use value 4; `H_back=1` targets `23:37` and must use value 3. With unit footprints, the hand-derived result is therefore `4 + 3 = 7`.

### Previous method

```mermaid
flowchart LR
    F3["Flux 23:00<br/>value 3"] --> B3["Shifted label 22:37"]
    F4["Flux 00:00<br/>value 4"] --> B4["Shifted label 23:37"]
    F5["Flux 01:00<br/>value 5"] --> B5["Shifted label 00:37"]

    R["Footprint release<br/>00:37"] --> H0["H_back=0<br/>takes shifted 00:37 → 5"]
    R --> H1["H_back=1<br/>takes shifted 23:37 → 4"]
    B5 --> H0
    B4 --> H1
    H0 --> S["Previous result = 9"]
    H1 --> S
```

The previous code derives the resampling phase from the padded footprint start. A release at `:37` therefore creates a grid at `:37`. Native flux timestamp `01:00` is placed in the shifted bin labelled `00:37`, so a target at `00:37` receives value 5 even though value 5 represents `[01:00, 02:00)`—an interval beginning 23 minutes in the future.

That is the semantic error: interval membership is changed, rather than merely the presentation of the coordinates.

### Corrected method

```mermaid
flowchart LR
    I3["[23:00, 00:00)<br/>value 3"] --> T1["Target 23:37<br/>H_back=1"]
    I4["[00:00, 01:00)<br/>value 4"] --> T0["Target 00:37<br/>H_back=0"]
    I5["[01:00, 02:00)<br/>value 5"] -. "contains neither target" .-> X["Not selected"]

    T1 --> S["Correct result = 3 + 4 = 7"]
    T0 --> S
```

The corrected method constructs lag windows on the native flux phase, then maps the completed window to the exact release timestamp:

```mermaid
flowchart LR
    F["Native flux grid<br/>…, 23:00, 00:00, 01:00, …"] --> W["Construct lag windows<br/>on flux grid"]
    W --> A["Select containing window<br/>for release 00:37"]
    A --> M["Multiply by footprint<br/>at original 00:37"]
    M --> O["Output timestamp<br/>00:37"]
```

The footprint and observations are not resampled. The test explicitly asserts that the result's time coordinate is identical to `fp.time`.

### Exact comparison

Running the exact pre-change helper and corrected helper against this fixture gives:

| Release | Interval-membership values | Correct result | Previous selected values | Previous result |
|---|---:|---:|---:|---:|
| `00:37` | `4 + 3` | **7** | `5 + 4` | **9** |
| `01:44` | `5 + 4` | **9** | `6 + 5` | **11** |
| `02:51` | `6 + 5` | **11** | `7 + 6` | **13** |
| `03:58` | `7 + 6` | **13** | `7 + 7` | **14** |

The expected values are therefore a hand-derived interval-membership oracle, not values copied from the corrected implementation.

Scope caveat: this test isolates and proves the footprint-derived **phase shift** is wrong under left-labelled interval semantics. It does not claim to validate every aspect of cadence inference for every irregular dataset.

rt17603:
I've read through the description and the regression test and this does seem like a genuine correction to me.

SutarPrasad (approved):

PR #1718, opened 2026-08-25
1716 enforce same max level values for obs and footprints

### LINKED ISSUE #1716: Enforce same max_level values for obs and footprints
### What happened?

Currently different max level values can be passed for obs and fp only for retrieval which seems correct. However during the model_scenario implementation having different values for obs and fp means they are in different space and can cause significantly distint values for prior outputs than expected.

For modelscenario the difference in max_level should be identified and error should be raised

```Python

```

### Relevant log output

```Python

```

_No response_

### Environment

<details>

</details>

SutarPrasad:
The PR implements a `max_level` check during Modelscenario creation.

1. Compares max_level passed to the obs column data selection function (either get_obs_column or ModelScenario directly) with the footprints max_level.
2. Reads the footprint value from dataset attributes, falling back to footprint metadata.
3. Raises a clear error when the values differ.
4. Raises an actionable error when the footprint value is missing or "unknown".
5. Advises users to re-standardise the footprint or use data_manager(...).update_attributes(...) and update_metadata(...) for writable stores.
6. Preserves column-footprint max_level during both ACRG and PARIS standardisation.
7. Prevents the footprint level from being inferred from the requested observation level.

The max_level supplied to get_obs_column(), either directly or through ModelScenario, is used when processing the retrieved column observations. The effective value is recorded in:
ObsColumnData.data.attrs["max_level"]
ModelScenario compares this value with the max_level used to process the footprint. The footprint value is read from its dataset attributes, with footprint metadata used as a fallback.
When an existing ObsColumnData object is supplied directly, an additional max_level argument is not required because the value is read from the object itself.
If the values differ, or the footprint does not contain a valid max_level, ModelScenario raises an error before calculating modelled observations.

**Tests**
9. Added coverage for:
10. Matching observation and footprint levels.
11. Mismatched levels from retrieved objects.
12. Ensuring dataset values are checked rather than only the constructor argument.
13. Missing footprint max_level.
14. Actionable object-store repair guidance.
15. Preservation of max_level by ACRG and PARIS footprint parsers.

- [x] Closes #1716 (Replace xxxx with the Github issue number)
- [x] Tutorials updated/added
- [x] Wiki updated

### DIFF
--- a/doc/source/tutorials/local/Adding_data/Adding_co2_satellite_data.rst
+++ b/doc/source/tutorials/local/Adding_data/Adding_co2_satellite_data.rst
@@ -92,6 +92,15 @@ calculation from the stored footprint variables: time-resolved data goes through
 the time-resolved CO2 calculation, while ``fp``-only data goes through the
 integrated calculation.
 
+.. important::
+
+   ``max_level`` must match the fixed value used when the column footprint was processed.
+   Using a different value for the column observations places the observations and footprint
+   in different vertical spaces. For an existing footprint, inspect ``footprint.metadata["max_level"]``
+   (or ``footprint.data.attrs["max_level"]``) and use that value when retrieving the column
+   observations and creating ``ModelScenario``. The example footprint below uses
+   ``max_level=17``.
+
 .. code:: ipython3
 
     from openghg.analyse import ModelScenario
--- a/doc/source/tutorials/local/Adding_data/Adding_co2_satellite_data.rst
+++ b/doc/source/tutorials/local/Adding_data/Adding_co2_satellite_data.rst
@@ -94,7 +94,8 @@ integrated calculation.
 
 .. important::
 
-   ``max_level`` must match the fixed value used when the column footprint was processed.
+   For comparison to the modelled observations, ``max_level`` must match the fixed value used
+   when the column footprint was processed.
    Using a different value for the column observations places the observations and footprint
    in different vertical spaces. For an existing footprint, inspect ``footprint.metadata["max_level"]``
    (or ``footprint.data.attrs["max_level"]``) and use that value when retrieving the column
--- a/doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
+++ b/doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
@@ -65,11 +65,12 @@ We can now create a ``ModelScenario`` linking satellite observations with ancill
 
 .. important::
 
-   For a column or satellite ``ModelScenario``, the observation ``max_level`` must be the
-   same as the fixed ``max_level`` used to process the footprint. Otherwise, the observations
-   and footprints describe different vertical spaces and the modelled prior contribution can
-   be incorrect. The footprint in this tutorial was processed with ``max_level=17``, so the
-   column observations must also use 17. For previously retrieved objects, these values are
+   For a column or satellite ``ModelScenario``, the ``max_level`` used to retrieve the column
+   observations must be the same as the fixed ``max_level`` used to process the footprint.
+   Otherwise, the observations and footprints describe different vertical spaces and the
+   modelled prior contribution can be incorrect. The footprint in this tutorial was processed
+   with ``max_level=17``, so the value used to retrieve the column observations must also be 17.
+   For previously retrieved objects, these values are
    available from ``obs_column_data.data.attrs["max_level"]`` and
    ``fp_column_data.metadata["max_level"]`` (or ``fp_column_data.data.attrs["max_level"]``).
 
@@ -172,7 +173,7 @@ same data needs to be used for multiple different scenarios.
 
 .. jupyter-execute::
 
-    scenario_direct = ModelScenario(obs_column=obs_column_data, footprint=fp_column_data, flux=flux_data, bc=bc_results, platform="satellite", max_level=17)
+    scenario_direct = ModelScenario(obs_column=obs_column_data, footprint=fp_column_data, flux=flux_data, bc=bc_results, platform="satellite")
 
 .. note::
 
--- a/doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
+++ b/doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
@@ -63,6 +63,16 @@ We'll use helper functions from ``openghg.tutorial`` to populate example data:
 
 We can now create a ``ModelScenario`` linking satellite observations with ancillary inputs.
 
+.. important::
+
+   For a column or satellite ``ModelScenario``, the observation ``max_level`` must be the
+   same as the fixed ``max_level`` used to process the footprint. Otherwise, the observations
+   and footprints describe different vertical spaces and the modelled prior contribution can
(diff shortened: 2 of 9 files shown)


rt17603:
Point of clarity:
> 1. Compares the effective max_level stored in column observation data with the footprint’s max_level.

may be more easily understood as:

> 1. Compares max_level passed to the obs column data selection function (either `get_obs_column` or `ModelScenario` directly) with the footprints max_level.

Another point which could be added to this (if correct):
> The input max_level value is taken from the metadata stored within the `ObsColumnData` object returned from the `get_obs_column` function.

doc/source/tutorials/local/Adding_data/Adding_co2_satellite_data.rst
```diff
@@ -92,6 +92,15 @@ calculation from the stored footprint variables: time-resolved data goes through
 the time-resolved CO2 calculation, while ``fp``-only data goes through the
 integrated calculation.
 
+.. important::
+
+   ``max_level`` must match the fixed value used when the column footprint was processed.
```
rt17603:
```suggestion
   For comparison to the modelled observations, ``max_level`` must match the fixed value used when the column footprint was processed.
```

doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
```diff
@@ -63,14 +63,24 @@ We'll use helper functions from ``openghg.tutorial`` to populate example data:
 
 We can now create a ``ModelScenario`` linking satellite observations with ancillary inputs.
 
+.. important::
+
+   For a column or satellite ``ModelScenario``, the observation ``max_level`` must be the
```
rt17603:
```suggestion
   For a column or satellite ``ModelScenario``, the ``max_level`` used to retrieve the column observations must be the
```

doc/source/tutorials/local/Analysing_data/Comparing_satellite_observations_with_emissions.rst
```diff
@@ -63,14 +63,24 @@ We'll use helper functions from ``openghg.tutorial`` to populate example data:
...
+   same as the fixed ``max_level`` used to process the footprint. Otherwise, the observations
+   and footprints describe different vertical spaces and the modelled prior contribution can
+   be incorrect. The footprint in this tutorial was processed with ``max_level=17``, so the
+   column observations must also use 17. For previously retrieved objects, these values are
```
rt17603:
```suggestion
   value used to retrieve the column observations must also use 17. For previously retrieved objects, these values are
```

openghg/analyse/_scenario.py
```diff
@@ -156,7 +156,8 @@ def __init__(
         network: Network name e.g. "AGAGE".
         domain: Domain name e.g. "EUROPE".
         platform: Platform name e.g "satellite", "column-insitu".
-        max_level: Maximum level for processing.
+        max_level: Maximum level for processing column observations. This must
```
rt17603:
```suggestion
        max_level: Maximum level for processing column observations. If comparing to the modelled observations using a footprint, this must
```

tests/analyse/test_scenario.py
```diff
@@ -1643,6 +1647,69 @@ def test_model_scenario_col_fp_data_merge(satellite_name_store):
...
+        max_level=3,
+        obs_region="brazil",
+        start_date="2016-01-01 14:59:12.500000+00:00",
+        end_date="2016-01-01 18:10:16.500000+00:00",
+        store=satellite_name_store,
+    )
+    fp_column_data = get_footprint(
+        satellite="gosat",
+        domain="southamerica",
+        obs_region="brazil",
```
rt17603:
```suggestion

```
I wouldn't think `obs_region` and `domain` would both be needed here. (I'm not sure if `obs_region` is needed for footprint at all actually, if this was added more recently it would be useful to understand the logic).

tests/analyse/test_scenario.py
```diff
@@ -1643,6 +1647,69 @@ def test_model_scenario_col_fp_data_merge(satellite_name_store):
...
+    assert int(fp_column_data.metadata["max_level"]) == 17
+
+    with pytest.raises(ValueError, match=r"observations use 3, while footprints use 17"):
+        ModelScenario(
+            obs_column=obs_column_data,
+            footprint=fp_column_data,
+            platform="satellite",
+            # Deliberately match the footprint to prove that validation uses
+            # the effective value stored on the column observations.
+            max_level=17,
```
rt17603:
```suggestion

```
If the obs data and footprint data objects were being passed directly, I wouldn't think `max_level` as an additional parameter should be needed here. If it is, that's perhaps a bug.

Removing this parameter would also be a better test of the expected behaviour as both values for `max_level` will need to be extracted from the provided data.

tests/analyse/test_scenario.py
```diff
@@ -1643,6 +1647,69 @@ def test_model_scenario_col_fp_data_merge(satellite_name_store):
...
+        max_level=17,
+        obs_region="brazil",
+        start_date="2016-01-01 14:59:12.500000+00:00",
+        end_date="2016-01-01 18:10:16.500000+00:00",
+        store=satellite_name_store,
+    )
+    fp_column_data = get_footprint(
+        satellite="gosat",
+        domain="southamerica",
+        obs_region="brazil",
```
rt17603:
```suggestion

```
Same comment as previous - `obs_region` shouldn't be needed

rt17603 (commented):
Looks good - a few small clarifications to review and then happy for this to be merged.

tests/analyse/test_scenario.py
SutarPrasad: Good point,obs_region isn’t needed to uniquely retrieve the footprint in this test, so I’ve removed it here.

The satellite footprint standardisation workflow currently requires satellite and obs_region together, but we don’t yet have a dedicated test showing obs_region disambiguating two footprints in the same domain.

rt17603: This is outside the scope of this PR but, unless something has changed in the way the keys are being used, I wouldn't think we'd need an `obs_region` associated with a footprint at all. The `obs_region` keyword is purely to describe the area over which a set of satellite observations have been selected. If the `obs_region` keyword  was associated with a footprint I imagine it would be for convenience of linking back to a set of observations(?) but if the domain is known it shouldn't be needed.

SutarPrasad: I have created the separate issue https://github.com/openghg/openghg/issues/1719.
This PR should be now ready for approval. Feel free to resolve the comment, if we want to move the discussion over to the issue.

rt17603:
> Point of clarity:
> 
> > 1. Compares the effective max_level stored in column observation data with the footprint’s max_level.
> 
> may be more easily understood as:
> 
> > 1. Compares max_level passed to the obs column data selection function (either `get_obs_column` or `ModelScenario` directly) with the footprints max_level.
> 
> Another point which could be added to this (if correct):
> 
> > The input max_level value is taken from the metadata stored within the `ObsColumnData` object returned from the `get_obs_column` function.

Based on the discussion, could the PR description be updated to include this?

rt17603 (approved):
Looks good!

Issue #1338, opened 2025-04-28
Ensure "units" is added for all data types


rt17603:
### Is your feature request related to a problem?

In general, we should ensure that the "units" data variable attribute is included for all data variables for all data types. Practically, we should at least for the data variables used for forward modelling for inversions.

This includes:
 - relevant data variables for boundary conditions
 - relevant data variables for flux/emissions

brendan-m-murphy:
Related issue on syncing metadata and attributes for all data types: https://github.com/openghg/openghg/issues/1253

We might not need to actually sync metadata and attributes to ensure units are added, but currently for surface obs, "required" attributes are enforced via `align_metadata_attributes` (or `sync_surface_metadata` ?)

SutarPrasad:
To start with the units should be forced based on the schema of the datatypes. There are lot of data variables which don't have units, or data variables which have binary values. 

- Data variables that are not required to be handled under the pint can be called dimensionless or ignore them.

- We need to check if pint.quantify() complains if all data variables don't have units is something we need to check before ignoring.

brendan-m-murphy:
I prototyped this in the stacked draft #1721 (on the xarray-validate refactor in #1720), and the implementation exposed why “all data variables have units” is too broad.

Proposed schema policy:

- Require a nonblank `units` attribute on the canonical surface-observation signal (for example `ch4`), but not on every ancillary variable.
- Do **not** require units on `number_of_observations` (or suffixed observation-count variables). Cardinalities should not participate in mole-fraction conversion.
- Require dimensionally compatible units on `flux`, all selected footprint signal variants, and boundary `vmr_n/e/s/w`.
- Require physical coordinate units for latitude/longitude, height, and numeric back time. Do not require units attrs on decoded datetime coordinates because xarray moves CF time units into `.encoding`.
- Use OpenGHG's existing CF-aware Pint registry so `mol/mol`, `1`, `ppm`, `ppb`, and `1e-9` remain valid scaled dimensionless representations.

The prototype also found a concrete version of the cardinality bug: `ModelScenario.convert_units()` excluded names ending in `_number_of_observations`, but not the exact ICOS-style `number_of_observations`. If that variable has `units="1"`, conversion to ppm multiplies its values by 1e6. #1721 adds an exact-name regression test and excludes both forms from conversion.

[...]

Issue #549, opened 2023-01-27
Incorporate attributes into schema for each data type


rt17603:
### Is your feature request related to a problem?

At the moment the .schema() method for each class provides details on the file format for the data variables and coordinates but does not include details on attributes. Would be useful to incorporate these details here.

### Describe the solution you'd like

Needed attributes are defined as a part of schema method.

brendan-m-murphy:
The schema/attribute work is now split into a four-PR draft stack so the behavior changes can be reviewed independently:

1. #1720 — behavior-preserving refactor of the existing `DataSchema` checks onto xarray-validate, including uv/Pixi/legacy-Conda setup and developer docs. No unit or attribute requirements.
2. #1721 — explicit unit requirements for named physical variables and coordinates. This deliberately avoids wildcarding ancillary variables such as `number_of_observations`.
3. #1722 — general non-unit attribute support via `required_attrs` (named variables/coordinates) and `dataset_attrs`, applied only to stable internal-format metadata: surface/footprint/BC `long_name` and flux `source`/`species`.
4. #1723 — moves all current declarations to packaged `openghg/store/data_schemas.json`, including exact dynamic surface/column substitutions and composable footprint fragments.

The JSON format is OpenGHG-owned rather than `xarray-validate.serialize()`. In xarray-validate 0.0.5, JSON round trips are not guaranteed; serialization drops callable checks and generic NumPy dtype information, and typed attribute schemas are not directly JSON-serializable. Those callable checks are currently required to preserve OpenGHG's dimension-membership and optional-coordinate behavior.

This keeps each schema readable as a datatype specification while preserving existing callers through `Class.schema()`, `Class.validate_data()`, and `DataSchema.validate_data()`.

PR #1720, opened 2026-08-28
Refactor DataSchema with xarray-validate

brendan-m-murphy:
* **Summary of changes**

Refactors the existing `DataSchema` implementation to use `xarray-validate==0.0.5` behind the current OpenGHG API.

This PR deliberately adds no new validation features. It preserves the existing behavior:

- required named data variables;
- required dimensions as membership checks (order and additional dimensions remain allowed);
- dtype checks, including optional dtype-only coordinates;
- extra data variables remain allowed;
- xarray-validate schema errors are translated to OpenGHG `ValidationError`;
- the historical missing-variable-dimension `ValueError` remains unchanged.

Packaging updates include the runtime dependency, an explicit `attrs>=22` lower bound required by the validator API, refreshed uv/Pixi locks, and the legacy Conda environment pip entry. The developer data-specification docs now explain the compatibility API and behavior.

This is the base of a four-PR draft stack:

1. #1720 — this behavior-preserving refactor;
2. #1721 — explicit named-variable and coordinate unit requirements;
3. #1722 — stable non-unit attribute requirements;
4. #1723 — JSON-backed schema declarations.

Related to #549 and #1338; it does not close either issue.

- [x] Tests added and passed: `tests/store/test_data_schema.py` and `tests/store/test_flux.py` (21 passed, 1 skipped; includes the checked-in CARDAMOM NetCDF)
- [x] Formatting/lint checks passed for changed Python files (Black and Flake8)
- [x] Documentation updated
- [x] New runtime requirements added to `pyproject.toml`, `uv.lock`, `pixi.lock`, and `environment.yaml`
- [x] Locked Pixi dev environment imports `attrs` and `xarray_validate`

*Not applicable / follow-up*

- No changelog entry: this is an internal behavior-preserving refactor.
- No tutorial or wiki update.
- The published Conda recipe cannot directly express the PyPI-only `xarray-validate` package; reproducible mixed Conda/PyPI installation is covered by Pixi and `environment.yaml`.
- The complete store suite currently reports one unrelated `SiteMet` object-store retrieval failure; all schema-focused tests pass.

### DIFF
--- a/doc/source/development/specifications/data_spec.rst
+++ b/doc/source/development/specifications/data_spec.rst
@@ -4,6 +4,52 @@ Data Specification
 
 Here we set out the specfication for the data we expect for the different storage classes.
 
+Schema validation
+-----------------
+
+Each storage class exposes its internal xarray format through ``schema()``.
+The returned :class:`openghg.store.DataSchema` records required data variables,
+their dimensions, and NumPy data types. For example:
+
+.. code-block:: python
+
+    from openghg.store import Flux
+
+    schema = Flux.schema()
+    schema.validate_data(dataset)
+
+Storage classes also provide ``validate_data()`` as a convenient class-level
+entry point:
+
+.. code-block:: python
+
+    Flux.validate_data(dataset)
+
+Validation is implemented by ``xarray-validate`` behind the existing
+``DataSchema`` API. Schema-library failures are exposed as OpenGHG
+``ValidationError`` exceptions, so callers do not need to depend on the
+validation library directly. For compatibility, a missing dimension on a
+required variable continues to raise ``ValueError``.
+
+When adding or changing an internal format, define required variables and
+data types in the storage class's ``schema()`` method:
+
+.. code-block:: python
+
+    import numpy as np
+
+    from openghg.store import DataSchema
+
+    DataSchema(
+        data_vars={"example": ("time",)},
+        dtypes={"example": np.floating, "time": np.datetime64},
+    )
+
+Extra variables remain allowed, and dimensions listed for a required variable
+must be present but may appear in a different order. Dtype constraints for
+coordinates remain optional when the coordinate is absent, matching the
+historical OpenGHG schema behaviour.
+
 ObsSurface
 ----------
 
--- a/openghg/store/_data_schema.py
+++ b/openghg/store/_data_schema.py
@@ -1,146 +1,62 @@
-from dataclasses import dataclass
-import logging
+from dataclasses import dataclass, field
+from functools import partial
+
 import numpy as np
-from xarray import Dataset
+from xarray import DataArray, Dataset
+import xarray_validate as xv  # type: ignore[import-untyped]
 
 from openghg.types import ValidationError
-
-logger = logging.getLogger("openghg.store")
-logger.setLevel(logging.DEBUG)  # Have to set level for logger as well as handler
 
 __all__ = ["DataSchema"]
 
 
+def _check_dims(expected: tuple[str, ...], data: DataArray) -> None:
+    for dim in expected:
+        if dim not in data.dims:
+            raise ValueError(
+                f"Missing dimension for data variable: {data.name}, {dim}. Current dims: {data.dims}"
+            )
+
+
 @dataclass
 class DataSchema:
-    """
-    Create schema for a data type based on inputs for components of the data.
-    Expected format is based against an xarray Dataset.
-
-    Example inputs:
-        DataSchema(
-            data_vars = {"fp": ("time", "lat", "lon"), ...},
-            dtypes = {"fp" : np.floating, ...},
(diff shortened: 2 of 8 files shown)


openghg/store/_data_schema.py
```diff
@@ -1,146 +1,65 @@
...
+from functools import partial
+
 import numpy as np
-from xarray import Dataset
+from xarray import DataArray, Dataset
+import xarray_validate as xv  # type: ignore[import-untyped]
 
 from openghg.types import ValidationError
 
-logger = logging.getLogger("openghg.store")
```
brendan-m-murphy, line 8: Why remove the logger setup?

SutarPrasad: any valid response on this?

SutarPrasad:
I can see the tests are created using dummy data. Its worth checking current test setup and using some of the real data.

openghg/store/_data_schema.py
brendan-m-murphy: The logger was only used by the three legacy `_check_*` methods to report that an optional constraint was absent. Those methods were replaced by the `xarray_validate` schema, so no logger calls remained. Since #1720 is intended to be behavior-preserving, I restored the exact module-level setup in bb33bde5e to retain the previous import-time logging configuration.

brendan-m-murphy:
Addressed the real-data testing feedback in bb33bde5e. `test_flux_schema_with_file_data` now opens the checked-in CARDAMOM file `tests/data/flux/co2-gpp-cardamom_EUROPE_2012.nc` and validates it directly with `Flux.schema()`. This exercises the real variable dimensions and dtypes (including the intentionally order-insensitive dimensions) rather than another constructed Dataset. The PR #1720 test slice now passes: 21 passed, 1 skipped; the rebased final stack passes 175 passed, 7 skipped, 2 xfailed.

brendan-m-murphy:
Docstring-review pilot against the proposed policy in #1738 (non-blocking while that policy is under review):

- `DataSchema` remains a public constructor, but its replacement one-line docstring no longer explains `data_vars`, `dtypes`, or `dims`. A concise `Args` contract should capture the dimension-membership and optional-coordinate semantics and link to the new data-specification page for detail.
- `DataSchema.validate_data` should document its `data` input and the deliberately preserved `ValidationError` versus missing-required-dimension `ValueError` outcomes. Those are caller-visible recovery branches already established by this PR's prose and tests.
- `_check_dims` owns that compatibility behaviour, and the three new tests in `test_data_schema.py` distinguish separate scenarios. One meaningful sentence on each would preserve their maintenance intent.

The proposed Ruff `D101`-`D103` ratchet passes here because the public objects have docstrings and the helper is private. That is the expected mechanical result; this semantic review is the complementary layer. The authored `data_spec.rst` material is strong, so the docstrings can stay concise and link rather than duplicate it.

doc/source/development/specifications/data_spec.rst
```diff
@@ -4,6 +4,52 @@ Data Specification
...
+    import numpy as np
+
+    from openghg.store import DataSchema
+
+    DataSchema(
+        data_vars={"example": ("time",)},
+        dtypes={"example": np.floating, "time": np.datetime64},
+    )
+
+Extra variables remain allowed, and dimensions listed for a required variable
```
rt17603: We may want to rephrase this so this doesn't describe a change in behaviour but just what the current behaviour is. e.g. replace "remain" with "are"

```suggestion
Extra variables are allowed, and dimensions listed for a required variable
```

rt17603 (commented):
Looks good to me - just a small comment on the documentation phrasing.

doc/source/development/specifications/data_spec.rst
```diff
@@ -4,6 +4,52 @@ Data Specification
...
+    import numpy as np
+
+    from openghg.store import DataSchema
+
+    DataSchema(
+        data_vars={"example": ("time",)},
+        dtypes={"example": np.floating, "time": np.datetime64},
+    )
+
+Extra variables are allowed, and dimensions listed for a required variable
```
rt17603, line 48: Apologies if this wasn't clear but I was requesting that this paragraph be reviewed for present tone. That meant "coordinates remain optional" could also be updated to "coordinates are optional" and a bit more detail on what "historical OpenGHG schema behaviour" means - if this is something different to what is described in this paragraph.

brendan-m-murphy: Sorry you left a specific fix, I thought you wanted that implemented.

rt17603: Yes, I did but was including that as a suggestion - will present differently next time if that's confusing.

rt17603 (dismissed):
Looks good!

rt17603 (approved):

PR #1728, opened 2026-09-03
safeguard against release tags

### LINKED ISSUE #1123: Check tag adhere to "v[0-9]" to trigger a release
### What is your issue?

Then we can use tags for other purposes, like marking code before deleting it.

SutarPrasad:

The issue suggested requiring release tags to use a v prefix. However, OpenGHG has a long history of using unprefixed version tags such as 0.19.0. I have therefore kept the existing convention and added a safeguard so that release jobs run only for tags matching the exact X.Y.Z format. Other tags can still be pushed for reference or to preserve checkpoints—for example, "deprecated-parsers" without publishing a new version to PyPI or conda. Normal CI tests will still run for those tags. This means the recent discussion we had for the work on Brendans CI can have a tag even after merging into devel.

- [x] Closes #1123 (Replace xxxx with the Github issue number)

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -12,11 +12,17 @@ on:
     paths-ignore:
       - 'CHANGELOG.md'
   workflow_dispatch:
+    inputs:
+      tag:
+        description: "Tag value to validate without releasing"
+        required: true
+        default: "1.2.3"
+        type: string
 
 jobs:
   release_tag:
     name: Check for a release tag
-    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/')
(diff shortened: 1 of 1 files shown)


rt17603 (commented):
Looking over the code this seems sensible - was there a way to check parts of this workflow without triggering a release? (Not necessarily the full workflow - just bits like creating the `release_tag=true/false` for instance)

rt17603 (dismissed):

SutarPrasad:
Added a manual text field for workflow dispatcher to test the conditions:
is_release=True
https://github.com/openghg/openghg/actions/runs/33857873333/job/100975263323
is_release=False
https://github.com/openghg/openghg/actions/runs/33858546721/job/100977392946
The added text field serve as a way to test the release workflow conditions added without actually releasing as the release workflow depends on the event: push && is_release:True

rt17603 (approved):
Looks good!

Issue #1714, opened 2026-08-20
Vectorize ModelScenario multi-sector fp x flux calculation


brendan-m-murphy:
## Context

PR #1713 routes `ModelScenario.footprints_data_merge(calc_fp_x_flux=True)` through `fp_x_flux_time_resolved_numba`, but preserves the current `split_by_sectors` behavior: `calc_modelled_obs` loops over sectors and calls the calculation once per sector. The integration therefore collapses the Numba operator's internal `source` axis on each call to retain the established output dimensions.

## Goal

Avoid the per-sector loop by passing source-resolved flux through the Numba operator once and deriving both total and sectoral outputs from that result.

## Acceptance criteria

- Build source-resolved flux with stable source labels for one Numba call.
- Produce total `fp_x_flux` with dimensions `(lat, lon, time)` by summing `source`.
- Produce `fp_x_flux_sectoral` with dimensions `(source, lat, lon, time)` without recomputing each sector.
- Preserve total and sectoral modelled-timeseries dimensions and numerical behavior.
- Add parity tests for one source, selected sources, and multiple sectors.
- Benchmark the vectorized path against the existing loop on a representative multi-sector workload.

Follow-up to #1713 and stacked Numba work in #1708.

Issue #1707, opened 2026-08-07
Guarantee monotonic time order when storing satellite footprints


brendan-m-murphy:
## Problem

Satellite footprints can be stored physically with a non-monotonic `time` coordinate. Retrieval normally sorts them, which masks the on-disk ordering, but direct Zarr access, time slicing, rechunking, and performance-sensitive Dask calculations are affected.

Irregular observation times do **not** require non-monotonic storage.

## Evidence from OCO-2 stores

Read-only inspection of four EASTASIA OCO-2 footprint stores found:

| Store | Records | Backward jumps | Notes |
|---|---:|---:|---|
| temporary integration store | 16,708 | 289 | 284 jumps within January 2023 |
| 3-day `H_back` store | 3,962 | 3 | four overlapping chronological runs |
| 5-day `H_back` store | 3,962 | 3 | identical timestamp set and physical order to 3-day store |
| newly created 1-day store | 3,962 | 0 | same April timestamp set as the 3/5-day stores, globally sorted |

The January 2023 block in the temporary store is especially structured: 5,830 timestamps are laid out in 285 runs, approximately “first release from each day, then second release from each day, ...”. This suggests that at least one incoming batch was already ordered like a flattened ragged `day × release` array. It is not explained by a single monthly append.

The 3/5-day stores were created and updated over several hours. Their four runs overlap in date, consistent with separately sorted batches being appended without a final global sort. Dataset-level date attributes also do not always describe the aggregate stored range.

## Likely code path

- Incoming datasets are sorted before assignment.
- Existing Zarr data are read in physical order for alignment.
- Accepted data are appended with `append_dim="time"`.
- No global on-disk sort occurs after append.
- Public retrieval sorts, hiding the issue.

## Work proposed

- Validate monotonic increasing, unique `time` before initial write.
- Guarantee the same invariant after every append/combine/update, without assuming an incoming batch occurs after the existing maximum time.
- Add a cheap fast path for true tail appends and a safe rewrite/merge path for interleaved data.
- Validate stored metadata dates against the actual aggregate coordinate range.
- Add tests for multiple individually sorted batches whose date ranges interleave.
- Check the external scripts used to create the OCO-2 satellite footprints and the source NetCDF ordering. In particular, determine why January 2023 is release-rank-major and whether files were passed one at a time, in groups, or flattened from a two-dimensional/ragged structure.
- Consider preserving source file / batch provenance so ordering problems can be traced later.
[...]

Issue #1705, opened 2026-08-07
Investigate overlap semantics for non-continuous time series


brendan-m-murphy:
## Problem

OpenGHG's default overlap detection uses exact coordinate equality. For non-continuous data, especially satellite observations and satellite footprints, regenerated records may describe the same physical sounding/release while differing slightly in timestamp. Such records are treated as non-overlapping and may be appended.

This deserves a deliberate policy rather than applying a tolerance blindly: nearby satellite timestamps can also represent genuinely distinct soundings.

## Current behaviour

- `OverlapDeterminer` uses exact `Index.get_indexer` matching unless `method` / `tolerance` options are explicitly supplied.
- Satellite footprints are stored with `continuous=False`.
- If no exact timestamp matches are found, `if_exists="auto"` can accept data covering an already represented date range.
- Zarr insertion appends accepted records along `time`.
- The same question applies to non-continuous satellite observations, which must remain aligned with their footprints.

In sampled OCO-2 stores, all timestamps were unique even where successive stored runs covered overlapping dates. Therefore exact duplicate removal did not identify the overlap.

## Questions to resolve

- What constitutes the identity of a satellite observation or footprint: timestamp, sounding ID, retrieval ID, spatial position, or a compound key?
- Should overlap detection consider date-range overlap separately for `continuous=False` inputs?
- Should parsers expose stable observation/release identifiers as coordinates?
- If a time tolerance is supported, how should ambiguity be handled when more than one record lies within tolerance?
- Must observation and footprint stores use the same matching policy?
- Should `if_exists="auto"` warn or fail when non-continuous inputs cover an existing date range but contain no exact timestamp matches?

## Acceptance criteria

- Define and document overlap semantics for non-continuous data.
- Add tests covering satellite observations and footprints with exact duplicates, slightly shifted timestamps, genuinely distinct nearby soundings, and partially overlapping date ranges.
- Ensure observation/footprint alignment cannot silently collapse or duplicate soundings.
- Emit an actionable warning or error for ambiguous updates.

PR #1731, opened 2026-09-07
Make Graphify regeneration reproducible

brendan-m-murphy:
* **Summary of changes** (development infrastructure)

Part 1 of the split of #1730. Adds a pinned, deterministic Graphify regeneration command and a focused navigation skill, scopes extraction to project code, commits the canonical graph while ignoring volatile artifacts, and checks freshness in CI. It also makes the smallest compatibility update to the existing Copilot instructions after removing `manifest.json`.

**What the graph contains**

`graphify-out/graph.json` is a generated navigation index of the OpenGHG codebase. It records symbols and source locations together with relationships such as imports, calls, and inheritance. It is useful for finding candidate dependencies and impact paths, but it is not a substitute for checking the source and tests. Consumers should use Graphify's `query`, `affected`, `explain`, and `path` commands rather than depend on the JSON's internal field names, which may change between Graphify releases. `.agents/skills/graphify-navigation/SKILL.md` provides the scoped query workflow and tells agents when Graphify is useful.

**How regeneration works**

- `.graphifyignore` limits extraction to `openghg/`, Python tests, and `pyproject.toml`, excluding fixtures and unrelated local files.
- `python scripts/update_graphify.py` invokes Graphify 0.9.55 with Python 3.12 through `uvx` and performs a fresh, code-only extraction in a temporary directory.
- The wrapper removes checkout-specific metadata and sorts the generated collections before atomically replacing `graphify-out/graph.json`. This makes equivalent graphs compare consistently across clones and worktrees.
- `python scripts/update_graphify.py --check` performs the same clean rebuild without updating the tracked file and fails when the committed graph is stale. CI runs this check on every pull request.
- `.gitattributes` marks the committed graph as generated so GitHub can collapse its large generated diff by default.
- The updater and navigation skill intentionally use the same pinned Graphify and Python versions; update both locations together when upgrading Graphify.

**Why `manifest.json` is not tracked**

The manifest is useful as a local incremental cache: it records per-file modification times and AST/semantic hashes so Graphify can skip unchanged files during incremental updates or watch operations. It is not project metadata or a map of global entry points.

It is unsuitable for version control because modification times vary between clones and worktrees, producing meaningless diffs. The deterministic updater performs a clean rebuild in a temporary directory and compares only the canonical `graph.json`; graph queries also consume `graph.json`, not the manifest. Graphify can recreate `manifest.json` whenever local incremental operation needs it, and `.gitignore` keeps that local copy untracked. The removal therefore affects only the committed artifact, not Graphify functionality.

- [x] Related to #1695
- [x] Infrastructure checks added and passed
- [x] All relevant code checks passing
- [x] Added an entry in `CHANGELOG.md`

Validation:

- `python scripts/update_graphify.py --check`
- Graphify navigation skill package validation
- Ruff format/check and mypy for `scripts/update_graphify.py`
- Generated-file and ignore attributes checked

*Not needed: behavioural pytest cases, tutorials, wiki changes, or runtime package requirements.*


brendan-m-murphy (commented):
I tested the pinned Graphify workflow in this PR at `3d12689ab` and asked four independent read-only code-mapping agents to use it on substantial open issues: #1707, #1705, #1701, and #1714.

## Overall result

Graphify is useful as a symbol index and impact-navigation aid, especially for locating exact definitions, immediate relationships, and likely tests. It is not reliable as a complete execution-path analyser, which is consistent with the caution already present in the skill.

The most useful workflow was:

1. Use a scoped query or source search to discover Graphify's exact node label.
2. Run `explain` on that exact label.
3. Use `affected` for callers/tests and `path` only as an architectural hint.
4. Verify the result against source and tests.

For example, this was concise and useful:

```bash
uvx --python 3.12 --from graphifyy==0.9.55 graphify explain ".add_timed_data()"
```

It returned a focused 17-edge neighbourhood containing `plan_timed_data_update()`, `add_data()`, date-range updates, and the storage calls. By contrast:

```bash
graphify affected "Datasource.add_timed_data"
```

could not uniquely resolve the qualified name. Generated method labels such as `.add_timed_data()` often worked better than conventional Python-qualified names.

## Query behavior observed

Worked well:

- `explain` with exact labels consistently found relevant definitions and direct relationships.
- `affected` found useful tests and direct callers for `standardise_footprint()`, `OverlapDeterminer`, `infer_date_range()`, `ModelScenario`, and calculation helpers.
- A focused natural-language query found the existing `test_add_data_out_of_order_no_combine` regression seam for #1707.
- Positive and negative paths helped establish that `ModelScenario` calls the legacy fp-x-flux operator and does not call the new vectorized Numba operator.

Limitations:

- Broad natural-language queries commonly returned 200–1,300 nodes and were truncated, even with an explicit call context.
- Dynamic factories and method dispatch break important paths. A path from `standardise_footprint()` to `.add_timed_data()` skipped the actual `BaseStore.standardise_and_store` route and instead crossed class/reference edges.
- `affected` missed callers of `ZarrStore.insert()` through `self._store.insert()`.
- Environment-variable relationships such as `OPENGHG_TUT_STORE` were invisible.
- Ambiguous common method names require exact generated IDs or labels.
- One result described `OverlapDeterminer` as re-exported by `storage/__init__.py`, but source verification showed that it is internal.

These are acceptable limitations for a navigation graph, but it may be worth adding two points to the skill:

- Tell users to discover the exact graph label and prefer `explain`/`affected` on that label; qualified Python names may not resolve.
- State explicitly that `path` is an architectural graph path, not necessarily a runtime call trace.

## Operational observation

The first pinned `uvx` invocation was not sandbox-friendly on this macOS setup: uv attempted to create its managed Python under `~/.local/share/uv/python`, outside the worktree. Redirecting only `UV_CACHE_DIR` was insufficient, and some sandboxed runs using temporary tool directories encountered a macOS SystemConfiguration panic. The exact pinned command worked after approval/outside the sandbox and subsequent cached use was fine.

This does not affect graph correctness, but agents in restricted environments may need either an already-provisioned Python 3.12/Graphify tool or guidance that the first `uvx` invocation can require permission beyond the repository.

## Issue maps produced

### #1707 — monotonic satellite-footprint storage

Graphify quickly located the core route and a near-regression test. Source verification showed that incoming batches are sorted independently, but `ZarrStore.insert()` appends them physically without proving that the new minimum follows the stored maximum. Retrieval sorts later and masks the defect.

The work crosses public standardisation, parsers, `BaseStore`, datasource update planning, Zarr insert/upsert behavior, version/date metadata, integrity checking, and likely repair tooling. This is genuinely substantial and is a good use case for the skill.

### #1705 — overlap semantics for non-continuous data

Graphify identified `OverlapDeterminer`, datasource update planning, storage implementations, and relevant tests. Source inspection was needed to establish that production matching is exact-time-only and that the GOSAT parser currently drops `exposure_id`, a potential stable sounding identifier.

The main blocker is a product/data-contract decision: stable identifier, compound identity, or time tolerance, including one-to-one ambiguity rules. A blind tolerance would be unsafe for distinct nearby soundings.

### #1701 — filesystem and process isolation

`affected` usefully exposed the broad fan-out of `clear_test_store(s)`, but Graphify could not model ownership, path collisions, cleanup races, or environment-variable state. Source inspection found the principal problem: all test sessions for one OS user share deterministic `/tmp/openghg-testing-<username>` stores and a fixed config file, while session/module/test cleanup recursively removes those paths.

The smallest first implementation is session/worker-owned pytest roots and narrowly owned cleanup. Runtime configuration redirection, logging state, and tutorial environment mutation can be separate follow-ups.

### #1714 — vectorized ModelScenario multi-sector fp-x-flux

Graphify clearly showed that the public vectorized `fp_x_flux_time_resolved_numba()` exists but has no path from `ModelScenario.calc_modelled_obs()`. The current route calculates a total and then repeats the full calculation once per source.

The likely change is to align/concatenate source fluxes once, run a source-dimensional calculation once, retain the sectoral result, and sum `source` for the total. Source inspection exposed important compatibility work around differing time grids, irregular footprints, optional Numba availability, dtype/NaN behavior, and incomplete ModelScenario cache identity.

## Conclusion

The PR's core positioning is correct: Graphify provides useful orientation and can substantially narrow initial source/test discovery, but important conclusions must be checked in source. Exact-label `explain` and `affected` queries gave the highest signal; broad free-text queries and dynamic call paths were much less dependable.

No repository files were changed during this evaluation.

SutarPrasad (approved):

PR #1741, opened 2026-09-16
Support Python 3.13/3.14 in CI; drop Python 3.10

SutarPrasad:

This PR adds Python 3.14 CI coverage and verifies that the existing zarr==2.18.3 / numcodecs<0.16 combination works when numcodecs can be compiled from source. It does not remove that source-build requirement. Local Python 3.14 installations therefore still require suitable build tools; Python 3.13 remains the recommended local version until the Zarr/numcodecs constraint can be upgraded.

- [ ] Closes #xxxx (Replace xxxx with the Github issue number) — not closed; see note above about #1742
- [ ] Tutorials updated/added
- [ ] Wiki updated
- [ ] Added any new package requirements to `pyproject.toml` [dependencies section] and `recipes/meta.yaml`

*Note: no new package requirements were added, only version constraints on existing dependencies were relaxed.*


rt17603:
Posted this on Teams but so we have a record:
 - Looking over this PR, it seems like it would be worth doing a release of openghg-defs so we don't have the restriction this seems to be adding, unless there was some reason not to.

SutarPrasad:
> Posted this on Teams but so we have a record:
>  - Looking over this PR, it seems like it would be worth doing a release of openghg-defs so we don't have the restriction this seems to be adding, unless there was some reason not to.

Release done on openghg_defs. And removed the fixes done over here.

doc/source/development/python_devel.rst
```diff
@@ -448,15 +453,15 @@ To run the full test suite, simply type:
 
 .. code-block:: bash
 
-   pytest tests/
+   python -m pytest tests/
```
rt17603: Was there a reason this needed to be updated?

brendan-m-murphy (changes requested):
Two correctness issues need addressing:

1. [P1] The declared minimum openghg-defs>=0.0.13 is not supported by the domain loader. openghg-defs 0.0.13 does not export data_path, but openghg/util/_domain.py still imports it in _get_coord_data. As a result, find_domain() fails immediately with ImportError in a valid 0.0.13 environment. CI resolves 0.0.14, which added a compatibility alias and masks this. Either require >=0.0.14 or use the 0.0.13-supported openghg_defs_files resource root.

2. [P2] The tracked .python-version containing 3.13 overrides the interpreter selected by actions/setup-python in the uv workflow. The nominal 3.11 and 3.12 job logs both show uv selecting CPython 3.13.15, and pytest also runs under 3.13.15. The conda jobs do cover those versions, but the uv/pip installation path does not. Pass the matrix version explicitly to uv, set UV_PYTHON, or otherwise override the repository default in CI.

The Traversable direction in util/_file.py is sound. Since Path and Traversable both provide .open(), str is the only special case; converting strings to Path first would allow a single shared .open() path, but I do not consider the current branching a blocking issue.

Issue #1253, opened 2025-03-17
Create equivalent `align_metadata_attributes` functions for all data types


rt17603:
Currently the metadata is checked and validated against the internal attributes for ObsSurface through [align_metadata_attributes](https://github.com/openghg/openghg/blob/devel/openghg/store/_obssurface.py#L369) function.

This should be applied for all data types. This should be done in a way that makes this more generalisable i.e. the `data_type` should determine what details are checked for this. This also helps to move towards implementing #1252 

This includes updates for:
 - [ ] BoundaryConditions
 - [ ] ObsColumn
 - [ ] Footprints
 - [ ] FootprintTimeseries
 - [ ] Flux
 - [ ] EulerianModel

brendan-m-murphy:
Try making "data type" an argument, and use this to select the "required attributes".

Could possibly refactor by using methods for merging dictionaries (in metadata util).

PR #1744, opened 2026-09-17
Removing the history keyword from the Eulerian metadata

rt17603:

This PR removes the `history` keyword is included in the EulerianModel metadata and updates this value in the attributes instead. The reason for doing this is:
1. The presence of this keyword in the attributes and updated within the metadata means they don't match
2. The history keyword is not a useful keyword to include in the metadata - it's not that informative and it's unlikely to be used for searching. This is better staying as just an attribute on the underlying dataset.

- [x] Linked to #1253 

Not needed

- Documentation updated/added
- Tutorials updated/added
- Wiki updated
- Added any new package requirements to `pyproject.toml` [dependencies section] and `recipes/meta.yaml`
- Added an entry in the latest `CHANGELOG.md` file if fixing a bug or adding a new feature

#1744 --> #1746 --> #1743


SutarPrasad (dismissed):
requires this to be ran `python scripts/update_graphify.py`. looks good

rt17603:
> requires this to be ran `python scripts/update_graphify.py`. looks good

Would be better if this could be done automatically rather than blocking a merge - we don't want using graphify to interfere with development really.

SutarPrasad (approved):

SutarPrasad:
> > requires this to be ran `python scripts/update_graphify.py`. looks good
> 
> Would be better if this could be done automatically rather than blocking a merge - we don't want using graphify to interfere with development really.

We can update the graphify ci to run the update command only when PR is merged into devel.

